Issue #10, opened 2024-09-27
Show the total number of active OWASP project on /projects page


arkid15r:
Based on the user's feedback.

Perhaps, we can add it into the search bar placeholder.

aravindparappil46:
Tried my hand at this one @arkid15r 🙂

PR #11, opened 2024-09-27
Add Total Count Of Projects To Search Bar Placeholder

aravindparappil46:
This PR addresses Issue #10 

Screenshot of the change:
image

Please feel free to suggest different phrasing for the placeholder text, or any other feedback :) 

**Question:** This change has only been made to the `Projects` page. The placeholder in the `Contribute` page search box also says "Search for a project to contribute to...". Perhaps we need to reword that sentence to be more generic (since people might search for keywords in an issue as well?)

### DIFF
--- a/backend/apps/owasp/api/search/project.py
+++ b/backend/apps/owasp/api/search/project.py
@@ -35,4 +35,7 @@ def get_projects(query, attributes=None, limit=25):
 
 def projects(request):
     """Search projects API endpoint."""
-    return JsonResponse(get_projects(request.GET.get("q", "")), safe=False)
+    return JsonResponse({
+        "active_projects_count": Project.active_projects_count(),
+        "projects": get_projects(request.GET.get("q", ""))
+    }, safe=False)
--- a/backend/apps/owasp/api/search/project.py
+++ b/backend/apps/owasp/api/search/project.py
@@ -35,7 +35,10 @@ def get_projects(query, attributes=None, limit=25):
 
 def projects(request):
     """Search projects API endpoint."""
-    return JsonResponse({
-        "active_projects_count": Project.active_projects_count(),
-        "projects": get_projects(request.GET.get("q", ""))
-    }, safe=False)
+    return JsonResponse(
+        {
+            "active_projects_count": Project.active_projects_count(),
+            "projects": get_projects(request.GET.get("q", "")),
+        },
+        safe=False,
+    )
--- a/backend/apps/owasp/templates/search/project.html
+++ b/backend/apps/owasp/templates/search/project.html
@@ -14,7 +14,7 @@
                                type="text"
                                @input="handleInput"
                                class="form-control"
-                               placeholder="Search for a project to contribute to...">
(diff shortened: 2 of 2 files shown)


backend/apps/owasp/templates/search/project.html
```diff
@@ -159,6 +159,11 @@ <h4 class="px-1">
         } = Vue;
         createApp({
             delimiters: ['${', '}'],
+            computed: {
+                searchPlaceholderText() {
+                    return `Search among ${this.projects.length} projects to contribute to...`;
```
arkid15r: This is the right implementation. But it's a half-done -- we'd like to emphasize the total amount of projects OWASP has (somewhat around 235). So you could either hardcode that value or create another API endpoint and get this number from `Project::active_projects_count` (requires backend work and API endpoint naming thoughts)

arkid15r (commented):
Thanks for the very first PR contribution into Nest code base.

arkid15r:
> **Question:** This change has only been made to the `Projects` page. The placeholder in the `Contribute` page search box also says "Search for a project to contribute to...". Perhaps we need to reword that sentence to be more generic (since people might search for keywords in an issue as well?)

I see what's you're saying. Could you create an issue or maybe a PR with an alternative wording and take care of this?

Maybe "Find an issue to solve..."

backend/apps/owasp/templates/search/project.html
aravindparappil46: @arkid15r -- I have added an endpoint `/api/v1/owasp/search/project/active-projects-count` which just returns the `Project.active_projects_count()`. Figured this is also a "search" of sorts, so kept it in the existing `search.py`. 

Please let me know if the naming/location is not intuitive and more than happy to iterate on this :)

backend/apps/owasp/api/search/project.py
```diff
@@ -36,3 +36,7 @@ def get_projects(query, attributes=None, limit=25):
 def projects(request):
     """Search projects API endpoint."""
     return JsonResponse(get_projects(request.GET.get("q", "")), safe=False)
+
+def active_projects_count(request):
```
arkid15r: Great idea. Do you think we could incorporate this data into the existing endpoint for projects? Basically, return a dict -- I think we can keep the API endpoints complexity lower this way.
```python
{"active_projects": Project.active_projects_count(), 
 "items": get_projects(request.GET.get("q", "")), safe=False
}```

arkid15r (commented):
I like the way you addressed the problem with an additional endpoint for showing the up to date value instead of hard-coding it.

Could you address the suggestion below?

backend/apps/owasp/api/search/project.py
aravindparappil46: Totally agree with this! Thanks @arkid15r !
Have pushed a commit :)

arkid15r (approved):
This is cool, I'm merging this in!

Thank you @aravindparappil46

Issue #6, opened 2024-09-24
Implement indexed entities cache


arkid15r:
As Algolia has a cap of 10K free search requests per month it would make sense to avoid querying index backend for the same search queries more than once.

The implementation should utilize Django cache https://docs.djangoproject.com/en/5.1/topics/cache/

The caching system should work for all entities (we currently have projects and issues).

The default TTL should be 86400 as we update the index on a daily basis. The update process must invalidate all cached entities.

arkid15r:
Implemented in #23

PR #19, opened 2024-09-30
Implemented indexed entities cache

onass1:
- Added cache logic to project.py and issue.py to retrieve search queries from the local memory cache system instead of sending repeated queries to algolia which helps in the problem of limited number algolia search queries.

- added CACHES dir to base.py to reconfigure the global settings of project.

- this refers to Issue #6 
- System: Linux/GNU Kali Rolling

### DIFF
--- a/backend/apps/owasp/api/search/project.py
+++ b/backend/apps/owasp/api/search/project.py
@@ -1,8 +1,9 @@
 """OWASP app project search API."""
 
 from algoliasearch_django import raw_search
-from django.http import JsonResponse
 from django.core.cache import cache
+from django.http import JsonResponse
+
 from apps.owasp.models.project import Project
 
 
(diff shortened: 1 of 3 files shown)


backend/apps/owasp/api/search/project.py
```diff
@@ -30,7 +31,20 @@ def get_projects(query, attributes=None, limit=25):
         "typoTolerance": "min",
     }
 
-    return raw_search(Project, query, params)["hits"]
+    project_cache_key = f"algolia_search_{query}"
```
arkid15r, line 34: Your project cache key clashes with the [issue cache key](https://github.com/OWASP/Nest/pull/19/files#diff-f70bd846c1b9929e4c147f39efca83fa890cbb9ae173d9d767d1f468487753e2R34).

backend/apps/owasp/api/search/project.py
```diff
@@ -30,7 +31,20 @@ def get_projects(query, attributes=None, limit=25):
...
+
+    if project_cache_result:
+        return project_cache_result
+
+    search_result = raw_search(Project, query, params)["hits"]
+
+    # save to cache
+
+    cache.set(project_cache_key, search_result, 86400)
+    return search_result
```
arkid15r, line 47: Let's make it more readable with something like
```

if issue_cache_result is None:
    search_result = raw_search(Issue, query, params)["hits"]
    cache.set(issue_cache_key, search_result, 86400)

return search_result
```

backend/apps/owasp/api/search/project.py
```diff
@@ -30,7 +31,20 @@ def get_projects(query, attributes=None, limit=25):
...
+    # check if cache exists
+
+    if project_cache_result:
+        return project_cache_result
+
+    search_result = raw_search(Project, query, params)["hits"]
+
+    # save to cache
+
+    cache.set(project_cache_key, search_result, 86400)
```
arkid15r, line 46: Could you create a separate constant for `86400` value in the constants.py file?

arkid15r (changes requested):
Great PR! Thanks for adding cache for project/issues

Please find my suggestions on how to improve the code below:

PR #23, opened 2024-10-05
Implement Indexed entities cache

onass1:
- Added cache logic to project.py and issue.py to retrieve search queries from the local memory cache system instead of sending repeated queries to algolia which helps in the problem of limited number algolia search queries.

- added CACHES dir to base.py to reconfigure the global settings of project.

- this refers to Issue #6 
- System: Linux/GNU Kali Rolling

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 build:
 	@docker compose build
 
+clear-cache:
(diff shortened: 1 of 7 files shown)


arkid15r (approved):
@onass1 thanks for the great contribution!
I've added some missing parts and going to merge it.

I'm looking forward to reviewing more PRs from you!

Issue #9, opened 2024-09-25
Add ability to exclude project from issue syncing/indexing


arkid15r:
Some projects have a unique approach for their issues workflows. This means that not all open issues are ready for picking up by potential contributors. For example, https://github.com/OWASP/ASVS/issues project heavily uses issues for discussing the work to be done w/o ever assigning anyone to the issue.

For this type of cases we could use a flag to exclude projects from issues indexing (or even sync) if needed.

This would require introducing a new boolean field for `Project` model and using it to skip certain actions based on the field's value.

arkid15r:
As an alternative we could keep them syncing/indexing but just skip AI summary/hint generation step for them.

Open to hear your ideas.

PR #45, opened 2024-10-26
 Add ability to exclude project from issue syncing/indexing

onass1:
Issue #9


backend/apps/github/models/repository.py
```diff
@@ -66,6 +66,7 @@ class Meta:
     has_pages = models.BooleanField(verbose_name="Has pages", default=False)
     has_projects = models.BooleanField(verbose_name="Has projects", default=False)
     has_wiki = models.BooleanField(verbose_name="Has wiki", default=False)
+    sync_issues = models.BooleanField(verbose_name="Sync Issues", default=True)
```
arkid15r:
```suggestion
    sync_issues = models.BooleanField(verbose_name="Sync issues", default=True)
```

backend/apps/owasp/models/project.py
```diff
@@ -87,6 +87,7 @@ class ProjectType(models.TextChoices):
     pushed_at = models.DateTimeField(verbose_name="Pushed at", blank=True, null=True)
     updated_at = models.DateTimeField(verbose_name="Updated at", blank=True, null=True)
 
+    sync_issues = models.BooleanField(verbose_name="Sync Issues", default=True)
```
arkid15r:
```suggestion
    sync_issues = models.BooleanField(verbose_name="Sync issues", default=True)
```

backend/apps/github/common.py
```diff
@@ -46,7 +46,7 @@ def sync_repository(gh_repository, organization=None, user=None):
     )
 
     # GitHub repository issues.
-    if not repository.is_archived and repository.project:
+    if not repository.is_archived and repository.project and repository.sync_issues:
```
arkid15r: You need to add the project level flag check too

backend/apps/github/models/repository.py
```diff
@@ -106,6 +107,7 @@ def is_indexable(self):
             not self.is_archived
             and not self.is_empty
             and not self.is_template
+            and self.sync_issues
```
arkid15r:
```suggestion
            and self.sync_issues
```
You need to modify issues indexing condition instead.

arkid15r (changes requested):
This is what we need -- it'll allow excluding on both project and repository levels.

It seems the migration files are missing. Could you add them?

backend/apps/github/common.py
naseyro: making it this way?
-----------------------------------------------------
    if not repository.is_archived and repository.project and repository.sync_issues and repository.project.sync_issues:

naseyro: I made an edit instead:
    `if not repository.is_archived and repository.project and repository.sync_issues and repository.project.is_indexable:`

also in project.py I modified the is_indexable function in Project model to:


`          return self.is_active and self.has_active_repositories and self.sync_issues`

naseyro: which way should I consider?

arkid15r: At this step let's stick with the first option that involves `sync_issues` flags only (don't touch `is_indexable`)

backend/apps/github/models/repository.py
naseyro: How? Isn't it already modified?

arkid15r: Please see the implementation in the recent commit.

arkid15r (approved):
Merging this in, thanks for the contribution @onass1!

PR #61, opened 2024-11-05
Set up new frontend

AbhayTopno:
set up of frontend with react + typescript and framework as vite and also added tailwind css for boosting the productivity to write code


harsh3dev:
I reviewed the code, and it functions correctly.
Just a small suggestion: You could add `/src/components` and `/src/pages` so that it can be easy to organize.

AbhayTopno:
Yeah it could be added in the later issues where we have add pagination

arkid15r (approved):
Thanks for adding this @AbhayTopno!

Issue #54, opened 2024-11-03
Implement reusable OWASP entity component


arkid15r:
The Nest app has four main pages: Contribute, Projects, Chapters, and Committees. Each page is different, but we can display their information in a similar style.

To make the code cleaner and ensure a consistent look across the site, we want to create one component that displays items on all pages. This component should be flexible, showing specific details based on what is passed to it. For example, tags or labels should appear on the Contribute and Projects pages, but not on the Chapters or Committees pages.

harsh3dev:
For this issue:
1. we need to create a card component which will display the information and it can be used in all the pages
2. It will be dynamic, based on the props that we pass onto it; It will display the relevant information 
I am ready to work on this issue

harsh3dev:
We will have to develop a card component which will 
1. show the information based on the props we pass
2. a container that will render all the cards
3. When the card is clicked it should perform the necessary actions like opening the modal
4. The card will dynamically show content and other details based on the props passed to it.
I am ready to work on this issue

arkid15r:
This issue it major one, I'm glad you want to work on it. @harsh3dev don't hesitate to reach out to me in Slack as there might be a lot of decisions to make during the implementation.

Assigning it to you, thank you!

arkid15r:
Implemented in #72 by @harsh3dev

PR #72, opened 2024-11-08
Implement Entity Component in React + TS

harsh3dev:
This PR solves issue #54 
This is how the project structure looks now:
image

### Overview
- The `EntityExampleComponent.tsx` is a temporary example component that displays the card and can be used as an example on how to use this component.
- This `EntityExampleComponent.tsx` shall be deleted later when the routing of the frontend is setup using [`React-Router`](https://reactrouter.com/en/main/start/overview).

#### Card Component (`src/components/Card.tsx`)
- The main component is `src/components/Card.tsx` which accepts few `props`.
- It displays information relevant to the props passed so that each page renders relevant details.
- For tooltip I have used `react-tooltip` package.
- The types and constants are exported from separate files.

- Also added `dark mode` and `light mode` functionality
- Using variables for colors in Tailwind CSS for consistency in styling.

Refer to this video for a quick look at the output:
https://www.loom.com/share/2e5be837be864bfcb5d02dff9ba5af17?sid=e87e552f-3ac4-4714-91da-04c4b5772d2c


frontend/package.json
arkid15r, line 1: I think it's fine to keep the non-dependencies attributes unchanged here.

frontend/index.html
```diff
@@ -1,8 +1,8 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.0.0-beta3/css/all.min.css" />
```
arkid15r: Shouldn't we use `@fortawesome/react-fontawesome` instead?

frontend/src/lib/constants.ts
```diff
@@ -0,0 +1,104 @@
...
+    level: "Production",
+  },
+  flagship: {
+    color: "#38a047",
+    icon: " text-white fa-solid fa-flag",
+    level: "Flagship",
+  },
+};
+
+export const urlMappings = [
```
arkid15r: It looks like `data.tsx` already contains same mapping.

frontend/src/lib/types.ts
```diff
@@ -0,0 +1,86 @@
...
+    idx_forks_count: number;
+    idx_leaders: string[];
+    idx_level: string;
+    idx_name: string;
+    idx_stars_count: number;
+    idx_summary: string;
+    idx_topics: string[];
+    idx_type: string;
+    idx_updated_at: number; // UNIX timestamp
+    idx_url: string;
```
arkid15r, line 23: I'd prefer to get rid of `idx_` prefix in interfaces if possible. It'd be more readable.
Do you think it's something that can be done?

arkid15r (commented):
@harsh3dev amazing work!
This PR gets us closer to the React-based frontend milestone.

Please find my questions/comments below:

frontend/index.html
harsh3dev: Yes, we can use react-fontawesome and i will replace it.

frontend/src/lib/constants.ts
harsh3dev: Okay, I see there are these redundant constants, I must have to clean them.

frontend/src/lib/types.ts
harsh3dev: Sure, I can remove the `idx_` prefix and make it more readable

harsh3dev:
@arkid15r I went through the suggestions and comments and I will update the commit soon.
Thanks

harsh3dev:
@arkid15r @tsu-ki Please have a look at the changes. Waiting for your approval.
Thanks.

arkid15r (approved):
Let's merge it!
Great work @harsh3dev 👍

PR #103, opened 2024-11-27
fixed vite reloading error

### LINKED ISSUE #102: Bug Report: Frontend Not Auto-Reloading in Docker Environment
When running the frontend application using Vite in a Docker container, changes in the code do not automatically trigger a page reload or reflect in the browser. 

https://github.com/user-attachments/assets/9fcb9a01-3e71-4806-a75e-6f33ce53a669

Yashgupta9330:
fixes #102 

https://github.com/user-attachments/assets/23a44612-9c7f-4de0-9ead-baf54a71f4f7

### DIFF
--- a/frontend/Dockerfile.local
+++ b/frontend/Dockerfile.local
@@ -1,16 +1,27 @@
 FROM node:18-slim
 
+
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
 
 RUN groupadd owasp && \
     useradd --create-home --home-dir /home/owasp -g owasp owasp && \
     apt-get update && apt-get upgrade -y && \
-    apt-get clean -y && rm -rf /var/lib/apt/lists/*
+    apt-get install -y \
+    curl \
+    && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+
 
 EXPOSE 3000
 
+
+WORKDIR /home/owasp
+
+
 USER owasp
+
 
 VOLUME /home/owasp
 
-WORKDIR /home/owasp
+
+CMD ["npm", "install", "&&", "npm", "run", "dev"]
--- a/frontend/Dockerfile.local
+++ b/frontend/Dockerfile.local
@@ -1,27 +1,16 @@
 FROM node:18-slim
 
-
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-
 
 RUN groupadd owasp && \
     useradd --create-home --home-dir /home/owasp -g owasp owasp && \
     apt-get update && apt-get upgrade -y && \
-    apt-get install -y \
-    curl \
-    && apt-get clean -y && rm -rf /var/lib/apt/lists/*
-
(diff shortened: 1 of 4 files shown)


frontend/vite.config.ts
arkid15r, line 1: I'm not sure but it seems this is the only file that needs to be changed for getting it done.
Could you check this and revert all unrelated changes (if any)?

I'd also like you to search for a way to enable polling for [Windows machines only](https://vite.dev/config/server-options#server-watch) if possible. Let's narrow it down to affected environments only.
```

import os from 'os';

// Get the OS type
const osType = os.type();
```

frontend/Dockerfile.local
arkid15r, line 1: I don't think we need to change this file as we run [npm commands here](https://github.com/OWASP/Nest/blob/e62de0696159ab6f858b64552c8fdfbfed51a46f/docker-compose.yaml#L56)

Yashgupta9330: I have removed extra part in dockerfile.local

frontend/vite.config.ts
```diff
@@ -1,7 +1,18 @@
 import { defineConfig } from 'vite'
 import react from '@vitejs/plugin-react'
+import os from 'os'
 
-// https://vite.dev/config/
+const platform = os.platform()
+console.log("platofrom" , platform)
```
arkid15r:
```suggestion
```

frontend/vite.config.ts
```diff
@@ -1,7 +1,18 @@
...
 export default defineConfig({
   plugins: [react()],
+  server: {
+    watch: {
+      usePolling: platform === 'win32' || platform === 'linux',
```
arkid15r:
```suggestion
      usePolling: os.platform() === 'win32',
```

arkid15r: We wanted it on Windows only, didn't we?

arkid15r (commented):
A couple more suggestions:

frontend/vite.config.ts
Yashgupta9330: I saw in google we need in linux also

arkid15r: I've made the changes based on https://vite.dev/config/server-options#server-watch

Could you provide your source?

Yashgupta9330: does the container os is linux?

Yashgupta9330: in console it is returning me linux

arkid15r: Let's focus on this: does hot reload work for you if you remove the 'linux' platform (keeping 'win32' only)?

Yashgupta9330: No

arkid15r: I see... As we run it in a container, the host OS is irrelevant here. Just submit it as always enabled (usePolling: true) and we'll take care of customization later when we have frontend env variables support.

Sorry for the hassle with this issue @Yashgupta9330

Issue #55, opened 2024-11-03
Implement reusable search component


arkid15r:
Like the approach for [a shared component for all items](https://github.com/OWASP/Nest/issues/54), we want a search bar that works across all pages (Contribute, Projects, Chapters, and Committees).

This search bar should be flexible and configurable, allowing it to connect with the backend and provide the right search features for each page.

Yashgupta9330:
I want to work on this issue can you please assign me this issue?

arkid15r:
You got it

tsu-ki:
@Yashgupta9330 Are you still working on this issue? I noticed that you've also assigned yourself another issue(https://github.com/OWASP/Nest/issues/68). So, if you aren't working on this, then I'd like to take over

Yashgupta9330:
I have made the changes and am waiting for the Django PR to be merged so that I can test it.

arkid15r:
> I have made the changes and am waiting for the Django PR to be merged so that I can test it.

Hey @Yashgupta9330 what PR blocks you (not sure if it's something I can help you with)?

Yashgupta9330:
To make API calls, we first need to resolve the CORS issue in Django.

arkid15r:
Gotcha! Yes, I've reviewed https://github.com/OWASP/Nest/pull/100 and waiting for the comments to be addressed. I hope it'll be merged today.

You can also reference the PRs in your issue replies as a blockers. That'd save some time for us.

arkid15r:
@Yashgupta9330 just FYI -- the https://github.com/OWASP/Nest/pull/100 has been merged.

arkid15r:
Hi @Yashgupta9330, are you still working on this? Please free up any issues you don't work on (if any).
Thank you!

PR #115, opened 2024-12-03
Added search component

Yashgupta9330:
made global search component https://github.com/OWASP/Nest/issues/55

### DIFF
--- a/backend/poetry.lock
+++ b/backend/poetry.lock
@@ -1881,13 +1881,13 @@ crt = ["botocore[crt] (>=1.33.2,<2.0a.0)"]
 
 [[package]]
 name = "sentry-sdk"
-version = "2.19.0"
+version = "2.19.2"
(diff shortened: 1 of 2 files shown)


arkid15r (changes requested):
@Yashgupta9330 I guess it's the last piece we need in order to replicate the old's frontend behavior. Thanks for implementing this. 
Please make sure to sync the most recent changes from the repo before addressing the following comments:

  - run linter and formatter (or just make check if you have set it up)
  - integrate the new search component to the Projects page (we need it to actually work and show relevant results)
  - make the search bar behavior consistent with the old backend: the results should be updated upon user typing event (also make sure not to send every key event to the backend but wait until user stops typing and then get the entity search result from BE; I think it currently 1 second delay -- the technique is called `debouncing`)
  - please also check that URLs like this `https://nest.owasp.dev/projects/?q=nest` work with the new component.

Thank you!

PR #151, opened 2024-12-08
Update Card component styles

kasya:
Updated styles for Card component to work better with dark mode:
- Moved leaders of the project to be lower on the card, after the summary.
- Added borders to blocks with number of forks/stars/contributors.
- Updated labels to have different background color based on type - language or topic.

Updated README file.


arkid15r (approved):
Thanks for better styling @kasya

PR #154, opened 2024-12-09
Update Card labels and contribute button layout.

kasya:
Updated styling and layout for the project labels and contribute button in the Card component:
 - wrapped labels into a separate div
 - aligned Contribute button with the labels div to always be at the same spot and not be dependent on labels
 - added conditional rendering for the whole labels div and for separate sub-divs (languages, topics, social media) so that there's no empty space at the bottom of the card if labels are missing
 - increased bottom-padding in DisplayIcon component
 - Decreased border-radius for DisplayIcon border

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -39,8 +39,11 @@ index-data:
 
 load-data:
 	@echo "Loading Nest data"
-	@gunzip backend/data/nest.json
+	@cp backend/data/nest.json.gz backend/data/copy.json.gz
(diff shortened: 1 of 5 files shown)


arkid15r (dismissed):
Awesome! 
Thank you!

arkid15r (dismissed):

arkid15r (dismissed):

arkid15r (approved):

PR #179, opened 2024-12-14
Implement Pagination Backend

### LINKED ISSUE #56: Implement reusable pagination component
Like the approach for [a shared component for all items](https://github.com/OWASP/Nest/issues/54), we want a pagination component that works across all pages (Contribute, Projects, Chapters, and Committees).

This pagination component should be flexible and configurable, allowing it to connect with the backend and provide the right pagination features for each page.

harsh3dev:
Resolves #56 

* This pull request introduces pagination to all API endpoints in the `backend/apps/owasp/api/search` 
* add page parameter
* include `nbPages` for getting total number of pages


Rajgupta36 (changes requested):
hi @harsh3dev ,please set args page=1 by default and use the page-1 logic

backend/apps/owasp/api/search/chapter.py
```diff
@@ -8,7 +8,7 @@
 from apps.owasp.models.chapter import Chapter
 
 
-def get_chapters(query, attributes=None, limit=25, meta=None):
+def get_chapters(query, attributes=None, limit=25, meta=None, page=0):
```
arkid15r:
```suggestion
def get_chapters(query, attributes=None, limit=25, meta=None, page=1):
```

backend/apps/owasp/api/search/chapter.py
```diff
@@ -26,23 +26,25 @@ def get_chapters(query, attributes=None, limit=25, meta=None):
         "hitsPerPage": limit,
         "minProximity": 4,
         "typoTolerance": "min",
+        "page": page,
```
arkid15r:
```suggestion
        "page": page - 1,
```

arkid15r (changes requested):
I agree `page=1` by default is more readable.

Could you apply the changes everywhere in the code?

Other than that looks good!

backend/apps/owasp/api/search/chapter.py
Rajgupta36: yes

Rajgupta36 (changes requested):
Thanks 👍👍

arkid15r (approved):
I updated this a bit and merging into main
Thanks for another contribution @harsh3dev !

PR #183, opened 2024-12-15
Feature : Pagination sharable comopnent

### LINKED ISSUE #56: Implement reusable pagination component

Rajgupta36:
Resolves #56  

## Description  
This PR includes the following changes:  
1. Implemented a reusable pagination component.  
2. Integrated the pagination component into the **Project** page.  
3. Integrated the pagination component into the **Committee** page.  
4. Adjusted Vue templates to ensure compatibility with the old project. However, pagination is not applied to the old project templates.  

## Checklist  
- [x] Pagination component is reusable.  
- [x] Pagination is functional on the **Project** page.  
- [x] Pagination is functional on the **Committee** page.  
- [x] Vue templates adjusted for compatibility with the old project.  

## Screenshots 
light mode
image
 ---
 dark mode
  image
  image


Rajgupta36:
@arkid15r , please merge it after the backend PR

frontend/src/components/pagination.tsx
arkid15r, line 1: I believe the convention here is `Pagination.tsx`?

backend/apps/owasp/templates/search/chapter.html
```diff
@@ -125,7 +125,7 @@ <h4 class="px-1">
             methods: {
                 async getChapters() {
 
-                    const response = await fetch(`/api/v1/owasp/search/chapter?q=${this.searchQuery}`)
+                    const response = await fetch(`/api/v1/owasp/search/chapter?q=${this.searchQuery}&page=1`)
```
arkid15r: That's the default one, right?
I think we can avoid the old backend code updates.

frontend/src/components/pagination.tsx
```diff
@@ -0,0 +1,101 @@
+import { MoreHorizontal } from 'lucide-react'
+import React from 'react'
+
+interface PaginationProps {
+  currentPage: number
+  totalPages: number
+  isLoaded: boolean
+  // eslint-disable-next-line no-unused-vars
+  onPageChange: (page: number) => void
```
arkid15r: Do you know if there is a way to avoid this suppression -- handle it in other/better way?

frontend/src/components/pagination.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    if (totalPages > 3 && !pageNumbers.includes(totalPages)) {
+      pageNumbers.push(totalPages)
+    }
+
+    return pageNumbers
+  }
+
+  const pageNumbers = getPageNumbers()
+  console.log(totalPages)
```
arkid15r:
```suggestion
```

arkid15r (changes requested):
It works fine, I just want it to show the top of the page after click on other page.

Please also look into these:

backend/apps/owasp/templates/search/chapter.html
Rajgupta36: Yes, but some users are encountering errors while setting up the React frontend. So, I think for now, if they work on the backend, the Vue frontend will reflect the results for them.

frontend/src/components/pagination.tsx
Rajgupta36: i will try it .

frontend/src/components/pagination.tsx
Rajgupta36: yes

backend/apps/owasp/templates/search/chapter.html
arkid15r: We don't use the new frontend yet. It's in dev only.
My point it that you don't need to change the query as page=1 is the default value on the backend.

Rajgupta36: okk

frontend/src/components/pagination.tsx
Rajgupta36: we can't resolve the ESLint issue because it is triggered by another function, and due to the state change, it causes that function to re-render.

PR #176, opened 2024-12-14
Added Sentry Monitoring

tsu-ki:
Resolves #(https://github.com/OWASP/Nest/issues/112)

@arkid15r can you please specify if I should create a DNS for Sentry, or one already exists for it. 
I'll test the proposed changes first and then update the PR accordingly

PR #187, opened 2024-12-16
enforced lint to avoid console.* in prod

### LINKED ISSUE #186: Enforce 'no-console': 'error' eslint rule
Implement changes to enable `'no-console': 'error',` eslint rule

Using console.* in prod is not a good practice. We need to clean up existing cases and enable the rule to enforce it.

yashpandey06:
Fixes #186 
This PR enforces linting rules to disallow the use of `console.*` statements in the project. These changes include adding a lint rule to the CI/CD pipeline and updating the `package.json` to ensure that the rule is checked and enforced during the development and deployment process.

### Summary of Changes
- **CI/CD Pipeline**: 
  - Updated the GitHub Actions CI/CD workflow to run `npm run lint` as part of the pipeline, ensuring that the `console.*` lint rule is enforced in every push/merge request.
  
- **package.json**:
  - Ensured the `eslint` configuration is applied to disallow `console.*` statements during linting.
  - Added a `lint:check` script to check for any `console.*` usage and fail the build if any are found.


yashpandey06:
@arkid15r please help review these changes .

.github/workflows/ci-cd.yaml
```diff
@@ -64,6 +64,9 @@ jobs:
         working-directory: frontend
         run: npm run lint
 
+      - name: Run npm lint check
```
arkid15r, line 63: It seems it should be enough to have just one: `npm run lint` or `npm run lint:check`

frontend/src/pages/Committees.tsx
```diff
@@ -17,7 +17,7 @@ export default function Committees() {
         const data = await response.json()
         setCommitteeData(data)
       } catch (error) {
-        console.error(error)
+        throw new Error(`Error: ${(error as Error).message}`)
```
arkid15r: From my perspective (not that proficient in the FE development) we need to show the error message to a user and then can throw it. It seems we need a unified way of showing the error messages.

frontend/src/pages/Projects.tsx
```diff
@@ -21,7 +21,7 @@ export default function Projects() {
         setProjectData(data)
         setDefaultProjects(data)
       } catch (error) {
-        console.error(error)
+        throw new Error(`Error: ${(error as Error).message}`)
```
arkid15r: what if it's not an `Error`?

frontend/package-lock.json
```diff
@@ -2675,6 +2676,13 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/@types/mocha": {
```
arkid15r: I guess it can be done it a separate update?

arkid15r (changes requested):
Thanks for the PR @yashpandey06 
Let's discuss some topics below:

frontend/src/pages/Committees.tsx
yashpandey06: I see we have created issue for introducing unified way showing the error to a user using some notification box ...right ?

so for here now i need to take of that thing but handle the error throwing gracefully ?

@arkid15r

frontend/package-lock.json
yashpandey06: handled

frontend/src/utils/logger.ts
```diff
@@ -0,0 +1,39 @@
+import { APPLICATION_ENV } from './credentials'
+
+/**
+ * This is the central logging configuration for the Nest.
+ * In development, logs are output to the console.
+ * In production, logs will be sent to Sentry (or a centralized logging service) for error tracking and monitoring.
+ * Sentry integration should be implemented here to ensure centralized and scalable error handling.
```
yashpandey06: @arkid15r 

Sentry integration should be written here so as to manage logs both in dev and prod env properly. Therefore added this comment.

frontend/src/utils/logger.ts
arkid15r, line 1: The problem w/ this approach is that Sentry is for error logging only. It's not a centralized logging service per se.
We want to send errors/unhandled exceptions only and be alerted about them.

frontend/src/utils/logger.ts
arkid15r: I believe Sentry has a documentation on how it should be set up. Not sure if that aligns w/ your expectations.

frontend/src/utils/logger.ts
yashpandey06: our expectation with this task was to avoid using console.* in production 

Sentry will capture the unhandled exceptions in production  (post integration)

For eg :

` error: (error: Error) => {
    logger.error(error);// working only locally and in dev mode
    Sentry.captureException(error); // handling exceptions in production
  }
`

notifying alert is a scope of this [issue](https://github.com/OWASP/Nest/issues/190 ):
 right?
 which will give a toast notification like dialogue appearance based on error alerting the end user about the issue.
 
 
 @arkid15r

yashpandey06: Sentry integration is the scope of this [PR](https://github.com/OWASP/Nest/pull/176) .

frontend/src/utils/logger.ts
yashpandey06: Sentry integration is the scope of this https://github.com/OWASP/Nest/pull/176

frontend/src/utils/logger.ts
arkid15r: My comment was mostly about log/warn placeholders you created in `logger.ts`. We only need errors to be caught by Sentry.

I don't see a full picture of logging + showing error messages to users + Sentry integration yet

frontend/package.json
```diff
@@ -4,8 +4,8 @@
   "version": "0.0.1",
   "type": "module",
   "scripts": {
-    "dev": "vite --host --port 3000",
-    "build": "tsc -b && vite build",
+    "dev": "cross-env VITE_APPLICATION_ENV=development vite --host --port 3000",
```
arkid15r:
```
nest-frontend  | sh: 1: cross-env: not found
nest-frontend exited with code 127
```

arkid15r (commented):
I don't see the eslint rule for actual `no-console` enforcement.

frontend/eslint.config.js
```diff
@@ -62,6 +62,8 @@ export default [
           'newlines-between': 'always',
         },
       ],
+      'no-console': 'error',
     },
+    ignores: ['src/utils/*'],
```
arkid15r: This is too broad. Suppress only one rules you want in the file.

frontend/src/sentry.config.ts
```diff
@@ -0,0 +1,21 @@
...
+  tracesSampleRate: 1.0,
+  replaysSessionSampleRate: 0.1,
+  replaysOnErrorSampleRate: 1.0,
+  beforeSend(event) {
+    if (event.user) {
+      delete event.user.email
+    }
+    return event
+  },
+})
```
arkid15r: I don't think we need this.

frontend/src/utils/logger.ts
```diff
@@ -0,0 +1,38 @@
...
+    const message = args
+      .map((arg) => (typeof arg === 'string' ? arg : JSON.stringify(arg)))
+      .join(' ')
+
+    if (APPLICATION_ENV === 'development') {
+      console.log(...args)
+    } else {
+      logMessage(message)
+    }
+  },
```
arkid15r: I like the `logger` abstraction. Please note that we don't want to treat regular log entries as errors in prod -- don't send them to Sentry.

frontend/package.json
```diff
@@ -29,6 +32,7 @@
     "react-dom": "^19.0.0",
     "react-icons": "^5.3.0",
     "react-router-dom": "^7.0.2",
+    "react-toastify": "^11.0.0",
```
arkid15r: Do you use it anywhere?

arkid15r (commented):
It looks better but I haven't tested it. Please address the recently added comments and let's merge it. I think it's good enough to be merged and improved incrementally.

Thanks for adding this!

frontend/package.json
yashpandey06: was testing something ...removing it

frontend/src/utils/logger.ts
yashpandey06: okay modified logger.ts

arkid15r (approved):
Let's merge and start using it

PR #192, opened 2024-12-16
Refactor Website Structure :  Added Loader, Pagination, Search, and SearchAPI

Rajgupta36:
Resolves #56  #55 

## Description  
This PR includes the following changes:  
1. Introduced a loader for our website
1. Implemented a reusable pagination component.  
2. Implemented a reusable search component
3. Implemented a reusable Searchapi component
5. Integrated the pagination and searching component into the all pagea.  
6. updated the types 
7. Update test cases

## Checklist  
- [x] Pagination component is reusable.  
- [x] Pagination is functional on the all pages.
- [x] Searchcomponent is reusable.   
- [x] Seachapiendpoint correctly working 
- [x] Loader is working
- [x] no extra api calls 
- [x] updated test cases

## media
----
### pagination component
light mode
image
 ---
 dark mode
  image
  image
---
### SearchBar component
image
---
Loader component
image
image
---
working

https://github.com/user-attachments/assets/544654bc-608e-4258-ad7c-41e23dd36c1d

PR #199, opened 2024-12-17
Added query for search url

### LINKED ISSUE #198: Add Query Parameter to Search Functionality
The current search component does not read the query from url

Yashgupta9330:
Resolves #198 

This pr fixes focus on search and type of results and added query for search
Added search in chapter,committee page
Added test for chapter,project,committee pages

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,14 @@
 
 Follow these steps to set up the OWASP Nest application:
 
+1. **Fork the Repository**:
+   - Fork <https://github.com/OWASP/Nest> repository using "Fork" button
+
 1. **Clone the Repository**:
-   - Clone the repository code from GitHub using the following command:
+   - Clone the repository code from your GitHub account using the following command:
 
      ```bash
-     git clone https://github.com/owasp/nest
+     git clone https://github.com/<your-account>/<nest-fork>
      ```
 
 1. **Create Environment Files**:
@@ -21,14 +24,21 @@ Follow these steps to set up the OWASP Nest application:
    - Copy the contents from the template file into your new local environment file:
 
      ```bash
-     cp backend/.env/template backend/.env/local
+     cat backend/.env/template > backend/.env/local
      ```
 
    - Create a local environment file in the `frontend` directory:
 
      ```bash
-     cp frontend/.env.example frontend/.env
+     touch frontend/.env
      ```
+
+   - Copy the contents from the template file into your new local environment file:
+
+     ```bash
+     cat frontend/.env.example > frontend/.env
+     ```
+
 
 1. **Configure Environment Variables**:
    - Open the `backend/.env/local` file in your preferred text editor and change the `DJANGO_CONFIGURATION` value to `Local`:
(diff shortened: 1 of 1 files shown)


Rajgupta36:
hiii @Yashgupta9330 , it is already include in pr #192 .

Yashgupta9330:
@Rajgupta36, actually in my previous PR for search, I hadn't added the feature to fetch the query from the search URL. After that, I realized there was an issue with focus and a type problem. @arkid15r told me this 2 days ago and advised me to handle it in a separate PR. I contacted @harsh3dev ,he was unable to add the search to the chapter due to the type issue, so I added this feature myself and also included the test for the committee project chapter pages in this PR

Rajgupta36:
> @Rajgupta36, actually in my previous PR for search, I hadn't added the feature to fetch the query from the search URL. After that, I realized there was an issue with focus and a type problem. @arkid15r told me this 2 days ago and advised me to handle it in a separate PR. I contacted @harsh3dev ,he was unable to add the search to the chapter due to the type issue, so I added this feature myself and also included the test for the committee project chapter pages in this PR

yes i created pagination component yesterday but when i integrate it with search bar . I also facing same issue because of statement management . thanks for the pr. i am working on pagination so maybe we require some changes in it. thanks

Rajgupta36:
The problem with this approach is that when we implement pagination, it calls the components multiple times that's make it inefficient. Can we discuss this so that we can create a better pagination system with the better search component?

frontend/src/pages/Projects.tsx
```diff
@@ -11,19 +12,36 @@ import { API_URL } from '../utils/credentials.ts'
...
   const [defaultProjects, setDefaultProjects] = useState<ProjectDataType | null>(null)
+  const [initialQuery, setInitialQuery] = useState<string>('')
 
   useEffect(() => {
     document.title = 'OWASP Projects'
+    const queryParams = new URLSearchParams(window.location.search)
+    const urlQuery = queryParams.get('q')
+    if (urlQuery) {
+      setInitialQuery(urlQuery)
+    }
```
arkid15r: Isn't this code the same for all pages? Can it be done inside `Search` component?

frontend/src/pages/Projects.tsx
```diff
@@ -11,19 +12,36 @@ import { API_URL } from '../utils/credentials.ts'
...
-        const data = await response.json()
+        let response
+        if (urlQuery) {
+          response = await axios.get(`${API_URL}/owasp/search/project`, {
+            params: { q: urlQuery },
+          })
+        } else {
+          response = await fetch(`${API_URL}/owasp/search/project`)
+          response = await response.json()
+        }
```
arkid15r: Could you narrow it down to `params` difference only based on `urlQuery`. 

BTW, why do you use both `axios` and `fetch`?

frontend/src/components/Search.tsx
arkid15r, line 1: It doesn't show the default search results if I search for something and then make the search query blank.

arkid15r (changes requested):
Please look into these when you get a chance:

frontend/src/pages/Projects.tsx
Yashgupta9330: I have made a hook for this

frontend/src/pages/Projects.tsx
Yashgupta9330: i have used axios in both one api is for without query and another one is with query

frontend/src/components/Search.tsx
Yashgupta9330: I have fixed it

Yashgupta9330:
> The problem with this approach is that when we implement pagination, it calls the components multiple times that's make it inefficient. Can we discuss this so that we can create a better pagination system with the better search component?

I will ping you in slack

arkid15r (commented):
https://github.com/OWASP/Nest/pull/192 has been just merged and it has totally new structure. I suggest focusing on adding `?q=` functionality in the new PR instead of fixing the conflicts (up to you).

PR #209, opened 2024-12-19
Configure Dockerfile for staging environment

### LINKED ISSUE #191: Prepare the new frontend for staging environment
Once we're done w/ migration to the new frontend we'll need to run it within our [docker-compose-staging.yaml ](https://github.com/OWASP/Nest/blob/0629eba2aec1e1742431cde3e155cebaa32cafa4/docker-compose-staging.yaml) stack. We currently have some content in the [Dockerfile.staging](https://github.com/OWASP/Nest/blob/0629eba2aec1e1742431cde3e155cebaa32cafa4/frontend/Dockerfile.staging) but it has never been tested/used.

Revisit Dockerfile.staging for the new frontend and modify/improve if needed. Update docker-compose-staging.yaml to include the new frontend in staging environment

harsh3dev:
Resolves #191 

This pull request includes several changes to the `docker-compose-staging.yaml`, `frontend/Dockerfile.staging`, and `nginx/staging.conf` files to integrate and serve the frontend application in the staging environment. The most important changes are listed below:

Docker Compose configuration updates:
- [x] Added a new `frontend` service configuration to the `docker-compose-staging.yaml` file. This includes setting up the container, build context, environment variables, and networking.
- [x] Updated the `nginx` service configuration to use `staging.conf` instead of `nginx.conf` and added dependency on the `frontend` service.
- [ ] Commit the requested changes and test in local env

### DIFF
--- a/docker-compose-staging.yaml
+++ b/docker-compose-staging.yaml
@@ -21,6 +21,18 @@ services:
     volumes:
       - ./data:/home/owasp/data
 
+  frontend:
+    container_name: nest-frontend
+    build:
+      context: ./frontend
+      dockerfile: Dockerfile.staging
+    env_file: ./frontend/.env
+    restart: unless-stopped
+    networks:
+      - nest_app_network
+    expose:
+      - "80"
+
   certbot:
     container_name: nest-certbot
     image: certbot/certbot
@@ -61,10 +73,11 @@ services:
       - '443:443'
     volumes:
       - ./letsencrypt:/etc/letsencrypt
-      - ./nginx/nginx.conf:/etc/nginx/conf.d/default.conf
+      - ./nginx/staging.conf:/etc/nginx/conf.d/default.conf
       - certbot-webroot:/var/www/certbot
     depends_on:
       - backend
+      - frontend
     networks:
       - nest_app_network
 
--- a/docker-compose-staging.yaml
+++ b/docker-compose-staging.yaml
@@ -73,7 +73,7 @@ services:
       - '443:443'
     volumes:
       - ./letsencrypt:/etc/letsencrypt
-      - ./nginx/staging.conf:/etc/nginx/conf.d/default.conf
+      - ./nginx/nginx.conf:/etc/nginx/conf.d/default.conf
       - certbot-webroot:/var/www/certbot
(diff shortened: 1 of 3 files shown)


frontend/Dockerfile.staging
```diff
@@ -7,6 +7,7 @@ RUN npm install
 
 COPY . .
 RUN npm run build
+COPY .env .env
```
arkid15r, line 10: This isn't right. In this case you'll have your secrets in the image which is available for downloading to anyone.

docker-compose-staging.yaml
```diff
@@ -21,6 +21,18 @@ services:
     volumes:
       - ./data:/home/owasp/data
 
+  frontend:
+    container_name: nest-frontend
+    build:
+      context: ./frontend
+      dockerfile: Dockerfile.staging
+    env_file: ./frontend/.env
```
arkid15r, line 29: Use backend naming approach here too

docker-compose-staging.yaml
```diff
@@ -21,6 +21,18 @@ services:
     volumes:
       - ./data:/home/owasp/data
 
+  frontend:
+    container_name: nest-frontend
+    build:
```
arkid15r, line 26: We don't build on staging, we want to use a prebuilt image (see how it's done for backend)

docker-compose-staging.yaml
```diff
@@ -21,6 +21,18 @@ services:
...
+      context: ./frontend
+      dockerfile: Dockerfile.staging
+    env_file: ./frontend/.env
+    restart: unless-stopped
+    networks:
+      - nest_app_network
+    expose:
+      - "80"
```
arkid15r, line 34: Let's use 3000 for the FE for consistency.

arkid15r (changes requested):
This needs some more work:

yashpandey06:
@arkid15r  can i work on this ?...the image will be served from docker-hub for frontend.

arkid15r:
Implemented in https://github.com/OWASP/Nest/commit/31626655b0cb5ea68ccbcc9f4bbb874496c6b83c

PR #226, opened 2024-12-20
synced search bar with data appearances

### LINKED ISSUE #215: Show search bar upon the main content rendering
Currently the search bar loads before the main content and it looks strange.

Let's make it visible upon/after the main content is loaded.

yashpandey06:
Resolves #215 

Correct positioning of Search component solves the weird appearance .

### DIFF
--- a/frontend/__tests__/src/pages/Chapters.test.tsx
+++ b/frontend/__tests__/src/pages/Chapters.test.tsx
@@ -39,9 +39,6 @@ describe('ChaptersPage Component', () => {
     })
   })
 
-
-
-
   test('renders chapter data correctly', async () => {
     render(<ChaptersPage />)
     await waitFor(() => {
@@ -80,7 +77,7 @@ describe('ChaptersPage Component', () => {
     })
   })
 
-  test('renders SearchBar,data and pagination component concurrently', async () => {
+  test('renders SearchBar, data, and pagination component concurrently after data is loaded', async () => {
     window.scrollTo = jest.fn()
     ;(loadData as jest.Mock).mockResolvedValue({
       ...mockChapterData,
@@ -91,14 +88,18 @@ describe('ChaptersPage Component', () => {
     const loadingSpinner = screen.getAllByAltText('Loading indicator')
     await waitFor(() => {
       expect(loadingSpinner.length).toBeGreaterThan(0)
+      expect(screen.queryByPlaceholderText('Search for OWASP chapters...')).not.toBeInTheDocument()
+      expect(screen.queryByText('Next Page')).not.toBeInTheDocument()
     })
-
     await waitFor(() => {
       expect(screen.getByPlaceholderText('Search for OWASP chapters...')).toBeInTheDocument()
       expect(screen.getByText('Chapter 1')).toBeInTheDocument()
+      expect(screen.getByText('Next Page')).toBeInTheDocument()
+
       const nextPageButton = screen.getByText('Next Page')
       fireEvent.click(nextPageButton)
     })
-      expect(screen.queryByAltText('Loading indicator')).not.toBeInTheDocument()
+
+    expect(screen.queryByAltText('Loading indicator')).not.toBeInTheDocument()
   })
 })
--- a/frontend/__tests__/src/pages/Chapters.test.tsx
+++ b/frontend/__tests__/src/pages/Chapters.test.tsx
@@ -96,8 +96,6 @@ describe('ChaptersPage Component', () => {
       expect(screen.getByText('Chapter 1')).toBeInTheDocument()
       expect(screen.getByText('Next Page')).toBeInTheDocument()
 
-      const nextPageButton = screen.getByText('Next Page')
-      fireEvent.click(nextPageButton)
     })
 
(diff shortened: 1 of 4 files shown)


yashpandey06:
@arkid15r can you please check if this is resolved your end ?

yashpandey06:
> @arkid15r can you please check if this is resolved at your end ?

@arkid15r please help review these changes !

arkid15r:
> > @arkid15r can you please check if this is resolved at your end ?
> 
> @arkid15r please help review these changes !

Could you resolve the conflicts after #198 merge?

arkid15r (commented):
Looks good!
Is there a way to update tests to make sure search/pagination are not loaded before the main content?

frontend/__tests__/src/pages/Chapters.test.tsx
```diff
@@ -39,6 +39,22 @@ describe('ChaptersPage Component', () => {
     })
   })
 
+  test('renders SearchBar and data concurrently', async () => {
```
arkid15r: Does it check whether search and pagination are not rendered before data is loaded?

My understanding it checks for the loading indicator (we already have such test) and then it checks that search is present.

Could you add the requested checks too?

Thank you!

yashpandey06: I am making some modifications 


as when data is only rendered when it is available in the frontend so in this test search and data are loaded concurrently if the data is not there the search component will also be not there .

for now I am also going to add same thing with pagination 


Which will ensure all three are rendered concurrently only when data is present.


All three would be rendered together which is checked inside single expect block .

yashpandey06: The already added tests are for specific cases like particular to pagination or being particular to render data .

I guess they aren't checking the concurrent appearance of the components.

yashpandey06: Let me check once again !

yashpandey06: @arkid15r 

The recent changes in the tests ensure that the search bar and pagination components are not rendered while the loader is visible initially over the page. Only after the data is successfully fetched, are the pagination and search bar rendered which is checked in the second expect block !

frontend/__tests__/src/pages/Chapters.test.tsx
```diff
@@ -76,4 +76,30 @@ describe('ChaptersPage Component', () => {
...
+      expect(screen.queryByPlaceholderText('Search for OWASP chapters...')).not.toBeInTheDocument()
+      expect(screen.queryByText('Next Page')).not.toBeInTheDocument()
+    })
+    await waitFor(() => {
+      expect(screen.getByPlaceholderText('Search for OWASP chapters...')).toBeInTheDocument()
+      expect(screen.getByText('Chapter 1')).toBeInTheDocument()
+      expect(screen.getByText('Next Page')).toBeInTheDocument()
+
+      const nextPageButton = screen.getByText('Next Page')
+      fireEvent.click(nextPageButton)
```
arkid15r: Why do we need this click?

frontend/__tests__/src/pages/Committees.test.tsx
```diff
@@ -40,6 +40,37 @@ describe('Committees Component', () => {
...
+      expect(screen.queryByText('Next Page')).not.toBeInTheDocument()
+    })
+
+    await waitFor(() => {
+      expect(screen.getByPlaceholderText('Search for OWASP committees...')).toBeInTheDocument()
+
+      expect(screen.getByText('Committee 1')).toBeInTheDocument()
+
+      const nextPageButton = screen.getByText('Next Page')
+      expect(nextPageButton).toBeInTheDocument()
```
arkid15r: Could you make this code be consistent for all pages tests (same syntax for `expects()`)

arkid15r (changes requested):
Yes, that's what we wanted to test. A couple more minor comments:

frontend/__tests__/src/pages/Chapters.test.tsx
yashpandey06: @arkid15r if the pagination component is rendered after the data is loaded ...it should be responsive to the click event right ?

Its just a extra secure check ..should i remove it ?

frontend/__tests__/src/pages/Committees.test.tsx
yashpandey06: sure !

frontend/__tests__/src/pages/Chapters.test.tsx
yashpandey06: for now i am removing it

arkid15r: Checking for presence should be enough. If you fire an event you should check for some expected outcome. Otherwise it's not that obvious (at least to me).

Yes, please remove.

yashpandey06: sure removed it in the latest commits ...thanks

arkid15r (approved):
LGTM! Thank you!

// bumped estimation for the tests efforts

Issue #212, opened 2024-12-19
Fix responsiveness of Footer and card component


Rajgupta36:
## Describe the bug
The **Card** and **Footer** components are not responsive yet. I want to create a new design for the footer that works well on mobile devices.

## Expected Behavior
- The **Card** and **Footer** should be responsive on all screen sizes.
- Add customization to the **Footer** for a better mobile design.

## Screenshots
Current design of the Card and Footer:

![Card Screenshot](https://github.com/user-attachments/assets/a18888a8-2166-4059-afd6-3f1168d7ba42)

![Footer Screenshot](https://github.com/user-attachments/assets/fedd63e0-448a-4335-bc1f-bbc098b39b15)

Tusharjamdade:
Hey @Rajgupta36,
I can work on this. I can make the card and footer responsive for all breakpoints (i.e., mobile, tablet, laptop, and so on). Could you please assign it to me?

Tusharjamdade:
Hi @arkid15r @Rajgupta36,
I have redesigned the Header and Footer. 
I added a hamburger menu to create a dropdown for small screen devices, such as mobile phones. 

Additionally, I redesigned the Footer to make it more responsive for mobile devices. Please let me know if it needs any further improvements. 
If everything looks good, I will proceed to open the PR. Thank you!

https://github.com/user-attachments/assets/d32e9d2d-1169-4b52-9846-9a5070254af7

![image](https://github.com/user-attachments/assets/a15bc554-a922-4aac-80da-dcb0e6e9be8f)

![image](https://github.com/user-attachments/assets/cdd1eb65-ed1b-4c3b-a70b-7b30251f2c88)

Tusharjamdade:
@arkid15r Sorry, what do you mean by "leaders' avatars"? I am not familiar with this term.

arkid15r:
> @arkid15r Sorry, what do you mean by "leaders' avatars"? I am not familiar with this term.

Technically it's contributors' avatars -- images between Leaders and tags blocks that looks squeezed

![image](https://github.com/user-attachments/assets/d2838e10-b6cf-4cd6-a401-042e575c9f1c)

Rajgupta36:
loooks cool @Tusharjamdade . i also created one design  for mobile. can you implement it

https://github.com/user-attachments/assets/3db63838-b8bc-4edf-8c42-08abaec020a3

Tusharjamdade:
[...]

PR #228, opened 2024-12-21
Made header, footer, and card responsive for mobile devices

Tusharjamdade:
Resolves [#212](https://github.com/OWASP/Nest/issues/212)

Add a hamburger icon for mobile devices to open menu options, 
implement a dropdown in the footer for improved responsiveness, 
also fixed the responsiveness for the leaders' avatars.

Here are some screenshots and screen recordings.

image

image

https://github.com/user-attachments/assets/0ae19aae-53df-4deb-9027-50fa9b988984

https://github.com/user-attachments/assets/ab979e79-8af5-49a6-9165-fc7569b3eb46

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,14 @@
 
 Follow these steps to set up the OWASP Nest application:
 
+1. **Fork the Repository**:
+   - Fork <https://github.com/OWASP/Nest> repository using "Fork" button
+
 1. **Clone the Repository**:
-   - Clone the repository code from GitHub using the following command:
+   - Clone the repository code from your GitHub account using the following command:
 
      ```bash
-     git clone https://github.com/owasp/nest
+     git clone https://github.com/<your-account>/<nest-fork>
      ```
 
 1. **Create Environment Files**:
@@ -21,14 +24,21 @@ Follow these steps to set up the OWASP Nest application:
    - Copy the contents from the template file into your new local environment file:
 
      ```bash
-     cp backend/.env/template backend/.env/local
+     cat backend/.env/template > backend/.env/local
      ```
 
    - Create a local environment file in the `frontend` directory:
 
      ```bash
-     cp frontend/.env.example frontend/.env
+     touch frontend/.env
      ```
+
+   - Copy the contents from the template file into your new local environment file:
+
+     ```bash
+     cat frontend/.env.example > frontend/.env
+     ```
+
 
 1. **Configure Environment Variables**:
    - Open the `backend/.env/local` file in your preferred text editor and change the `DJANGO_CONFIGURATION` value to `Local`:
@@ -113,13 +123,19 @@ This troubleshooting section is tailored for Windows users.
 
 The `make run` command requires WSL to function properly. Make sure WSL is installed and configured on your system. If you haven't installed WSL yet, follow [Microsoft's official guide](https://learn.microsoft.com/en-us/windows/wsl/install).
 
-#### 2. Install the `make` Command
+#### 2. Ensure WSL integration is enabled in Docker Desktop settings
+
+Check `Resources -- WSL integration` in Docker application settings.
+
+#### 3. Install the `make` Command
 
 When running the `make` command for the first time, you may be prompted to install the `make` tool. Run the following commands in your WSL terminal to install it:
 
 ```bash
-sudo apt update
-sudo apt install make
+$ sudo apt update
+...
+$ sudo apt install make
+...
 ```
 
 ### Resolving Frontend Permission Issues


.gitignore
arkid15r, line 1: Please revert this

backend/Makefile
arkid15r, line 1: Please revert this

backend/data/nest.json.gz
arkid15r, line 1: Please revert this

arkid15r (changes requested):

frontend/src/components/Card.tsx
```diff
@@ -99,78 +98,76 @@ const Card = ({
...
-                          rel="noopener noreferrer"
-                          className="flex items-center justify-center gap-1"
-                        >
-                          <FontAwesomeIcon icon={item.icon as FontAwesomeIconProps['icon']} />
-                        </a>
-                      ))}
-                  </div>
-                )}
-              </div>
+          {(languages || topics || social) && (
```
kasya: These can be empty arrays, so we'd better check length on them. With this there will be an empty div rendered at the bottom of the card and that does not look good.

frontend/src/components/Card.tsx
```diff
@@ -60,11 +60,10 @@ const Card = ({
...
           <a href={url} target="_blank" rel="noopener noreferrer">
-            <h1 className="text-2xl font-semibold dark:text-sky-600">{title}</h1>
+            <h1 className="text-lg font-semibold sm:text-2xl dark:text-sky-600">{title}</h1>
           </a>
         </div>
-
-        <div className="flex min-w-[30%] flex-wrap items-center justify-end">
+        <div className="flex w-full flex-wrap items-center justify-end gap-2 sm:w-auto">
           {icons &&
             Object.keys(Icons).map((key, index) =>
```
kasya: Also, these are now don't look good on the mobile view. I'm thinking for a mobile device view we should switch to something like a list items instead. 
We can address that in a separate task though! 👍🏼

frontend/src/components/Card.tsx
```diff
@@ -46,9 +46,9 @@ const Card = ({
   }
 
   return (
-    <div className="flex h-fit w-full flex-col items-start justify-normal gap-4 rounded-md border border-border py-6 pl-6 pt-0 md:max-w-6xl md:gap-2">
+    <div className="flex h-fit w-full flex-col items-start gap-4 rounded-md border border-border py-6 pl-4 pt-0 sm:pl-6 md:max-w-6xl md:gap-2">
       <div className="flex w-full flex-wrap items-center justify-between gap-2">
-        <div className="mt-4 flex items-center justify-center gap-2">
+        <div className="mt-4 flex flex-col items-start justify-start gap-2 sm:flex-row sm:items-center">
```
kasya: Can we make The Icon and the Title for the card appear on the same line? 
Screenshot 2024-12-21 at 6 28 04 PM

frontend/src/components/Card.tsx
```diff
@@ -99,78 +98,76 @@ const Card = ({
...
+                      <FontAwesomeIcon icon={item.icon as FontAwesomeIconProps['icon']} />
+                    </a>
+                  ))}
+                </div>
+              )}
             </div>
           )}
         </div>
-        <div className="w-38 content-end justify-items-center pr-6">
+        <div className="mt-4 w-full sm:mt-0 sm:w-auto">
```
kasya: This now doesn't work with the regular screen size. The button should have a right padding that would match the rest of the card.

```suggestion
        <div className="mt-4 w-full sm:mt-0 sm:w-auto mr-6">
```

frontend/src/components/Card.tsx
kasya: Also something I noticed. When there are a lot of labels on mobile view they are sometimes too close to the right border of the card. Could this be updated?
Screenshot 2024-12-21 at 6 32 56 PM

frontend/src/components/Card.tsx
kasya: Also, the button should always be at the top of the page, no matter what labels are there. 
For comparison, same card in current `main` state and with this PR:
Screenshot 2024-12-21 at 6 35 57 PM
Screenshot 2024-12-21 at 6 36 19 PM

kasya (changes requested):

frontend/src/components/Card.tsx
Tusharjamdade: @kasya 
Can we sort the labels/topics based on their length and then arrange them to balance the alignment? This can solve the issue of labels being close to the right border of the card.

Before:
image

After:
image

frontend/src/components/Card.tsx
Tusharjamdade: @kasya 
I am a bit confused about the positioning of the Contribute button. Could you please elaborate a little more on its placement?

These are some of the changes I made on my localhost:


image


image


image

nitinawari:
Hey @Tusharjamdade, for issue #232, I wasn’t aware that you were working on this. You can skip the sticky header part as I have already created a PR for it. Thank you!

PR #230, opened 2024-12-21
Feat : Get indexed items total count from Algolia

### LINKED ISSUE #185: Get indexed items total count from Algolia
For consistency we want to get the total count of indexed items from the source -- Algolia.

Create a one method to determine the total count of items for all entities (issue, project, chapter, committee). Cache the value for better performance. Name the field `total_items`. 

**Additional context**
Clean up the currently used method if they are no longer used.

Rajgupta36:
Resolves #185
---
## Overview

This pull request aims to replace the existing index count logic with the use of the Algolia search client. The updates include a new method to get the count of records in the Algolia index and refactoring the models and search endpoint to accommodate the new logic.

---

## Checklist
- [x] Replaced old method with a new method using the Algolia search client.
- [x] Implemented `get_algolia_index_count` method.
- [x] Updated the search endpoint with the new methods.
- [x] Tested the new search functionality thoroughly.

### DIFF
--- a/backend/apps/common/index.py
+++ b/backend/apps/common/index.py
@@ -17,7 +17,7 @@ class IndexBase:
 
     @staticmethod
     def _get_client():
-        """Reusable method to get the Algolia client."""
+        """Get the Algolia client."""
         return SearchClient.create(
             settings.ALGOLIA_APPLICATION_ID,
             settings.ALGOLIA_API_KEY,
@@ -49,7 +49,7 @@ class IndexBase:
     @staticmethod
     @lru_cache(maxsize=1024)
     def get_total_count(index_name):
-        """Get  total count of records in an Algolia index."""
+        """Get total count of records in index."""
         try:
             index = IndexBase._get_client().init_index(
                 f"{settings.ENVIRONMENT.lower()}_{index_name}"
--- a/backend/apps/common/index.py
+++ b/backend/apps/common/index.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+from functools import lru_cache
 from pathlib import Path
 
 from algoliasearch.search_client import SearchClient
@@ -14,12 +15,19 @@ class IndexSynonymsMixin:
     """Nest index synonyms mixin."""
 
     @staticmethod
-    def reindex_synonyms(app_name, index_name):
-        """Reindex synonyms."""
-        index = SearchClient.create(
+    def _get_algolia_client():
+        """Reusable method to get the Algolia client."""
+        return SearchClient.create(
             settings.ALGOLIA_APPLICATION_ID,
             settings.ALGOLIA_API_KEY,
-        ).init_index(f"{settings.ENVIRONMENT.lower()}_{index_name}")
+        )
+
+    @staticmethod
+    def reindex_synonyms(app_name, index_name):
+        """Reindex synonyms."""
+        index = IndexSynonymsMixin._get_algolia_client().init_index(
+            f"{settings.ENVIRONMENT.lower()}_{index_name}"
+        )
 
         file_path = Path.open(
             f"{settings.BASE_DIR}/apps/{app_name}/index/synonyms/{index_name}.json"
@@ -38,3 +46,15 @@ class IndexSynonymsMixin:
 
         index.clear_synonyms()
         index.save_synonyms(synonyms, {"replaceExistingSynonyms": True})
+
+    @staticmethod
+    @lru_cache(maxsize=1024)
+    def get_algolia_index_count(index_name):
+        """Get the number of records in an Algolia index."""
+        try:
+            index = IndexSynonymsMixin._get_algolia_client().init_index(index_name)
+            return index.search("", {"hitsPerPage": 0, "analytics": False})["nbHits"]
+
+        except Exception:
+            logger.exception("Error retrieving index count for '%s'", index_name)
+            return None
--- a/backend/apps/common/index.py
+++ b/backend/apps/common/index.py
@@ -1,4 +1,4 @@
-"""Algolia index synonyms support."""
+"""Algolia index synonyms support and record count."""
 
 import json
 import logging
@@ -11,11 +11,11 @@ from django.conf import settings
 logger = logging.getLogger(__name__)
 
 
-class IndexSynonymsMixin:
-    """Nest index synonyms mixin."""
+class IndexBase:
+    """Nest index synonyms mixin and record count."""
 
     @staticmethod
-    def _get_algolia_client():
+    def _get_client():
         """Reusable method to get the Algolia client."""
         return SearchClient.create(
             settings.ALGOLIA_APPLICATION_ID,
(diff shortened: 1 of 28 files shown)


backend/apps/common/index.py
```diff
@@ -14,12 +15,19 @@ class IndexSynonymsMixin:
     """Nest index synonyms mixin."""
 
     @staticmethod
-    def reindex_synonyms(app_name, index_name):
-        """Reindex synonyms."""
-        index = SearchClient.create(
+    def _get_algolia_client():
```
arkid15r:
```suggestion
    def _get_client():
```

backend/apps/common/index.py
```diff
@@ -38,3 +46,15 @@ def reindex_synonyms(app_name, index_name):
 
         index.clear_synonyms()
         index.save_synonyms(synonyms, {"replaceExistingSynonyms": True})
+
+    @staticmethod
+    @lru_cache(maxsize=1024)
+    def get_algolia_index_count(index_name):
```
arkid15r:
```suggestion
    def get_total_count(index_name):
```

backend/apps/owasp/api/search/chapter.py
```diff
@@ -4,6 +4,7 @@
 from django.http import JsonResponse
 
 from apps.common.geocoding import get_ip_coordinates
+from apps.common.index import IndexSynonymsMixin
```
arkid15r: This needs to be renamed into `IndexBase` or similar as we extended it beyond synonyms.

backend/apps/owasp/api/search/chapter.py
```diff
@@ -39,11 +40,11 @@ def chapters(request):
     """Search chapters API endpoint."""
     page = int(request.GET.get("page", 1))
     query = request.GET.get("q", "")
-
+    active_chapters_count = IndexSynonymsMixin.get_algolia_index_count("local_chapters")
```
arkid15r: The `local_*` naming will only work for local environment. 
Please update the base class to add environment prefix to the index name and pass only `chapters` (same for other indexed models).

backend/apps/owasp/api/search/chapter.py
```diff
@@ -39,11 +40,11 @@ def chapters(request):
...
     chapters = get_chapters(query=query, page=page, meta=request.META)
     return JsonResponse(
         {
-            "active_chapters_count": Chapter.active_chapters_count(),
+            "active_chapters_count": active_chapters_count,
```
arkid15r: You can inline the value, like:
```suggestion
            "active_chapters_count": IndexBase.get_total_count("chapters")
```

arkid15r (changes requested):
Here is some initial feedback:

backend/apps/common/index.py
```diff
@@ -38,3 +44,17 @@ def reindex_synonyms(app_name, index_name):
...
+    @lru_cache(maxsize=1024)
+    def get_total_count(index_name):
+        """Get  total count of records in an Algolia index."""
+        try:
+            index = IndexBase._get_client().init_index(
+                f"{settings.ENVIRONMENT.lower()}_{index_name}"
+            )
+            return index.search("", {"hitsPerPage": 0, "analytics": False})["nbHits"]
+
+        except Exception:
```
arkid15r: Avoid using too broad exceptions. If you don't need it now just remove it.

backend/apps/github/models/issue.py
```diff
@@ -180,12 +178,6 @@ def bulk_save(issues, fields=None):
         """Bulk save issues."""
         BulkSaveModel.bulk_save(Issue, issues, fields=fields)
 
-    @staticmethod
-    @lru_cache
-    def open_issues_count():
-        """Return open issues count."""
-        return Issue.open_issues.assignable.count()
-
```
arkid15r: As you removed existing method you run tests to check it still works.
The problem is that Nest isn't fully covered with tests. NestBot still uses these methods and need to be updated too. So here you have 2 options:
  - keep the old methods and their signatures while updating their logic with `IndexBase` stuff
  - replace old methods with `IndexBase.get_total_count` everywhere by using search

arkid15r (commented):
Looks better. We also should address:

backend/apps/common/index.py
Rajgupta36: okkk

backend/apps/owasp/models/chapter.py
```diff
@@ -61,12 +59,6 @@ def __str__(self):
         """Chapter human readable representation."""
         return f"{self.name or self.key}"
 
-    @staticmethod
-    @lru_cache
-    def active_chapters_count():
-        """Return active chapters count."""
-        return Chapter.active_chapters.count()
```
arkid15r, line 68: The issues refactoring approach needs to be kept with other models too.

arkid15r (changes requested):
You need to refactor all existing methods to the new IndexBase approach. Keep the methods, change the logic.

backend/apps/owasp/models/chapter.py
Rajgupta36: ohh i didn't notice sorry

backend/apps/owasp/api/search/chapter.py
```diff
@@ -39,11 +40,10 @@ def chapters(request):
     """Search chapters API endpoint."""
     page = int(request.GET.get("page", 1))
     query = request.GET.get("q", "")
-
     chapters = get_chapters(query=query, page=page, meta=request.META)
     return JsonResponse(
         {
-            "active_chapters_count": Chapter.active_chapters_count(),
+            "active_chapters_count": IndexBase.get_total_count("chapters"),
```
arkid15r: It seems you can revert these too.

Rajgupta36: yes

backend/apps/common/index.py
```diff
@@ -1,25 +1,32 @@
...
-    """Nest index synonyms mixin."""
+class IndexBase:
+    """Nest index synonyms mixin and record count."""
 
     @staticmethod
-    def reindex_synonyms(app_name, index_name):
-        """Reindex synonyms."""
-        index = SearchClient.create(
+    def _get_client():
+        """Reusable method to get the Algolia client."""
```
arkid15r:
```suggestion
        """Get the Algolia client."""
```

backend/apps/common/index.py
```diff
@@ -38,3 +45,17 @@ def reindex_synonyms(app_name, index_name):
 
         index.clear_synonyms()
         index.save_synonyms(synonyms, {"replaceExistingSynonyms": True})
+
+    @staticmethod
+    @lru_cache(maxsize=1024)
+    def get_total_count(index_name):
+        """Get  total count of records in an Algolia index."""
```
arkid15r:
```suggestion
        """Get total count of records in index."""
```

arkid15r (approved):
LGTM 👍

PR #219, opened 2024-12-20
Implement Responsive header

harsh3dev:
Resolves #210 

- [x] Implement clean and minimal code by replacing reused code block.
- [x] Added hamburger menu for mobile devices which is expandable for more items.
- [x] Highlight the current active page on the header.
- [x] Fix the overall style and responsiveness of the component.

## Mobile View

https://github.com/user-attachments/assets/63c4f744-2aa4-44a6-a743-3fa260e4ead8

---
## Desktop view
image

PR #252, opened 2024-12-23
Fixed the responsiveness of the header, footer, and card.

### LINKED ISSUE #212: Fix responsiveness of Footer and card component

Tusharjamdade:
Resolves #212 

Added a hamburger icon for mobile devices to open menu options, implemented a dropdown in the footer for better responsiveness, and fixed the responsiveness of the leaders' avatars. Additionally, made the header sticky at the top and fixed the alignment of the summary and labels, which were aligned to the right in mobile view

image

image


Tusharjamdade:
@arkid15r @kasya, please verify the changes.

Thank you!

Tusharjamdade:
@arkid15r, Please review the changes!

frontend/src/components/Footer.tsx
kasya, line 1: I like this folded view for mobile devices! 👍🏼

frontend/src/components/Header.tsx
kasya, line 1: Can we update the header to have the same layout as before? The logo and Nest text (as well as the theme color) is now not at the ends of the div. Also, the menu optons (nav) should be on the side and not centered.
I also feel like the `Nest` text got bolder 🤔 Can we check that?
Screenshot 2024-12-25 at 12 34 06 PM
Screenshot 2024-12-25 at 12 34 13 PM

frontend/src/components/Card.tsx
kasya, line 1: Also, are you sure your PR is up to date? I believe we already updated the Card component for contribute page to have Project name above the summary text. As well as some other details, like placement of the button. Could you make sure it's updated? 

Screenshot 2024-12-25 at 12 37 26 PM

kasya (changes requested):
Nice work!
Couple of things I noticed.
There's this huge gap now above the search component. It's noticeable even on regular page, but on the mobile - it's just crazy big. Can we decrease that to be as it was before?

Some other things:

Tusharjamdade:
> Nice work! Couple of things I noticed. There's this huge gap now above the search component. It's noticeable even on regular page, but on the mobile - it's just crazy big. Can we decrease that to be as it was before?
> 
> Some other things:

@kasya, if possible, could you please share a screenshot of it? I am not seeing this on my localhost.

Tusharjamdade:
@kasya, do we need to keep the button next to the labels? It might not look good when there are many labels in the mobile view.


image

Similarly to the Projects page:


image


Alternatively, we could arrange it like this:


image

kasya:
> > Nice work! Couple of things I noticed. There's this huge gap now above the search component. It's noticeable even on regular page, but on the mobile - it's just crazy big. Can we decrease that to be as it was before?
> > Some other things:
> 
> @kasya, if possible, could you please share a screenshot of it? I am not seeing this on my localhost.

of course!! I was talking about this gap between header and the search component. It's there both on mobile and desctop view now and it wasn't there before:

Screenshot 2024-12-26 at 9 12 29 AM

kasya:
> @kasya, do we need to keep the button next to the labels? It might not look good when there are many labels in the mobile view.

> Alternatively, we could arrange it like this:
> 
> image

I do like this second option more, but only if we are able to keep desktop card as is, where the button is always at the bottom of the card. 
Another option - we could also decrease a little bit the number of shown labels for mobile view to make it look better with the button next to it. So there are a couple of ways to address this.. What do you think?

Tusharjamdade:
> > @kasya, do we need to keep the button next to the labels? It might not look good when there are many labels in the mobile view.
> 
> > Alternatively, we could arrange it like this:
> > image
> 
> I do like this second option more, but only if we are able to keep desktop card as is, where the button is always at the bottom of the card. Another option - we could also decrease a little bit the number of shown labels for mobile view to make it look better with the button next to it. So there are a couple of ways to address this.. What do you think?

@kasya, we can reduce the number of labels for mobile, but when we expand it (by clicking on the 'Show More' option), the problem still exists, in my opinion. So, which design should we choose? Additionally, I have fixed the header as per the suggestion. The only remaining part is the cards.

harsh3dev:
@Tusharjamdade I have implemented the Responsive header #219  with additional functionality.
I had discussed with @kasya 
We can take that header and Footer and Card from your PR as your implemented designs are amazing for the Footer.
Have you worked on the changes requested on this PR? So that we can continue.

Tusharjamdade:
> @Tusharjamdade I have implemented the Responsive header #219 with additional functionality. I had discussed with @kasya We can take that header and Footer and Card from your PR as your implemented designs are amazing for the Footer. Have you worked on the changes requested on this PR? So that we can continue.


@harsh3dev, I have completed the header and footer. The only remaining part is the card. Since you are working on the responsive header as well, as suggested by @kasya, we can take the header from your PR and the footer and card from my PR.

kasya:
@Tusharjamdade I don't think you've updated the PR regarding my last comments yet 🤔 I can still see that big gap on the page above the search component. 
As fore the card labels layout - I think it makes sense if the button get's pushed down when the user clicks "show more" link. That's kind of expected.

frontend/src/components/DisplayIcon.tsx
```diff
@@ -18,7 +16,7 @@ export default function DisplayIcon({
     <div
       data-tooltip-id={`icon-tooltip-${item}`}
       data-tooltip-content={`${Icons[item as keyof typeof Icons]?.label}`}
-      className={`flex flex-col items-center justify-center gap-1 border-b border-l border-border px-4 pb-1 ${idx == 0 ? 'rounded-bl-md' : ''}`}
+      className={`flex flex-col items-center justify-center gap-1 border border-border px-4 pb-1 ${idx == 0 ? 'rounded-bl-md' : ''}`}
```
kasya: I understand that you added this to accomodate the mobile view to have the border, but we didn't put a top border here, so that there's no double border on the desktop view. Right now, it's pretty noticeable when we view it in desktop mode:
Screenshot 2024-12-29 at 2 44 36 PM
Screenshot 2024-12-29 at 2 44 54 PM

Also, as you can see there was a bottom left corner radius added. This was also only intended for mobile view to match with a top right corner of the card. 
If we just move this block down under the title - it does look weird with that corner being rounded 🤔 

My original idea was to show these as blocks with borders only for desktop view, and then show it somehow differently (as a list as an option) for mobile view. That might be too much to address within this PR since I would I like to merge it in at least for the footer changes. 

If you could just update this to only show rounded  bottom-left corner and top-border for mobile devices - this should work for now! We can address the rest in a different task.
What do you think?

frontend/src/components/DisplayIcon.tsx
kasya, line 1: I also pushed some changes that should help with card meta-data blocks getting resized on mobile view when the number were too large. I added a  package to turn those into human readable view (1500 -> 1.5k, etc).

frontend/src/components/Card.tsx
```diff
@@ -28,165 +27,182 @@ const Card = ({
...
+  const [toggleTopics, setToggleTopics] = useState(false);
+  const [isMobile, setIsMobile] = useState(false);
 
-  const [toggleLanguages, setToggleLanguages] = useState(true)
-  const [toggleTopics, setToggleTopics] = useState(true)
+  useEffect(() => {
+    const checkMobile = () => {
+      const mobile = window.innerWidth < 768;
+      setIsMobile(mobile);
+      setVisibleLanguages(mobile ? 4 : 18);
```
kasya: For some reason this is not working as it supposed to. I attached a video recording that shows what I mean. When I do hard-refresh on the page it takes several seconds for this to apply and reduce the number of labels on the card. This is really confusing and it shouldn't be like this for the end user. Can we figure out what could be changed here?

https://github.com/user-attachments/assets/792cc3b1-9a84-4031-876c-0c200ae03465

kasya (changes requested):
Nice updates! Left some comments and also pushed small changes to help address some of those!
Also, since we merged in the Responsive header PR there are conflicts now that need to be resolved! 👌🏼

kasya:
> @kasya, I have found a way to reduce the delay, but the overall approach of reducing the number of labels for mobile (i.e., 4 for mobile) remains the same. Should this be fine, or do I need to change the approach?

@Tusharjamdade  I don't see any pushed changes regarding that 🤔 . Only that you've updated your branch with *main* changes. However, right now, when I check out your branch I can not run it locally due to a incorrect resolved conflicts. The `Header` component file now has duplicate lines for exporting default component. Could you please fix that? As well as address my previous comments. Thanks!

Tusharjamdade:
@

> > @kasya, I have found a way to reduce the delay, but the overall approach of reducing the number of labels for mobile (i.e., 4 for mobile) remains the same. Should this be fine, or do I need to change the approach?
> 
> @Tusharjamdade I don't see any pushed changes regarding that 🤔 . Only that you've updated your branch with _main_ changes. However, right now, when I check out your branch I can not run it locally due to a incorrect resolved conflicts. The `Header` component file now has duplicate lines for exporting default component. Could you please fix that? As well as address my previous comments. Thanks!

@kasya, apologies for that. I have now committed the changes to fix the label delay when the page refreshes, and also addressed the double border issue on desktop. 
Please review the changes and let me know if any further improvements are needed.
Thank You!!

kasya:
Hey @Tusharjamdade, thanks for the updates! I reviewed the card and found a few more issues.
One was with classes for mobile devices. With [tailwind](https://tailwindcss.com/docs/responsive-design#targeting-mobile-screens), the classes without the prefix applied to all screen sizes and classes with prefixes (like `sm:`) will only apply to the class mentioned and above. So those were not applying properly on meta data borders.
```
Where this approach surprises people most often is that to style something for mobile, 
you need to use the unprefixed version of a utility, not the sm: prefixed version. Don’t think of 
sm: as meaning “on small screens”,  think of it as “at the small breakpoint“.
```
I pushed changes to address that. Also, there were a conflict with the Footer that needed to be resolved. Also did that.

And lastly, there were many small updates throughout the files, like extra semi-colons etc. Whenever you do that, could you please run `make check` before pushing your changes? This is something we have mentioned in our ["DOs and DON'Ts"](https://owasp.slack.com/docs/T04T40NHX/F086JCSHLGN). This way our linter will update anything that needs to be updated. With your last push CI\CD was failing due to pre-commit fail.

I updated everything and this looks good to merge now! Thank you!

kasya (approved):

Issue #193, opened 2024-12-17
Increase slackbot test coverage


arkid15r:
We want to have it at 75%+ level.

Commands:
```bash
apps/slack/commands/contribute.py                                                  35     22     10      1    31%   20-92, 96
apps/slack/commands/gsoc.py                                                        19     11      6      1    36%   15-37, 41
apps/slack/commands/projects.py                                                    34     22      8      1    31%   19-104, 108
```

Event handlers:
```bash
apps/slack/events/user_joined_channel/contribute.py                                18     11      4      1    36%   18-28, 66-70
apps/slack/events/user_joined_channel/gsoc.py                                      16     10      4      1    35%   12-21, 42-46
```

PR #256, opened 2024-12-24
Increased Coverage for Slack Management files

Yashgupta9330:
Resolves https://github.com/OWASP/Nest/issues/193

 I have added test files for the slack/management/commands and commands and events files, which have increased the test coverage to 52%.

image
image


backend/apps/common/utils.py
```diff
@@ -34,7 +34,11 @@ def join_values(fields, delimiter=" "):
 
 def natural_date(value):
     """Return humanized version of a date."""
-    return naturaltime(datetime.fromtimestamp(value, tz=timezone.utc))
+    if isinstance(value, str):
+        value = datetime.strptime(value, "%Y-%m-%d").replace(tzinfo=timezone.utc)
+    elif isinstance(value, int):
+        value = datetime.fromtimestamp(value, tz=timezone.utc)
+    return naturaltime(value)
```
arkid15r, line 41: Could you add tests for this update too?

backend/tests/slack/commands/projects_tests.py
```diff
@@ -0,0 +1,107 @@
...
+def test_handler_with_results(
+    mock_get_projects, mock_slack_client, mock_slack_command, mock_project
+):
+    mock_get_projects.return_value = {"hits": [mock_project]}
+    settings.SLACK_COMMANDS_ENABLED = True
+
+    handler(ack=MagicMock(), command=mock_slack_command, client=mock_slack_client)
+
+    blocks = mock_slack_client.chat_postMessage.call_args[1]["blocks"]
+    assert any("Here are top 10 most OWASP projects" in str(block) for block in blocks)
```
arkid15r: Could you fix the wording here:
"Here are top 10 most relevant OWASP projects"

backend/tests/slack/events/user_joined_channel/contribute_tests.py
```diff
@@ -0,0 +1,69 @@
...
+        "apps.slack.events.user_joined_channel.contribute", fromlist=["contribute_handler"]
+    )
+    check_contribute_handler = getattr(
+        contribute_module,
+        "check_contribute_handler",
+        lambda x: x.get("channel") == OWASP_CONTRIBUTE_CHANNEL_ID,
+    )
+
+    assert check_contribute_handler({"channel": OWASP_CONTRIBUTE_CHANNEL_ID}) is True
+    assert check_contribute_handler({"channel": "C999999"}) is False
```
arkid15r:
```suggestion
    assert check_contribute_handler({"channel": OWASP_CONTRIBUTE_CHANNEL_ID})
    assert not check_contribute_handler({"channel": "C999999"})
```

arkid15r (changes requested):
Let's make tests part of a class instead of just functions

Also let's use asserts w/o `is True` and `is False` -- just `assert` and `assert not`

backend/tests/slack/commands/projects_tests.py
Yashgupta9330: but in projects.py Here are top 10 most OWASP projects is used

PR #270, opened 2024-12-25
Added Slackbot over HTTPS endpoints

### LINKED ISSUE #241: Implement NestBot endpoints to support events handling horizontal scaling
Currently NestBot uses Socket mode for all environments. It's really convenient for development environment due to simplified bot events handler run locally. However, the Socket mode approach has its drawbacks in terms of scalability (https://api.slack.com/apis/event-delivery)

Screenshot 2024-12-22 at 11 42 12 AM

Create new OWASP Nest endpoints that would replicate (reuse the code) existing Slack slash commands and event handlers.
We still want use Socket mode for local environment.

**Additional context**
  - https://api.slack.com/concepts
  - https://api.slack.com/apis/events-api
  - https://api.slack.com/apis/event-delivery

ShashaankS:
Resolves #241 

- [x] Added OWASP Nest endpoint for slack api communication
- [x] Provision for websocket use when slackbot configured over local environment

Screenshot from 2024-12-25 20-47-23


backend/apps/slack/views.py
```diff
@@ -0,0 +1,13 @@
+from django.views.decorators.csrf import csrf_exempt  # noqa: D100
```
arkid15r: Please remove the rule suppression and add the module name.

backend/apps/slack/views.py
```diff
@@ -0,0 +1,13 @@
...
+
+from apps.slack.apps import SlackConfig
+from apps.slack.commands import *  # noqa: F403
+from apps.slack.events import *  # noqa: F403
+
+slack_handler = SlackRequestHandler(SlackConfig.app)
+
+
+@csrf_exempt
+def slack_events(request):  # noqa: D103
```
arkid15r: Same here -- add method docstring (see other modules for an example).

arkid15r (changes requested):
Looks good and works for me in general.
I wonder how you were able to test it w/o `url_verificaion` handler?

Please look into this when you get a chance:

ShashaankS:
For URL verification, I just tried adding the the URL to the Slack API and it got verified and the same can be seen on the incoming POST request.
Moreover into the SlackRequestHandler documentation it didn't asked to do it, I huess it has some built-in handler for it.

arkid15r (approved):
LGTM 👍

PR #271, opened 2024-12-25
Fix : SearchBar unexpected behaviour

Rajgupta36:
- impleted the debouncing cleanup logic
- increase the debouncing timeout from 500ms to 750ms


arkid15r (approved):
Works for me, thanks for fixing this!

PR #336, opened 2025-01-03
Users page Frontend

### LINKED ISSUE #68: Implement Users page
We have a GitHub [user entity](https://github.com/OWASP/Nest/blob/ad5a48f650b119cb30e40ec99ff926fa39d29af4/backend/apps/github/models/user.py#L9) model that's currently used in Projects page only. 

The idea is to create a standalone Users page that will be responsible for showing/searching users within GitHub OWASP organization. Each user entry will be clickable and should lead to another view showing a detailed user page (this is going to be covered in another issue).

The task involves both backend and frontend changes. Currently we don't have user index in Algolia. So as the first step for backend we need to index users (we have plenty of examples how to do that, e.g. [issues](https://github.com/OWASP/Nest/blob/ad5a48f650b119cb30e40ec99ff926fa39d29af4/backend/apps/github/index/issue.py#L10), [projects](https://github.com/OWASP/Nest/blob/ad5a48f650b119cb30e40ec99ff926fa39d29af4/backend/apps/owasp/index/project.py#L10), [chapters](https://github.com/OWASP/Nest/blob/ad5a48f650b119cb30e40ec99ff926fa39d29af4/backend/apps/owasp/index/chapter.py#L10)).

After creating the users index the data needs to be exposed via users search endpoint (it also needs to be implemented using existing examples, e.g. [project](https://github.com/OWASP/Nest/blob/ad5a48f650b119cb30e40ec99ff926fa39d29af4/backend/apps/owasp/api/search/project.py#L13)).

The frontend part of the task is to create a component for user entry and use it for populating Users page. After the [pagination component](https://github.com/OWASP/Nest/issues/56) is ready it'll be added to the users page as well.

Yashgupta9330:
resolves #68

### DIFF
--- a/frontend/src/pages/Users.tsx
+++ b/frontend/src/pages/Users.tsx
@@ -1,0 +1,41 @@
+import { useSearchPage } from 'lib/hooks/useSearchPage'
+import { user } from 'lib/types'
+import SearchPageLayout from 'components/SearchPageLayout'
+
+const UsersPage = () => {
+  const {
+    items: users,
+    isLoaded,
+    currentPage,
+    totalPages,
+    searchQuery,
+    handleSearch,
+    handlePageChange,
+  } = useSearchPage<user>({
+    indexName: 'users',
+    pageTitle: 'OWASP Users',
+  })
+
+  const renderUserCard = (user: user, index: number) => {
+    return (
+       <div key={index}>{user.idx_name}</div>
+    )
+  }
+
+  return (
+    <SearchPageLayout
+      isLoaded={isLoaded}
+      totalPages={totalPages}
+      currentPage={currentPage}
+      searchQuery={searchQuery}
+      onSearch={handleSearch}
+      onPageChange={handlePageChange}
+      empty="No Users found"
+      searchPlaceholder="Search for OWASP users..."
+    >
+      {users && users.map(renderUserCard)}
+    </SearchPageLayout>
+  )
+}
+
+export default UsersPage
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -13,6 +13,7 @@ import { Routes, Route, useLocation } from 'react-router-dom'
 
 import Footer from 'components/Footer'
 import Header from 'components/Header'
+import UsersPage from 'pages/Users'
 
 function App() {
   const location = useLocation()
@@ -33,6 +34,7 @@ function App() {
         <Route path="/committees/:committeeKey" element={<CommitteeDetailsPage />}></Route>
         <Route path="/chapters" element={<ChaptersPage />}></Route>
         <Route path="/chapters/:chapterKey" element={<ChapterDetailsPage />}></Route>
+        <Route path="/users" element={<UsersPage />}></Route>
       </Routes>
       <Footer />
     </main>
--- a/frontend/src/lib/types.ts
+++ b/frontend/src/lib/types.ts
@@ -4,6 +4,26 @@ export interface ProjectDataType {
   total_pages: number
 }
 
+
+export type user= {
+  idx_email: string
+  idx_login: string
+  idx_name: string
+  idx_company: string
+  idx_location: string
+  idx_bio: string
+  idx_followers_count: number
+  idx_created_at: number
+  idx_avatar_url: string
+  idx_following_count: number
+  idx_key: string
+  idx_public_repositories_count: number
+  idx_title: string
+  idx_updated_at: number
+  idx_url: string
+  objectID: string
+};
+
 export type project = {
   idx_top_contributors: {
     avatar_url: string


frontend/src/components/Usercard.tsx
arkid15r, line 1: Rename to `UserCard.tsx`

arkid15r, line 1: Could you add dark theme support?

frontend/src/components/Usercard.tsx
```diff
@@ -0,0 +1,26 @@
...
+      <div className="mt-auto">
+        <button
+          onClick={button.onclick}
+          className="flex items-center rounded-md bg-blue-500 px-4 py-2 text-white"
+        >
+          {button.icon}
+          <span className="ml-2">{button.label}</span>
+        </button>
+      </div>
+    </div>
```
arkid15r: Let's make the entire card clickable -- there are too many buttons.

frontend/src/utils/constants.ts
```diff
@@ -21,6 +21,10 @@ export const headerLinks: Link[] = [
     text: 'Committees',
     href: '/committees',
   },
+  {
+    text: 'Users',
+    href: '/users',
```
arkid15r:
```suggestion
    text: 'Community',
    href: '/community/users',
```

frontend/src/utils/paramsMapping.ts
```diff
@@ -72,6 +72,23 @@ export const getParamsForIndexName = (indexName: string, distinct = false) => {
...
+          'idx_email',
+          'idx_name',
+          'idx_followers_count',
+          'idx_company',
+          'idx_login',
+          'idx_updated_at',
+          'idx_url',
+          'idx_avatar_url',
+        ],
+        minProximity: 4,
```
arkid15r, line 96:
```suggestion
        hitsPerPage: 24,
        minProximity: 4,
```

frontend/src/utils/paramsMapping.ts
```diff
@@ -72,6 +72,23 @@ export const getParamsForIndexName = (indexName: string, distinct = false) => {
...
+          'idx_created_at',
+          'idx_key',
+          'idx_email',
+          'idx_name',
+          'idx_followers_count',
+          'idx_company',
+          'idx_login',
+          'idx_updated_at',
+          'idx_url',
+          'idx_avatar_url',
```
arkid15r: Keep only used attributes.

arkid15r (changes requested):
This works for me, great job!
Let's improve at a bit when you get a chance:

frontend/src/components/Usercard.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <div className="my-2 flex h-64 w-64 flex-col items-center justify-center overflow-hidden rounded-lg bg-white p-4 shadow-md">
+      <div className="mb-4 h-16 w-16">
+        <img src={avatar} alt={`${name} Avatar`} className="h-16 w-16 rounded-full object-cover" />
+      </div>
+      <div className="text-center">
+        <h3 className="text-xl font-semibold">{name}</h3>
+        <p className="text-gray-600">{company}</p>
+      </div>
+      <div className="mt-auto">
+        <button
```
kasya: If we are getting rid of this button completely - can we add some hover effect to the card? Not too aggressive though. Something like a small transform to y-4px or something like that. With a pointer cursor it should make it obvious it's clickable 🤔 

If we decide to leave it here - can we make this button the same style as the button on the card? So far all the buttons that we have only have outline and get filled with color on hover. Something like this:
Screenshot 2025-01-05 at 5 28 20 PM
Screenshot 2025-01-05 at 5 28 24 PM

frontend/src/pages/Users.tsx
```diff
@@ -0,0 +1,63 @@
...
+      isLoaded={isLoaded}
+      totalPages={totalPages}
+      currentPage={currentPage}
+      searchQuery={searchQuery}
+      onSearch={handleSearch}
+      onPageChange={handlePageChange}
+      empty="No Users found"
+      searchPlaceholder="Search for OWASP users..."
+    >
+      <div className="grid grid-cols-1 gap-6 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4">
```
kasya: It would be nice to make these responsive right from the start. Right now on iPadPro or iPad Air view there's no padding between cards. Also, there's no padding on the right side of the page on that view. 🤔
Can we make sure all major sizes are fully covered with these classes?
Some examples of what I'm seeing right now:
Screenshot 2025-01-05 at 5 35 51 PM
Screenshot 2025-01-05 at 5 35 57 PM

frontend/src/pages/Users.tsx
kasya, line 1: The page shows 25 items per page right now. This works nice for other pages sine we show them as a list. With this layout 25 pages look weird since the last row has only one item. 

Can we show total number of cards that is divided by 4? Or better yet - change the layout to have 5 cards per row (which would be easier to implement)?


Screenshot 2025-01-05 at 5 39 28 PM

kasya (changes requested):

kasya (commented):
@Yashgupta9330  Could you also check the issue with CI\CD failing? It seems that it cannot find the component after some of the changes were made.

arkid15r (dismissed):
Let's try it!

LGTM 👍

kasya (approved):

PR #343, opened 2025-01-05
Fixed show more link breaks card labels

### LINKED ISSUE #333: Show more link breaks card labels
**Describe the bug**
When we are on the Contribute page the card should show `Show more` when there are more labels available.
We have 2 types of labels - for languages and for topics. We should only show one `Show more` link if any of those have more items to show, but we should not show it after each type. 
[Example](http://localhost:3000/projects/contribute?q=use+proxy) (first card). Also, clicking `Show more` doesn't really do anything in this case.

**Expected behavior**
We should only show `Show more` after all the labels if any of those have anything extra to show.

**Screenshots**
Screenshot 2025-01-02 at 10 18 49 AM

Naveen-Pal:
Resolves #333


frontend/src/components/Card.tsx
```diff
@@ -179,12 +171,15 @@ const Card = ({
                     type="topic"
                   />
                 ))}
-                {topics.length > 18 && (
+                {(languages?.length > 8 || topics?.length > 18) && (
```
kasya: I believe we only show 4 items on the mobile view for each - topics and languages. So if the card has 8 languages and 1 topic, on a desktop view it should not show the link, however on the mobile view - the link should be there as soon as we have more then 4 items. 
However it doesn't not work like that right now for the desktop view. Here's an [example](http://localhost:3000/projects/contribute?q=kube-hunter+is+not+under+active+development+please+use+another+tool+like+KBOM):
Screenshot 2025-01-05 at 5 56 29 PM

The show more link is still present even though all languages are shown and the topic has only one item. 
Also, since there's nothing else to show - the button click doesn't really change anything. It works as expected on the mobile view though,
Screenshot 2025-01-05 at 6 04 47 PM
Screenshot 2025-01-05 at 5 58 25 PM

kasya (changes requested):
Thanks for working on this! Something I noticed still:

Naveen-Pal:
@kasya . Requesting to please review this PR

kasya (approved):
This is perfect. Thanks!

PR #382, opened 2025-01-08
Added Issue Template: Slack

### LINKED ISSUE #381: Add Issue Template: Slack Link
### Add a field in the issues template to prompt users to check if their issue is better suited for discussion rather than creating an issue. If so, they'll be redirected to the OWASP/Nest Slack channel for more collaborative conversations.

abhayymishraa:
Resolves #381 

Add the PR description here.
This PR added a ISSUE_TEMPLATE  that will be redirecting user to the slack.
Screenshot from 2025-01-09 02-55-57


.github/ISSUE_TEMPLATE/slack_config.yml
arkid15r, line 1: Can it be just `config.yml`?

abhayymishraa: yes it can be for now but in future if we want to implement more issue_template then it will conflict with them

arkid15r: Okay, let's use `contact_links.yml` then

abhayymishraa: done

arkid15r (approved):
Let's try it!

.github/ISSUE_TEMPLATE/slack_config.yml
abhayymishraa: yeah @arkid15r  i found the bug actually if we define any config which will not create issue it needs to be ```config.yml``` i thought file name wouldn't matter

PR #397, opened 2025-01-09
Created a static get_logins and updated  User::is_indexable and User:…

### LINKED ISSUE #359: Exclude organizations logins from users index
We have 2 separate entities for users and organizations. The problem comes from the users entities contain organization logins.

  - Create a static `Organization::get_logins` method
  - Update [User::is_indexable](https://github.com/OWASP/Nest/blob/main/backend/apps/github/models/user.py#L26) and [User::get_queryset](https://github.com/OWASP/Nest/blob/1961b919f97f93e781476840af571b7cafb2e001/backend/apps/github/index/user.py#L59) method to exclude organizations' logins.

abhayymishraa:
Resolves #359 

Add the PR description here.
I have implemented the logic that is asked in the issue also i have maded some of the testcases as testcases is something that would go out of my mind so i maded basis of it ofcourse it can be implemented more further

### DIFF
--- a/backend/apps/github/models/organization.py
+++ b/backend/apps/github/models/organization.py
@@ -41,6 +41,11 @@ class Organization(
                 setattr(self, model_field, value)
 
     @staticmethod
+    def get_logins():
+        """Retrieve all organization logins """
+        return list(Organization.objects.values_list("name", flat=True))
+
+    @staticmethod
     def bulk_save(organizations):
         """Bulk save organizations."""
         BulkSaveModel.bulk_save(Organization, organizations)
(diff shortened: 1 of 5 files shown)


backend/apps/github/models/organization.py
```diff
@@ -40,6 +40,11 @@ def from_github(self, gh_organization):
             if value is not None:
                 setattr(self, model_field, value)
 
+    @staticmethod
+    def get_logins():
+        """Retrieve all organization logins."""
+        return list(Organization.objects.values_list("name", flat=True))
```
arkid15r:
```suggestion
        return set(Organization.objects.values_list("login", flat=True))
```

arkid15r (changes requested):
Sometimes it doesn't work right even with tests :)

arkid15r (commented):
Other than that LGTM, let's merge it after fixing the comments

abhayymishraa:
i don't know it's working in local though

arkid15r (approved):
Merging this, thanks @abhayymishraa 👍

PR #417, opened 2025-01-10
Refactor frontend file structure 

### LINKED ISSUE #400: Refactor Project Structure
## Description

The current folder and file structure of the project could benefit from a refactor to align with standard industry practices. This will improve maintainability, scalability, and overall developer experience by ensuring that files are well-organized and easier to navigate.

### Proposed Changes
1. **Organize the directory.
2. **Separate concerns clearly** (e.g., `components`, `utils`, `services`, `tests`, etc.).
3. Adopt naming conventions that are consistent and intuitive.
4. Align with best practices

### Goals
- Improve code readability.
- Simplify onboarding for new contributors.
- Lay a strong foundation for future scalability.

### Task Checklist
- [ ] Audit the current file/folder structure.
- [ ] Propose a new structure in the issue comments.
- [ ] Move files to their new locations, ensuring paths are updated across the project.
- [ ] Verify the application works as expected after the changes.

abhayymishraa:
Resolves #400


frontend/src/components/Modal.tsx
```diff
@@ -3,9 +3,9 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import React, { useEffect, useRef } from 'react'
 import { createPortal } from 'react-dom'
 
+import { ModalProps } from 'types/modal'
+import { useModal } from 'lib/hooks/useModal'
```
arkid15r: Could you move hooks to a top level directory?

Maybe also rename `lib` to `helpers`?

abhayymishraa: I believe "lib" is the standard name  we usually use in react , but I can change it to "helpers" if you'd prefer.

arkid15r: I've never seen that (but it's just me I think). I'd use either `helpers` or `utils`.
Let's ask other folks opinions?

abhayymishraa: Sure, let’s see! Based on my research, we can use both "helper" and "lib"  but I believe sticking with "helper" would be more justifiable given the type of files it will contain. It aligns better with the intended purpose of the directory.

frontend/src/helpers/algoliaClient.ts
```diff
@@ -1,6 +1,6 @@
 import { algoliasearch } from 'algoliasearch'
```
yashpandey06: As per industry standards , we should either have helpers directory or utils  one ...utils are supposed to have all helpers/utility functions ...please put all the helpers related thing in utils directory to avoid function.

frontend/src/helpers/algoliaClient.ts
```diff
@@ -1,6 +1,6 @@
...
+import { AppError, handleAppError } from 'helpers/wrappers/ErrorWrapper'
 import { ALGOLIA_APP_ID, ALGOLIA_SEARCH_API_KEY } from 'utils/credentials'
-import { AppError, handleAppError } from 'lib/ErrorWrapper'
 
 export const createAlgoliaClient = () => {
```
yashpandey06: Utils  should never have business logic of the application it should only have common helper type functions ...you can make one directory named `api` and put fetching and pai related things there...(just a suggestion).

frontend/src/components/ui/Toaster.tsx
```diff
@@ -1,4 +1,4 @@
-import { useToast } from 'lib/hooks/useToast'
```
yashpandey06, line 1: this is also a component right ...? would suggest to move  it in component folder because in future i don't think we will have more additions to it because the models and everything will placed in components in future (if added).

frontend/src/pages/ChapterDetails.tsx
```diff
@@ -1,8 +1,8 @@
+import { fetchAlgoliaData } from 'helpers/api'
+import FontAwesomeIconWrapper from 'helpers/wrappers/FontAwesomeIconWrapper'
 import { useEffect, useState } from 'react'
```
yashpandey06, line 2: below one is the better approach ... do like this ...



/pages
├── /Chapters
├── /Project
├── /Community
└── /Users


then put relevant stuffs into it ... so let's say if we add anything specific page that will be added in the that particular directory .

yashpandey06 (changes requested):

frontend/src/helpers/algoliaClient.ts
abhayymishraa: Yeah, yeah i think we can do that , great idea!!!!!

harsh3dev (changes requested):
That's a lot of effort! 💯 
I suggest that we make folders for each component such as `Card/Card.tsx` as it is better to organise it that way.
Also when more of the components are added the structure won't get messy.

nitinawari:
hey @abhayymishraa great work ! just add one suggestion suggested by yash and good to go 👍

nitinawari (approved):
**LGTM** 👍

frontend/src/components/ui/Toaster.tsx
Rajgupta36: I think it's a `ShadCN `component, so it would be better to keep it in a separate place. Correct me if I'm wrong. This makes it easier for contributors to identify ShadCN components, and it also helps ShadCN recognize these components more easily. Otherwise, ` (e.g., card.tsx)`, it might override the existing component

Rajgupta36 (changes requested):
I think @yashpandey06's suggestion is good.

Also, one points from my side:

1. api.ts contains two functions. If we are creating an api folder, splitting the functions for better modularity would be a good idea.

frontend/src/helpers/algoliaClient.ts
abhayymishraa: Yeah  i think we can do that !!! Great idea!! to have api calls at one place

frontend/src/components/ui/Toaster.tsx
abhayymishraa: I agree with Raj's suggestion to keep this as a separate ShadCN component. Keeping it in a dedicated location helps with better organization and makes it easier for contributors to identify and work with ShadCN-specific components. It also avoids potential conflicts with existing components (like ```card.tsx```) and ensures integration with ShadCN in the future.

yashpandey06: Great !

frontend/src/pages/ChapterDetails.tsx
abhayymishraa: Thanks for your suggestion, I think this approach works really well for Next.js projects since file-based routing is supported out of the box. Having all components defined by the route in the pages folder definitely keeps things organized in that context.

However, for a project that relies on explicit route definitions (like in a react-router-dom setup), creating a folder for each route might not be the most optimal choice. It could introduce unnecessary complexity since routes are defined in code, not inferred from the file structure like in Next.js.

yashpandey06: I am not sure but it's just the addition of directories for a clean view nothing like the complexity concept ...

abhayymishraa:
@harsh3dev  Thanks for the suggestion! While creating folders for each component can work in some cases, I think it might lead to unnecessary folder clutter, especially for smaller components. Since we already have a dedicated ```components``` folder, it seems more practical to keep everything organized there.

What do you think? 😊

abhayymishraa:
> I think @yashpandey06's suggestion is good.
> 
> Also, one points from my side:
> 
> 1. api.ts contains two functions. If we are creating an api folder, splitting the functions for better modularity would be a good idea.

Great idea!!!! I agree with this

frontend/src/pages/ChapterDetails.tsx
abhayymishraa: yeah Let's have more suggestion for this!!!!

frontend/__tests__/src/pages/ChapterDetails.test.tsx
```diff
@@ -1,12 +1,12 @@
 import { screen, waitFor } from '@testing-library/react'
 
+import { fetchAlgoliaData } from 'helpers/api'
+import { render } from 'helpers/wrappers/test-util'
```
arkid15r: Should we change `test-util` to camel case?

frontend/src/components/Card.tsx
```diff
@@ -1,12 +1,12 @@
 import { FontAwesomeIcon, FontAwesomeIconProps } from '@fortawesome/react-fontawesome'
+import { CardProps, tooltipStyle } from 'helpers/constants'
+import FontAwesomeIconWrapper from 'helpers/wrappers/FontAwesomeIconWrapper'
 import { useState, useEffect } from 'react'
 import { Tooltip } from 'react-tooltip'
-import { CardProps, tooltipStyle } from 'lib/constants'
-import FontAwesomeIconWrapper from 'lib/FontAwesomeIconWrapper'
-import { cn } from 'lib/utils'
+import { Icons } from 'utils/data'
+import { cn } from 'utils/utils'
```
arkid15r: `utils/utils` looks odd

frontend/src/helpers/constants.ts
arkid15r, line 1: These looks like types

frontend/src/pages/ChapterDetails.tsx
arkid15r: The current structure is okay, let's revisit this if the number of entity related pages grows.

abhayymishraa: sure!!!!

abhayymishraa:
Thanks You , @arkid15r 
Any further changes ??

kasya (approved):
LGTM!

abhayymishraa:
Thanks @kasya !!!!

PR #448, opened 2025-01-13
Added switch based mapping for filters in searchbar

### LINKED ISSUE #305: Extend search bar to accept filter queries
It'd be great to implement a feature to support GitHub like filtering in search component, e.g. `project:project-key`, `stars>500`, `is:code`, etc.

Implement this functionality in the search component to translate it to Algolia index queries.

ShashaankS:
Resolves #305 

- [x] Added filters functionality to the projects index, with some filters like contributors>10, stars<100, is:tool, is:documentation.

There is a provision for other indexes also but github like filters may be not meaningful.

Screenshot from 2025-01-13 17-35-31

Screenshot from 2025-01-13 17-35-22

there are some configs to bet set on algolia app:
Screenshot from 2025-01-13 14-38-42

### DIFF
--- a/frontend/src/api/fetchAlgoliaData.ts
+++ b/frontend/src/api/fetchAlgoliaData.ts
@@ -1,6 +1,7 @@
 import { SearchResponse } from 'algoliasearch'
 import { AlgoliaRequestType, AlgoliaResponseType } from 'types/algolia'
 import { ENVIRONMENT } from 'utils/credentials'
+import { getParamsForFilters } from 'utils/filterMapping'
 import { client } from 'utils/helpers/algoliaClient'
 
 import { getParamsForIndexName } from 'utils/paramsMapping'
@@ -17,17 +18,37 @@ export const fetchAlgoliaData = async <T>(
   }
   try {
     const params = getParamsForIndexName(indexName)
+
+    let filters = ''
+    let searchQuery = query
+
+    const filterRegex = /(\w+)([:><=!]+)([^ ]+)/g
+    let match
+
+    while ((match = filterRegex.exec(query)) !== null) {
+      const [fullMatch, attribute, operator, value] = match
+      filters += `${attribute}${operator}${value} AND `
+      searchQuery = searchQuery.replace(fullMatch, '').trim()
+    }
+
+    filters = filters.trim().replace(/AND$/, '')
+    filters = getParamsForFilters(indexName, filters)
+
+    if (filterKey) {
+      filters = filters ? `(${filters}) AND idx_key:${filterKey}` : `idx_key:${filterKey}`
+    }
+
     const request: AlgoliaRequestType = {
       attributesToHighlight: [],
       hitsPerPage: 25,
       indexName: `${ENVIRONMENT}_${indexName}`,
       page: currentPage - 1,
-      query: query,
+      query: searchQuery,
       removeWordsIfNoResults: 'allOptional',
       ...params,
     }
-    if (filterKey) {
-      request.filters = `idx_key: ${filterKey}`
+    if (filters) {
+      request.filters = filters
     }
 
     const { results } = await client.search({
--- a/frontend/src/utils/filterMapping.ts
+++ b/frontend/src/utils/filterMapping.ts
@@ -1,0 +1,20 @@
+export const getParamsForFilters = (indexName: string, filter: string) => {
+  filter = filter.trim()
+  switch (indexName) {
+    case 'projects':
+      switch (true) {
+        case /^contributors([><]=?|=)(\d+)$/.test(filter):
+          return filter.replace(/^contributors([><]=?|=)(\d+)$/, 'idx_contributors_count$1$2')
+        case /^stars([><]=?|=)(\d+)$/.test(filter):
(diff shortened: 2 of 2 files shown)


frontend/src/utils/filterMapping.ts
```diff
@@ -0,0 +1,20 @@
...
+        case /^contributors([><]=?|=)(\d+)$/.test(filter):
+          return filter.replace(/^contributors([><]=?|=)(\d+)$/, 'idx_contributors_count$1$2')
+        case /^stars([><]=?|=)(\d+)$/.test(filter):
+          return filter.replace(/^stars([><]=?|=)(\d+)$/, 'idx_stars_count$1$2')
+        case /^forks([><]=?|=)(\d+)$/.test(filter):
+          return filter.replace(/^forks([><]=?|=)(\d+)$/, 'idx_forks_count$1$2')
+        case /^is:([a-zA-Z]+)$/.test(filter):
+          return filter.replace(/^is:([a-zA-Z]+)$/, 'idx_type:$1')
+        default:
+          return filter
```
arkid15r, line 15: Can we extend it to datetime fields (created/updated) too?

frontend/src/utils/filterMapping.ts
```diff
@@ -0,0 +1,20 @@
+export const getParamsForFilters = (indexName: string, filter: string) => {
+  filter = filter.trim()
+  switch (indexName) {
+    case 'projects':
```
arkid15r, line 4: What about contribute page (issues), will it be addressed in a separate PR?

arkid15r (commented):
This works!

Please add some tests for your code.
Also you can change the index [settings](https://github.com/OWASP/Nest/blob/1da1aae896c68ba378a82a25aa50fa35583bae75/backend/apps/owasp/index/project.py#L45) programmatically.

frontend/src/utils/filterMapping.ts
ShashaankS: @arkid15r We can do that, how the user will input the dates?
Is it like 1sep24 or 01092024 or ?
As the index contains the that flied in UNIX timestamp

frontend/src/utils/filterMapping.ts
ShashaankS: I have gone through the issues index, its most of the attributes are seachable,
only attribute we can use was `comments<10` or `is:open` `is:close`
Even there is no attribute for open and close

Rajgupta36: I think we can also add languages and topics based filtering  in contribute page

Rajgupta36 (changes requested):
1.is: documentation is working but for this is:documentation no result same as in case of project. can we trim space and add one space automatically

2.also when I write dual query in search bar is: documentation, stars<10 or is: documentation stars<10   or is: documentation,star>10 the result is something like second param  override first one can we add multiple query support together

3.suppose this is the query stars<10@ it throws error on page but i think it should return no project found something. look at filters
image

 i think there are some regex error pls fix that

frontend/src/api/fetchAlgoliaData.ts
```diff
@@ -17,17 +18,37 @@ export const fetchAlgoliaData = async <T>(
   }
   try {
     const params = getParamsForIndexName(indexName)
+
+    let filters = ''
+    let searchQuery = query
+
+    const filterRegex = /(\w+)([:><=!]+)([^ ]+)/g
```
Rajgupta36, line 27: can we put this logic in another function

frontend/src/api/fetchAlgoliaData.ts
```diff
@@ -17,17 +18,37 @@ export const fetchAlgoliaData = async <T>(
   }
   try {
     const params = getParamsForIndexName(indexName)
+
+    let filters = ''
+    let searchQuery = query
```
Rajgupta36, line 25: Could you please define the types for this? Otherwise, we may unintentionally mutate the variable from a string to another type, which could lead to unexpected behavior.

harsh3dev (commented):
@arkid15r @ShashaankS I think Raj has already suggested some changes.
I think there should be more parameters across pages to use.
Also please include tests for this PR.

I have a question: Currently there is no way users can know what are all the available options for filters. So should that be tackled in the new issue?

Thanks.

frontend/src/utils/filterMapping.ts
harsh3dev: @ShashaankS we can always convert the dates into desirable format.

kasya:
> I have a question: Currently there is no way users can know what are all the available options for filters. So should that be tackled in the new issue?
> 


This was also my concern. We could do something like a sample of possible options bellow the search bar or a link with a modal to look those up. Anything like that should help. Without it it's super unclear how to use search this way if person never had experience with this type of search before.
Example from task management tool that I've used:
Screenshot 2025-01-18 at 2 22 06 PM

kasya (commented):
@arkid15r  @ShashaankS Should we proceed as is and add possible options later on? 
I still need to review it on the frontend, just wanted to make sure it's not still work in progress.

PR #453, opened 2025-01-13
Sliced local index limit to 1000

### LINKED ISSUE #440: LImit local index to 1000 records per entity type
As the amount of data and the type of entities grow it takes a while to build the index locally.

Let's try limiting the number of indexed records to some reasonable number (start from 1000). This might introduce some unexpected behavior though (let's keep that in mind).

Naveen-Pal:
Resolves #440 
( _what a number_)

Sliced the queryset to 1000 for all the entity and reduced the reindexing time from 4m43s to 1m38s (for my system).


arkid15r (changes requested):
This change is not scoped to the local environment only -- it affects all of them. You've limited it everywhere. 

Please add it for local setups only and use a constant for the `1000` value.

arkid15r (approved):
LGTM 👍

PR #454, opened 2025-01-13
Add OpenStreetMap

### LINKED ISSUE #393: Add a map to chapters page
For a better UX we could add a map and show pins for existing chapters

We need to decide on the preferred map service provider (Google or other) and render map for all OWASP chapters around the world.

abhayymishraa:
Resolves #393 

image

image

Add the PR description here.


frontend/src/components/ChapterMap.tsx
kasya, line 1: Something's not working for me  🤔 When i get deep into a specific chapter on the map - I have this empty tooltip showing up. Looking at their docs it seems that the pop-up is customizable.
Screenshot 2025-01-13 at 3 09 44 PM

Also, should the results on the page get updated when we click into "ares" on the map to show only chapters inside that area? Or is it just for the info?

kasya (changes requested):
This looks great! One concern that I have is about the dark theme. Is there a way to change map theme? If so, would be good to have a dark map for dark theme.


Also, LOVE the package you've chosen! 👏🏼

frontend/src/components/ChapterMap.tsx
kasya: I realized that filtering projects based on the map state won't be that simple to implement at this point. If we ever decide to implement it we can get back to that. So disregard my last question. The only thing that's bothering me right now is the color scheme for dark theme 👌🏼

frontend/src/api/fetchAlgoliaGeoLoc.ts
arkid15r, line 1: Would it be possible to reuse `fetchAlgoliaData.ts` for this instead of creating another file?

frontend/src/api/fetchAlgoliaGeoLoc.ts
```diff
@@ -0,0 +1,33 @@
...
+import { client } from 'utils/helpers/algoliaClient'
+import { AppError } from 'wrappers/ErrorWrapper'
+
+export const fetchAlgoliaGeoLoc = async <T>(): Promise<AlgoliaResponseType<T>> => {
+  const indexName = 'chapters'
+  try {
+    const request = {
+      attributesToHighlight: [],
+      indexName: `${ENVIRONMENT}_${indexName}`,
+      hitsPerPage: 1000,
```
arkid15r: what if we have 2000 chapters?

frontend/src/components/ChapterMap.tsx
```diff
@@ -0,0 +1,56 @@
+import L from 'leaflet'
+import { useEffect, useRef } from 'react'
+import 'leaflet/dist/leaflet.css'
+import 'leaflet.markercluster/dist/MarkerCluster.css'
+import 'leaflet.markercluster/dist/MarkerCluster.Default.css'
```
arkid15r, line 5: this looks odd, just want to make sure importing something from `/dist/` is a standard practice

arkid15r (commented):
Great start! Could you look into these comments when you get a chance:

frontend/src/api/fetchAlgoliaGeoLoc.ts
abhayymishraa: maximum limitt of the algolia and the openstreetmap pointer is 1000 so that's why i hardcoded that!!!

frontend/src/components/ChapterMap.tsx
abhayymishraa: actually its directly importing from the node_modules and for that we can't do that much!!! it's css present inside of node modules

frontend/src/components/ChapterMap.tsx
abhayymishraa: what i choose the approach is having a separate api call for the map rendering and the card is rendered by different api call . But yeah i can update the tooltip by modifying the api call for it .
and for that speciic ```ares ``` thing it can be completed in another issue .

kasya: @abhayymishraa yes, forget about filtering by clicking the area in the map. We'll either address it in a separate issue or won't do that at all.

What about the dark theme? is there such an option?

abhayymishraa: yeah, i totally forgot about that , yes we can implement it but it looks  a bit odd like dark map isn't that much attractive but yes we can change saturation and all properties of it  to make it look like a dark mode happy to discuss further i have sent you some examples in the slack!!

frontend/src/api/fetchAlgoliaGeoLoc.ts
abhayymishraa: about this we can do that, then two calls go for the same thing. also the fetchAlgoliaData.ts api is returning only 25 responses but for map we need 1000 , so if i modify the fetchAlgoliaData.ts then the pagination logic needs to be rewritten !!!

arkid15r: You can make whatever API calls you need. My point is that you should reuse the existing code instead of creating almost a copy of that. Use/extend `fetchAlgoliaData` with optional params like`hitsPerPage` and `attributesToRetrieve`.

frontend/src/api/fetchAlgoliaGeoLoc.ts
arkid15r: I am not sure how to deal w/ OSM limitations but for the 1000+ chapters situation you should have implemented fetching all entities logic (consume until no more pages left).

arkid15r (changes requested):

abhayymishraa:
yes i can do that!!

frontend/src/api/fetchAlgoliaGeoLoc.ts
abhayymishraa: done

frontend/src/api/fetchAlgoliaData.ts
```diff
@@ -10,7 +10,8 @@ export const fetchAlgoliaData = async <T>(
   indexName: string,
   query = '',
   currentPage = 0,
-  filterKey?: string
+  filterKey?: string,
+  hitsperpage = 25
```
arkid15r:
```suggestion
  hitsPerPage = 25
```

frontend/src/pages/Chapters.tsx
```diff
@@ -19,6 +25,27 @@ const ChaptersPage = () => {
...
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const data: AlgoliaResponseType<ChapterType> = await fetchAlgoliaData(
+          'chapters',
+          '',
+          1,
+          '',
+          1000
```
arkid15r: Could you add params names to make it more readable?

kasya (changes requested):
Love the dark theme! Thanks for adding that! 👍🏼 
Could we also update the tooltips for dark theme to not be white? Maybe  gray with white text?

frontend/src/pages/Chapters.tsx
```diff
@@ -19,6 +25,34 @@ const ChaptersPage = () => {
...
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const searchParams = {
+          indexName: 'chapters',
+          query: '',
+          currentPage: 1,
+          filterKey: '',
+          hitsPerPage: 1000,
+        }
```
arkid15r, line 38: Can we just inline it w/o an extra const?

frontend/src/components/ChapterMap.tsx
```diff
@@ -0,0 +1,57 @@
...
+  useEffect(() => {
+    if (!mapRef.current) {
+      mapRef.current = L.map('chapter-map', {
+        worldCopyJump: false,
+        maxBounds: [
+          [-90, -180],
+          [90, 180],
+        ],
+        maxBoundsViscosity: 1.0,
+      }).setView([20, 0], 2)
```
arkid15r: Could you add comments and links to the docs on what those magic numbers mean?

arkid15r (changes requested):
Thanks, the code looks better. A couple more minor suggestions and we're ready to merge from my perspective:

frontend/src/pages/Chapters.tsx
abhayymishraa: In order to do  that i need to change the function  implementation then i need to change whole api calls to use object

abhayymishraa: image
 
 we can do this or we can change the funcrionality of api to accept object as argurement !! i think this change is heptic cause then we need to change whole implementaion from all pages  wherewe used that api call

frontend/src/components/ChapterMap.tsx
abhayymishraa: yep

frontend/src/pages/Chapters.tsx
arkid15r: I prefer working with an easily readable code, without comments where possible. If it needs some refactoring -- this is fine, we'll do it later.

arkid15r (commented):
It seems it's the best we can do for code quality for now.
I'm okay to merge this is UI looks good to @aramattamara and @kasya

aramattamara (commented):
Everything looks good to me. I’m not sure if this can be addressed in this issue, but it might be something you could easily include in this PR. It would be great to add a link to the city name so that, when clicked, information about the OWASP chapter appears below the map.

Screenshot 2025-01-15 at 3 27 06 PM

frontend/src/pages/Chapters.tsx
abhayymishraa: so what's next

abhayymishraa:
@aramattamara  actually for that i need to rewrite the whole logic, currently i am making a separate call for the map data. so the curent implementation couldn't resolve this thing.

arkid15r:
> Everything looks good to me. I’m not sure if this can be addressed in this issue, but it might be something you could easily include in this PR. It would be great to add a link to the city name so that, when clicked, information about the OWASP chapter appears below the map.

I thinks it's a good idea and we can add a lot of interesting stuff related to the map interactivity (like narrow down the chapter list items to only what is shown on the map, update the list while user is exploring the map). 

Let's improve it gradually as I'm sure @abhayymishraa wants to merge this work asap.

aramattamara (dismissed):

abhayymishraa:
actually wait @aramattamara i found something form it . we can do one thing

abhayymishraa:
on clicking the tooltip we can push user to /chapters/vit-bhopal-university

abhayymishraa:
hey @aramattamara  @arkid15r 
image

for this i need to create this approach by inseting the java script

frontend/src/index.css
```diff
@@ -186,6 +186,15 @@ a {
   }
 }
 
+.dark .leaflet-popup-content-wrapper {
```
kasya, line 189: We also need to update the tip of the tooltip in dark-theme. Right now it's still white. 
You should be able to do that easily, I believe it's a `leaflet-popup-tip` class 👌🏼 
Screenshot 2025-01-15 at 6 13 27 PM

kasya (changes requested):
Love it! One more thing I noticed that is missing now:

frontend/src/index.css
abhayymishraa: yes yes !! it's easy just need to add a css class

abhayymishraa:
### After this commit it looks like this :
* actually my cursor isn't recording!!!

[Screencast from 16-01-25 02:39:45 PM IST.webm](https://github.com/user-attachments/assets/0819a6e3-536f-4cce-9992-a49fb6ec65c7)

arkid15r (approved):
LGTM 👍

abhayymishraa:
Thanks for the merge!!

PR #435, opened 2025-01-12
Update Sentry Config

harsh3dev:
Resolves #370

PR #465, opened 2025-01-14
Extend Sentry config

### LINKED ISSUE #370: Improve Sentry configuration
For both backend/frontend make sure we set `environment` and `release` values. While environment is pretty straightforward, the release can be tricky: for staging we can use git commit SHA (the short one), whereas production should use real release version. In both cases the release values can be populated via ci/cd .env files generation.

**Additional context**
  - https://docs.sentry.io/platforms/python/configuration/releases/
  - https://docs.sentry.io/platforms/javascript/configuration/releases/

harsh3dev:
Resolves #370 

Duplicate of #435


arkid15r (approved):
LGTM, let's try it

PR #512, opened 2025-01-16
Change Refactor index synonyms

Raghavg18:
Resolves #(478)

I have renamed synonym files from plural to singular in
[link](backend/apps/github/index/issue.py) and 
[link](backend/apps/owasp/index/project.py)
as per required


arkid15r (changes requested):
Please make sure your code works locally before creating a PR

You need to change the corresponding file names too.

Raghavg18:
@arkid15r issue file name is already singular, I have changed for projects in my local. is there any other file also which needs to be addressed?

arkid15r:
> @arkid15r issue file name is already singular, I have changed for projects in my local. is there any other file also which needs to be addressed?

yes, try running `make index-data` to see the errors

PR #515, opened 2025-01-16
added View issue button with test case

### LINKED ISSUE #511: Making issues navigation easy
While going through the issues in contribute section i found that there is detailed description of the issue but navigating to that issue in github was a bit troublesome as i had to go to the repository manually and then access that issue

I would like to add a button that facilitates a redirect to the issue, so i will add a button besides the close button in the modal and that will allow the user to access the issue easily

**Additional context**
Here's the quick demo of the feature

https://github.com/user-attachments/assets/0864a2aa-3f74-4695-84ea-74664953a503

Dishant1804:
Resolves #511 

Added the view issue button, also i have written test case for the modal as new prop 'url' was added


frontend/src/components/Modal.tsx
```diff
@@ -100,13 +105,19 @@ export const Modal: React.FC<ModalProps> = ({
...
         {/* Actions */}
-        <div className="mt-6 flex justify-end">
+        <div className="mt-6 flex justify-end gap-4">
           <button
             onClick={onClose}
             className="rounded-md bg-gray-600 px-4 py-2 text-sm font-medium text-white hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-gray-500 focus:ring-offset-2 dark:bg-gray-700 dark:hover:bg-gray-600 dark:focus:ring-gray-600"
           >
             Close
           </button>
+          <button
```
arkid15r: Could you implement it in a more abstract way so we could reuse it for other pages if needed.

The button could be a general action button that would render only `handleActionClick` is passed (untie it from issues)

Also you should pass the button text from the upper level (add GH icon before "View Issue" text)

Also position it before the close button

arkid15r (changes requested):
Looks good, thanks for the tests too!

Please improve this based on:

frontend/src/components/Modal.tsx
Dishant1804: @arkid15r  Thank you for guiding, i have made the necessary changes and decoupled the button and made it abstract as we already had an action button and now everything comes from top level component along with GH icon, i have also modified test case for the modal. Let me know if anything else needs to be changed :)

image

arkid15r (approved):
LGTM 👍

PR #517, opened 2025-01-16
Use add instead of max

### LINKED ISSUE #480: Update project index to use sum() instead of max()
We currently use [max()](https://github.com/OWASP/Nest/blob/556a36652bddf2a446350cb23ea50f11b9094e3a/backend/apps/owasp/management/commands/owasp_aggregate_projects.py#L82) of all repositories for stars/forks/contributors/etc values. 
  - Change it with sum() to show aggregated values instead of one max value. 
  - Update the tooltips to indicate it's a total value

Naveen-Pal:
Resolves #480


arkid15r (approved):
LGTM 👍

Issue #520, opened 2025-01-16
Chapter map overlaps hamburger menu


yashgoyal0110:
<img width="769" alt="Image" src="https://github.com/user-attachments/assets/9ed66dea-e85f-4079-b0ab-eb1bb6d39b48" />

PR #525, opened 2025-01-17
Fix/hamburger overlapped by map

yashgoyal0110:
fixes: #520


yashgoyal0110:
@arkid15r here's the updated PR with map zooming problem resolved

Uploading Screen Recording 2025-01-17 at 9.03.56 AM.mp4…

arkid15r (approved):
LGTM

PR #530, opened 2025-01-17
Added Skeleton for every pages

### LINKED ISSUE #420: Replace Current Loader with the Skeleton Component
### **Is your feature request related to a problem? Please describe.**  
Currently, the website uses a loader to display components while content is being fetched or rendered. While functional, the loader is good  but we can use skelton instead of that as it will look amazing when the state is loading!!!

---

### **Describe the solution you'd like**  
Implement the [ Skeleton component](https://ui.shadcn.com/docs/components/skeleton) as a replacement for the current loader. The Skeleton component provides a more visually appealing and professional loading state, elevating the overall look and feel of the website.

---

### **Describe alternatives you've considered**  
1. Retaining the current loader

2. Creating a custom skeleton loader: Although this would work, it would require additional development time and effort compared to using the ready-made ShadCN Skeleton component.

---
### Some Examples 
-
1.  24 8-SkeletonLoadingIntro
2. image

---

### **Additional context**  
- The ShadCN Skeleton component can significantly improve the perceived performance of the website by offering a sleek and modern loading experience.  
- Here’s the [documentation for ShadCN Skeleton](https://ui.shadcn.com/docs/components/skeleton).  
- Incorporating this feature will enhance the overall user experience, ensure aesthetic consistency, and align the website with current UI/UX best practices.

abhayymishraa:
Resolves #420 

Add the PR description here.

- [x]  Added dark/light mode support in skeleton components
- [x]  Fixed testcases for the Skeleton 
- [x]  Replaced Loaders with the  Skeleton 

 ## Preview 
[Skeleton.Preview.webm](https://github.com/user-attachments/assets/30669c10-7a46-496d-821f-8a4a0e325a41)


frontend/src/components/skeletons/ChapterSkeleton.tsx
arkid15r, line 1: Chapters.tsx

frontend/src/components/skeletons/CommunitySkeleton.tsx
arkid15r, line 1: Users.tsx

frontend/src/components/skeletons/ContributeSkeleton.tsx
arkid15r, line 1: Contribute.tsx

frontend/src/components/skeletons/ProjectSkeleton.tsx
arkid15r, line 1: Projects.tsx

frontend/src/components/skeletons/UserDetailSkeleton.tsx
arkid15r, line 1: UserDetails.tsx

frontend/src/components/ui/card.tsx
arkid15r, line 1: why lowercase?

frontend/src/components/skeletons/ProjectSkeleton.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Skeleton className="h-4 w-[94%]" />
+      </div>
+
+      <div className="flex items-center gap-2 py-1">
+        <Skeleton className="h-4 w-[300px]" />
+      </div>
+
+      <div className="flex w-full flex-col gap-4 pr-4 pt-1">
+        <div className="flex w-full flex-wrap items-center gap-2">
+          {Array.from({ length: 16 }).map((_, i) => (
```
arkid15r: what's 16 here? could you define consts instead of using magic numbers?

arkid15r (changes requested):
Works for me, could you resolve these:

frontend/src/components/ui/card.tsx
abhayymishraa: shadcn defaults!!!! whenever you install any component from shadcn , we'll get only lowercase named components by default !!

frontend/src/components/skeletons/ProjectSkeleton.tsx
abhayymishraa: actually its the no. of times compoenent need to be repeat. i think its good instead of using using const we could use this

frontend/src/components/ui/card.tsx
arkid15r: can you rename it according to our naming rules?

frontend/src/components/skeletons/ProjectSkeleton.tsx
arkid15r: The magic numbers are never good. Please create a descriptively named const for these cases.
And why exactly 16?

abhayymishraa: that's the no, of cards rendering

frontend/src/components/ui/card.tsx
abhayymishraa: wait let me see that

abhayymishraa: done

frontend/src/components/SearchPageLayout.tsx
```diff
@@ -50,9 +52,36 @@ const SearchPageLayout = ({
         />
       </div>
       {!isSearchBarReady || !isLoaded ? (
-        <div className="mt-20 flex h-64 w-full items-center justify-center">
-          <LoadingSpinner imageUrl={loadingImageUrl} />
-        </div>
+        <>
+          {indexName === 'projects' && (
+            <div className="flex min-h-screen w-full flex-col items-center justify-center">
+              {[...Array(4)].map((_, idx) => (
```
arkid15r: another magic number

frontend/src/components/SearchPageLayout.tsx
```diff
@@ -50,9 +52,36 @@ const SearchPageLayout = ({
...
+                <ProjectSkeleton key={idx} />
+              ))}
+            </div>
+          )}
+          {indexName === 'issues' && (
+            <div className="flex min-h-screen w-full flex-col items-center">
+              {[...Array(4)].map((_, idx) => (
```
arkid15r: please fix it everywhere in your code

frontend/src/components/SearchPageLayout.tsx
```diff
@@ -50,9 +52,36 @@ const SearchPageLayout = ({
...
+          {(indexName === 'chapters' || indexName === 'committees') && (
+            <div className="flex min-h-screen w-full flex-col items-center justify-center">
+              {[...Array(4)].map((_, idx) => (
+                <ChapterSkeleton key={idx} />
+              ))}
+            </div>
+          )}
+          {indexName === 'users' && (
+            <div className="grid grid-cols-1 gap-6 sm:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4">
+              {[...Array(12)].map((_, idx) => (
```
arkid15r: ☝️

arkid15r (changes requested):

frontend/src/components/SearchPageLayout.tsx
abhayymishraa: somehow i missed that!!

frontend/src/components/SearchPageLayout.tsx
abhayymishraa: done resolved

frontend/src/components/SearchPageLayout.tsx
abhayymishraa: done
 :+1:

Dishant1804 (approved):
Looks good, covers all the loaders

yashgoyal0110:
@abhayymishraa concept is good but what if there's a change in card layout or designing or some change in page design occurs in near future due to any requirement, then will you keep changing skeleton layout again and again?

@arkid15r what's your take?

arkid15r:
> @abhayymishraa concept is good but what if there's a change in card layout or designing or some change in page design occurs in near future due to any requirement, then will you keep changing skeleton layout again and again?
> 
> @arkid15r what's your take?

I don't know if there is a way to avoid it, e.g. autogenerate the skeletons? I think keeping the skeletons structure simple (not 1:1 copy of a real component) may allow us to avoid additional work?

yashgoyal0110:
> > @abhayymishraa concept is good but what if there's a change in card layout or designing or some change in page design occurs in near future due to any requirement, then will you keep changing skeleton layout again and again?
> > @arkid15r what's your take?
> 
> I don't know if there is a way to avoid it, e.g. autogenerate the skeletons? I think keeping the skeletons structure simple (not 1:1 copy of a real component) may allow us to avoid additional work?

yes i also think we should have generic template to reduce frequent updates with ui changes

Dishant1804:
> > @abhayymishraa concept is good but what if there's a change in card layout or designing or some change in page design occurs in near future due to any requirement, then will you keep changing skeleton layout again and again?
> > @arkid15r what's your take?
> 
> I don't know if there is a way to avoid it, e.g. autogenerate the skeletons? I think keeping the skeletons structure simple (not 1:1 copy of a real component) may allow us to avoid additional work?

There is a library that i've used in past it auto generates the skeleton for the components, it is called react-loading-skeleton, this will automatically adapt to the compoennt

yashgoyal0110:
> > > @abhayymishraa concept is good but what if there's a change in card layout or designing or some change in page design occurs in near future due to any requirement, then will you keep changing skeleton layout again and again?
> > > @arkid15r what's your take?
> > 
> > 
> > I don't know if there is a way to avoid it, e.g. autogenerate the skeletons? I think keeping the skeletons structure simple (not 1:1 copy of a real component) may allow us to avoid additional work?
> 
> There is a library that i've used in past it auto generates the skeleton for the components, it is called react-loading-skeleton, this will automatically adapt to the compoennt

@abhayymishraa can you please use same sort of library for future maintainability

abhayymishraa:
I think @arkid15r and @yashgoyal0110 creating a skeleton is easy!! Like you just need to copy the classes and just need to put skeleton in the place of input or any button , so I think it will be easy to make skeleton for the components, also there is no autogenerated skeleton i think.

abhayymishraa:
Making it generic wouldn't work cause we have a lot of different types of components and also there will be a lot of addition so what i think making a skeleton for every page will work as we only have to make the skeleton for the pages we have

Dishant1804:
@abhayymishraa for a longer run we cannot just do that for every component it will resulting in creation of lot of files just for skeleton
@arkid15r your view? i am suggesting that react-loading-skeleton is good library to look forward to

abhayymishraa:
@Dishant1804  i will look into it !!!!

arkid15r:
> @abhayymishraa for a longer run we cannot just do that for every component it will resulting in creation of lot of files just for skeleton @arkid15r your view? i am suggesting that react-loading-skeleton is good library to look forward to

It sounds like we could automate the skeletons generation? if that's the case it would definitely save time and be the preferred approach.

abhayymishraa:
@arkid15r  yess i will work on it !!!

harsh3dev (changes requested):
I think we can figure out is there is a way to automate the skeleton generation, but for now there could be a modular approach for handling this. @abhayymishraa 
1. For example as all the pages except for the community pages have similar layout which is based on the CardComponent, so a single skeleton for the card can be created.
2. You can pass null for the fields which are not to be shown like the user avatars on the issue page etc.
3. This way no need of creating separate components for every page.

abhayymishraa:
@arkid15r  now I'm totally confused which way should i choose!!?

kasya (commented):
@abhayymishraa hi! Is this still work in progress or is it ready for review? I was not sure after the discussion. Also, there are conflicts that need to be resolved before any review 👍🏼

abhayymishraa:
@arkid15r  ?? should i go with @harsh3dev  approach ?

abhayymishraa:
Heyy what from my point of view what i saw in the react-loading-skeleton that we still need to manually add the skeleton component and configure to look like it , so we can't say it that it generates skelton so i think like maximum we have now 4-5 pages and most of these are re-usable  so we can use them also configure if we have  a different ui, i think we should go with my approach for now as harsh suggested it cant be dont but we have to use conditional statements in all places!! i think we can go the approach what i thought of, we can go with harsh approach but in that case in future we have more component then we need to make it for all the added component so i think making it  for the pages can be a good approach that i took !!
@harsh3dev @arkid15r

arkid15r:
> Heyy what from my point of view what i saw in the react-loading-skeleton that we still need to manually add the skeleton component and configure to look like it , so we can't say it that it generates skelton so i think like maximum we have now 4-5 pages and most of these are re-usable so we can use them also configure if we have a different ui, i think we should go with my approach for now as harsh suggested it cant be dont but we have to use conditional statements in all places!! i think we can go the approach what i thought of, we can go with harsh approach but in that case in future we have more component then we need to make it for all the added component so i think making it for the pages can be a good approach that i took !! @harsh3dev @arkid15r

@Dishant1804 could you share more details about your experience of using `react-loading-skeleton`?

Dishant1804:
> > Heyy what from my point of view what i saw in the react-loading-skeleton that we still need to manually add the skeleton component and configure to look like it , so we can't say it that it generates skelton so i think like maximum we have now 4-5 pages and most of these are re-usable so we can use them also configure if we have a different ui, i think we should go with my approach for now as harsh suggested it cant be dont but we have to use conditional statements in all places!! i think we can go the approach what i thought of, we can go with harsh approach but in that case in future we have more component then we need to make it for all the added component so i think making it for the pages can be a good approach that i took !! @harsh3dev @arkid15r
> 
> @Dishant1804 could you share more details about your experience of using `react-loading-skeleton`?

Dishant1804:
> > Heyy what from my point of view what i saw in the react-loading-skeleton that we still need to manually add the skeleton component and configure to look like it , so we can't say it that it generates skelton so i think like maximum we have now 4-5 pages and most of these are re-usable so we can use them also configure if we have a different ui, i think we should go with my approach for now as harsh suggested it cant be dont but we have to use conditional statements in all places!! i think we can go the approach what i thought of, we can go with harsh approach but in that case in future we have more component then we need to make it for all the added component so i think making it for the pages can be a good approach that i took !! @harsh3dev @arkid15r
> 
> @Dishant1804 could you share more details about your experience of using `react-loading-skeleton`?

I used it and configured it a bit, the main moto why i used it was to get the building structure of the component ready without worrying much about the appearance, it does not completely auto generate as we expect here(but it gives us structure we need), we have to set height and width according to our component, very minimal configuration was required when i used it

abhayymishraa:
@Dishant1804 @arkid15r   but we have some pages that require more than just simple configuration like there is project details page we can't only just use a single skeleton and configure it ,  if used it definitely doesn't look good as it doesn't have that much details !! And if we still need to configure  it manually i think its breaking our main motive of using it, that is for autogeneration. Cause in that page we have different components so we need to  manually change settings !! I think we thought for autogeneration .. that doesn't provided by this library.... What i think is making a skeleton per page will fullfill tha use case here

arkid15r:
I checked the `react-loading-skeleton` package and it seems using pretty convenient approach for keeping skeletons tied to the code. My understanding it that using this library we would need update 3 files only: 
 - `components/Card`
 - `pages/Users`
 - `pages/UserDetails`
 
 Please correct me if I'm wrong @Dishant1804
 
 This approach is more structured and minimizes code repetition and maintenance efforts.

Dishant1804:
> I checked the `react-loading-skeleton` package and it seems using pretty convenient approach for keeping skeletons tied to the code. My understanding it that using this library we would need update 3 files only: 
>  - `components/Card`
>  - `pages/Users`
>  - `pages/UserDetails`
>  
>  Please correct me if I'm wrong @Dishant1804
>  
>  This approach is more structured and minimizes code repetition and maintenance efforts.

@arkid15r you are absolutely correct, the main advantage is the uniformity amongst the structure it provides, it reduces overhead of configuring everything

arkid15r:
> @Dishant1804 @arkid15r but we have some pages that require more than just simple configuration like there is project details page we can't only just use a single skeleton and configure it , if used it definitely doesn't look good as it doesn't have that much details !! And if we still need to configure it manually i think its breaking our main motive of using it, that is for autogeneration. Cause in that page we have different components so we need to manually change settings !! I think we thought for autogeneration .. that doesn't provided by this library.... What i think is making a skeleton per page will fullfill tha use case here

Let's focus on keeping it simple and maintainable w/ as low effort as possible. Having to make changes for the skeletons every time we change related components is a no-go. I suggest updating the Card code with generic skeleton structure that will cover all card-based entities (projects/chapters/committees) and also take care of user-related pages.

Let me know your thoughts about this approach @abhayymishraa

abhayymishraa:
@arkid15r  @Dishant1804 so should i wait for the component to be replaced by the chakra ui or  should i install chakra ui dependecy and make this skeleton !!!!? i this talk should wait until the components were replaced by chakra ui !!!

Dishant1804:
> @arkid15r @Dishant1804 so should i wait for the component to be replaced by the chakra ui or should i install chakra ui dependecy and make this skeleton !!!!? i this talk should wait until the components were replaced by chakra ui !!!

You can implement the skeleton loading using chakra ui as we were already finding a long term solution for skeleton structure right, so i think as chakra ui is approved we need to maintain the uniformity and stick to one thing as long as possible. So in my opinion you can move forward with chakraUI's skeleton if @arkid15r agrees to this approach

abhayymishraa:
@Dishant1804  So can i move forward to setup chakra ui for repo and work on skeleton???

Dishant1804:
> @Dishant1804  So can i move forward to setup chakra ui for repo and work on skeleton??? 

We need to take @arkid15r opinion to move forward with it let him approve as he is the maintainer of repo

arkid15r:
> > @Dishant1804  So can i move forward to setup chakra ui for repo and work on skeleton???
> 
> We need to take @arkid15r opinion to move forward with it let him approve as he is the maintainer of repo

We're good to go with Chakra UI as per the recent discussion and the poll. I'm not sure about the wider refactoring vs this particular issue implementation situation though. @Dishant1804 I'll let you decide on how to handle this:
  - wait until we have the initial setup (or more) implemented and finish this later
  - go ahead with completing this issue as it might not need a lot of code changes (I don't know)

Dishant1804:
> > > @Dishant1804  So can i move forward to setup chakra ui for repo and work on skeleton???
> > 
> > 
> > We need to take @arkid15r opinion to move forward with it let him approve as he is the maintainer of repo
> 
> We're good to go with Chakra UI as per the recent discussion and the poll. I'm not sure about the wider refactoring vs this particular issue implementation situation though. @Dishant1804 I'll let you decide on how to handle this:
> 
> * wait until we have the initial setup (or more) implemented and finish this later
> * go ahead with completing this issue as it might not need a lot of code changes (I don't know)

@abhayymishraa  We can go ahead with the implementation of chakraUI skeleton here, as it won't cause much hindrance in the code and wont require much refactoring after the implementation of this issue, try to make it more reusable and abstract, so that it can be used for future needs

abhayymishraa:
@Dishant1804 @arkid15r Sure, I'll take this up and try to implement it. By the way, I have a quick question. Shadcn provides a components.json file that specifies where everything should be installed, but from what I’ve researched, Chakra UI doesn’t seem to have a similar feature. I think manually placing the components will help maintain the current structure.

Dishant1804:
> @Dishant1804 @arkid15r Sure, I'll take this up and try to implement it. By the way, I have a quick question. Shadcn provides a components.json file that specifies where everything should be installed, but from what I’ve researched, Chakra UI doesn’t seem to have a similar feature.?? SO how can we manage our code structure

@abhayymishraa some components doesn't need to be installed (e.g. Card), however for some components like skeleton as reference for this issue, it will be default installed in components/ui folder, i dont think so we need to configure that. So just install the skeleton with given command and just use it

abhayymishraa:
i'm on it

Issue #531, opened 2025-01-17
hamburger menu can't be closed after resizing


yashgoyal0110:
Currently If I open hamburger menu and increase screen size then cross button to close the menu disappears and we can't close menu until we refresh the screen.

My solution: Introduced a useEffect to set isMobileView as false after a certain screen size.

Current State:

https://github.com/user-attachments/assets/566312b0-ddb5-48ab-91ef-cc02bf5cefdb

After making changes:

https://github.com/user-attachments/assets/204c07ef-9406-44e7-9dd0-37e62f93e011

yashgoyal0110:
@arkid15r Kindly review the PR

PR #532, opened 2025-01-17
hamburger-close-problem

yashgoyal0110:
fixes: #531


Dishant1804:
@yashgoyal0110 yes it makes sense to handle the state along with the tailwind visual changes as we never know if some users might prefer to resize the pane back and forth, so the combination of both state changes along with the tailwind visual changes will make do the job

yashgoyal0110:
@Dishant1804 i tnink only using useEffect will get the job done there's no need for **md:hidden**

frontend/src/components/Header.tsx
```diff
@@ -12,6 +12,17 @@ export default function Header() {
   const [mobileMenuOpen, setMobileMenuOpen] = useState(false)
   const location = useLocation()
   const toggleMobileMenu = () => setMobileMenuOpen(!mobileMenuOpen)
+  useEffect(() => {
+    const handleResize = () => {
+      if (window.innerWidth >= 768) {
```
arkid15r: Is there a constant for this value in any packages we already use so we could import it instead?

If there isn't let's create one ourselves.

arkid15r (commented):
It works

frontend/src/components/Header.tsx
yashgoyal0110: are you talking about constant value for width = 768px?

arkid15r: yes, `768`

yashgoyal0110: I have defined a constant in recent commit, kindly review

arkid15r: Please re-request review when PR is ready for another round

yashgoyal0110: got it ✅

yashgoyal0110: @arkid15r can u review it now?

arkid15r: Please click the re-request review button when it's ready as I suggested earlier.

frontend/src/utils/utility.ts
```diff
@@ -58,3 +58,4 @@ export const removeIdxPrefix = (obj: IndexedObject): IndexedObject => {
   }
   return newObj
 }
+export const desktopViewWidth = 768
```
arkid15r: Please move to `utils/constants`

arkid15r:
```suggestion
export const desktopViewMinWidth = 768
```

arkid15r (changes requested):

arkid15r (approved):
LGTM 👍 thank you!

Issue #523, opened 2025-01-17
Refactor frontend index attribute names


arkid15r:
All indexed entities fields start with `idx_` prefix. We have this [code](https://github.com/OWASP/Nest/blob/f60cd186a8b9c188852b5ab4524f7b744edbb31e/frontend/src/utils/utility.ts#L51) that we use for user pages only.

The idea is to apply it earlier, probably [here](https://github.com/OWASP/Nest/blob/f60cd186a8b9c188852b5ab4524f7b744edbb31e/frontend/src/api/fetchAlgoliaData.ts#L43) and get rid of the `idx_` prefix for all fields. This small change will affect a lot of files (somewhat around 30 I believ) but ultimately we will get more readable code without the redundant prefix.

yashgoyal0110:
@arkid15r can i resolve this?

Naveen-Pal:
I am interested.

PR #534, opened 2025-01-17
refactored-_idx-attribute

yashgoyal0110:
fixes: #523 

1. removed all _idx suffix from frontend part and some test files
2. used **removeIdxPrefix** inside fetchAlgoli.ts only for globally removing _idx suffix
3. created removeIdxPrefix function inside **api/utility.ts** as esLint wasn't allowing importing from other packages


arkid15r (approved):
LGTM 👍

PR #542, opened 2025-01-18
Improved FE test coverage

### LINKED ISSUE #537: Improve FE test coverage
We have a line or two in multiple files that are not covered with tests.

Take care of those cases hopefully making it 100%.

Image

Naveen-Pal:
Resolves #537


frontend/src/pages/ProjectDetails.tsx
```diff
@@ -147,6 +147,7 @@ const ProjectDetailsPage = () => {
             </div>
             {project.languages.length > 10 && (
               <button
+                data-testid="show-more-languages"
```
arkid15r: I believe this should be used as a last resort.
Could you try something like

`const button = screen.getByRole('button', { name: /submit/i });`

arkid15r (changes requested):
It seems we can remove try/catch for Chapters at least for consistency (48-49 still haven't been covered)

frontend/src/pages/ProjectDetails.tsx
arkid15r: Could you fix this so it'd be ready for the merge?

frontend/__tests__/src/pages/ProjectDetails.test.tsx
```diff
@@ -60,19 +65,191 @@ describe('ProjectDetailsPage Component', () => {
...
     render(<ProjectDetailsPage />)
-
     await waitFor(() => {
       const link = screen.getByText('https://example.com')
       expect(link.tagName).toBe('A')
       expect(link).toHaveAttribute('href', 'https://example.com')
     })
   })
+
+  // Below are additional tests to cover toggling “Show more/Show less” for languages, topics, and contributors.
```
arkid15r: Please avoid adding excessive comments. You should comment only difficult to understand code.

Naveen-Pal:
image
I am struggling with chapters.tsx. I will again work on it tomorrow.

arkid15r (approved):
LGTM 👍

Issue #536, opened 2025-01-17
Add map to chapter details page


arkid15r:
**Describe the solution you'd like**
After adding the map component we can reuse it on the map details page too.

**Describe alternatives you've considered**
Show the chapter's location on a map, adjust zoom for the best view.

yashgoyal0110:
@arkid15r would like to discuss it

arkid15r:
I believe you know what to do if you have any questions?
I suggest asking your questions here. Slack works too.

Rajgupta36:
I think arkid this is the part of my chapterdeatils page development issue

arkid15r:
> I think arkid this is the part of my chapterdeatils page development issue [#474](https://github.com/OWASP/Nest/issues/474)

Sorry, didn't see the map in that task's scope. I guess it's fine to parallel the work.

Rajgupta36:
Hi @yashgoyal0110, are you working on this task ?

yashgoyal0110:
yes

PR #544, opened 2025-01-19
Created ChapterDetails Component

### LINKED ISSUE #474: Implement chapter details page layout
Is your feature request related to a problem? Please describe.
We've recently implemented chapter details page and currently using the same Card component as we do for the chapter list page.

Describe the solution you'd like
Design a chapter details page with user friendly layout

Rajgupta36:
Resolves #474 

### Tasks:

- [x] Create `ChapterDetails` Component
- [x] Add Test Cases for `ChapterDetails`
- [x] Add Missing Test Cases for `ProjectDetails`
- [x]  Fix Minor Issues
----
Media
Screenshot 2025-01-19 025950
Screenshot 2025-01-19 030008


frontend/src/pages/ChapterDetails.tsx
```diff
@@ -1,17 +1,50 @@
...
+  }
+
+  const hostname = new URL(url).hostname.toLowerCase()
+
+  if (hostname === 'discord.com' || hostname.endsWith('.discord.com')) return faDiscord
+  if (hostname === 'instagram.com' || hostname.endsWith('.instagram.com')) return faInstagram
+  if (hostname === 'linkedin.com' || hostname.endsWith('.linkedin.com')) return faLinkedin
+  if (hostname === 'youtube.com' || hostname.endsWith('.youtube.com')) return faYoutube
+  if (hostname === 'twitter.com' || hostname === 'x.com' || hostname.endsWith('.x.com'))
+    return faTwitter
```
kasya: @Rajgupta36 can we use an icon for `X` instead? Like the one we on the Project card?

frontend/src/pages/ChapterDetails.tsx
kasya, line 1: When viewing a Chapter details page that has lot's of social media icons on the mobile view the layout get's messed up. Could we maybe wrap those? 
Screenshot 2025-01-20 at 9 16 57 AM

kasya (changes requested):
Thanks for working on this. Some comments ⬇️

frontend/src/pages/ChapterDetails.tsx
Rajgupta36: okk

frontend/src/pages/ChapterDetails.tsx
Rajgupta36: okk

Rajgupta36: image
fix both x and flex-wrap

arkid15r (commented):
It seems we need to be smarter here in a long-term perspective. How about creating a generic entity details page component (like the `Card` we have for projects/chapters/committees) that could be reused for any entity depending on the fields passed to the function. So here and for committee details page too we'd just use the same code.

@Rajgupta36 does it make sense?

Rajgupta36:
> It seems we need to be smarter here in a long-term perspective. How about creating a generic entity details page component (like the `Card` we have for projects/chapters/committees) that could be reused for any entity depending on the fields passed to the function. So here and for committee details page too we'd just use the same code.
> 
> We could use a generic component, but putting all logic into a single file might make things more complex. Would it be better to make it modular? Perhaps we can use two generic components and handle some  features like the map (ChapterPage) and issues/releases (ProjectPage) in their respective pages. What do you think, @arkid15r and @kasya ?

Rajgupta36:
Should I do this in this PR, @arkid15r?

Rajgupta36:
> > Should I do this in this PR, @arkid15r?
> 
> Up to you, we can merge this one but it'll require more refactoring then. Sorry I haven't suggested this approach earlier.

Okay, I completed this in a few minutes and created an issue (like the test cases earlier).

Rajgupta36:
Also, @arkid15r, what about these issues: #536, #549, #550, #551 ? I think the refactor depends on these issues.

Rajgupta36:
@kasya It’s ready for review.

arkid15r:
> Also, @arkid15r, what about these issues: #536, #549, #550, #551 ? I think the refactor depends on these issues.

The only one with map seems significant. And other don't look like they being taken care of.

arkid15r (approved):
I'm going to merge it as is -- we're going to unify it from the code reuse perspective.

Thanks for adding this @Rajgupta36!

kasya (approved):

PR #552, opened 2025-01-19
Added Project Page Sorting

### LINKED ISSUE #73: Implement Project page sorting
For the new frontend add ability to sort projects by starts/contributors/forks/other metadata attributes.

Based on the user's feedback:
> and also to sort by the number of contributors/stars

Rajgupta36:
Resolves #73 

### Task Progress

- [x] Added replicas into indexes
- [x] Added custom ranking to the indexes
- [x] Created `sortBy` component (reusable )
- [x] Customize `api` and `searchPageLayout` components
- [x] Writing test cases

----
### Media

https://github.com/user-attachments/assets/8a9b2151-1b9f-4dad-9082-821a4732456d


kasya (commented):
Haven't checked out and tested this yet locally, but from the video you posted I'm thinking it would be good to show which sorting is currently applied even when the sorting dropdown is collapsed. 
What do you think?

Rajgupta36:
> Haven't checked out and tested this yet locally, but from the video you posted I'm thinking it would be good to show which sorting is currently applied even when the sorting dropdown is collapsed. What do you think?

Yeah, we can do that. Currently, I show a tick mark next to the attribute where sorting is applied. If I display the component value, it might increase the size too much and look weird, since each component has a different value length.

backend/apps/owasp/index/project.py
```diff
@@ -91,4 +92,26 @@ def get_queryset(self):
...
+            f"{env}_projects_name_desc": ["desc(idx_name)"],
+            f"{env}_projects_stars_asc": ["asc(idx_stars_count)"],
+            f"{env}_projects_stars_desc": ["desc(idx_stars_count)"],
+            f"{env}_projects_contributors_asc": ["asc(idx_contributors_count)"],
+            f"{env}_projects_contributors_desc": ["desc(idx_contributors_count)"],
+            f"{env}_projects_forks_asc": ["asc(idx_forks_count)"],
+            f"{env}_projects_forks_desc": ["desc(idx_forks_count)"],
+        }
+
+        client.set_settings(f"{env}_projects", {"replicas": list(replicas.keys())})
```
arkid15r, line 113: It seems replicas can be created in `index/projects.py`

backend/apps/owasp/index/project.py
```diff
@@ -91,4 +92,26 @@ def get_queryset(self):
     @staticmethod
     def update_synonyms():
         """Update synonyms."""
+        ProjectIndex.configure_project_replicas()
```
arkid15r: This should be a separate flow just similar to `update_synonyms`.
Name it `def configure_replicas`

backend/apps/owasp/index/project.py
```diff
@@ -91,4 +92,26 @@ def get_queryset(self):
...
+
+    @staticmethod
+    def configure_project_replicas():
+        """Configure the settings for project replicas."""
+        env = settings.ENVIRONMENT.lower()
+        client = IndexBase.get_client()
+        replicas = {
+            f"{env}_projects_name_asc": ["asc(idx_name)"],
+            f"{env}_projects_name_desc": ["desc(idx_name)"],
+            f"{env}_projects_stars_asc": ["asc(idx_stars_count)"],
```
arkid15r: Use full name for a new index (stars_asc -> stars_count_asc)

arkid15r (changes requested):
I also have some suggestions on how the search bar + sorting should look like. Let's deal with that later.

frontend/src/components/ui/dropdown-menu.tsx
arkid15r, line 1: DropdownMenu.tsx

frontend/src/utils/sortingoptions.ts
arkid15r, line 1: SortingOptions.ts

arkid15r (changes requested):

frontend/src/components/ui/dropdown-menu.tsx
Rajgupta36: okk

frontend/src/utils/sortingoptions.ts
Rajgupta36: ok

arkid15r (approved):
Even though I see a lot of room for improvement for both code and UI (and we'll do it gradually) I'm merging this PR as it's good enough (it's actually great!) for PoC.

Awesome work on nailing Algolia index replicas for exhaustive sorting @Rajgupta36 👍

Rajgupta36:
> Even though I see a lot of room for improvement for both code and UI (and we'll do it gradually) I'm merging this PR as it's good enough (it's actually great!) for PoC.
> 
> Awesome work on nailing Algolia index replicas for exhaustive sorting @Rajgupta36 👍

Thanks sir🫡.

Issue #547, opened 2025-01-19
Move summary block higher on project details page


arkid15r:
**Describe the solution you'd like**
It should be right after the page header (before project details and statistics block)

<img width="1216" alt="Image" src="https://github.com/user-attachments/assets/b1394078-f4a2-495e-a490-6d15e55d375d" />

yashgoyal0110:
@arkid15r can i take it?

arkid15r:
The ones labeled as "good first issue" are for newcomers only (1-3 PRs tops)

Raghavg18:
@arkid15r Can u assign me this issue ?

PR #557, opened 2025-01-21
Move summary block higher on project details page

Raghavg18:
Resolves: #547 

Moved the summary block right after the page header
as per the requirement


arkid15r (approved):
LGTM 👍

PR #571, opened 2025-01-21
Add OWASP project schema

arkid15r:
An initial setup for OWASP schemas that includes a draft of the project schema.


.github/workflows/label-pull-requests.yaml
```diff
@@ -1,4 +1,4 @@
-name: 'Pull Request Labeler'
+name: Label pull requests
```
kasya: Should we make it upper case as before?
```suggestion
name: Label Pull Requests
```

.github/workflows/test-schema.yaml
```diff
@@ -0,0 +1,112 @@
+name: Test OWASP Schema
+
+on:
+  merge_group:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - schema/**
```
kasya, line 9: @arkid15r Why in CI\CD file we ignore `schema/*` and here we do `schema/**`?

kasya (changes requested):
Nice work! Left a couple of questions!

kasya (approved):
LGTM!

PR #592, opened 2025-01-23
Added Graphql Poc for projectDetails page

### LINKED ISSUE #524: Implement GraphQL PoC
Currently we have some additional data in the index but its size is quite limited. So the idea is to use our backed as a source for the data that can't and/or doesn't need to be in the index.

Research available libraries for both Django/React and create a PoC implementation of GraphQL server/client for project details page (see [pending PR](https://github.com/OWASP/Nest/pull/470))

REST API

**Additional context**
https://github.com/OWASP/Nest/discussions/514

abhayymishraa:
Resolves #524

Add the PR description here.
- [x] Implemented graphene for backend, exposing graphql
- [x] Apolloclient for the frontend 
- [x]  grapqhl for frontend 
- [x]  Initally setup of the graphql structure

[Screencast from 24-01-25 04:51:49 AM IST.webm](https://github.com/user-attachments/assets/99799930-5626-49be-b3a4-a8d102334ecc)


abhayymishraa:
i will resolve these workflow later

arkid15r (changes requested):
Thanks for adding this!

I didn't review the code as pre-commit failed but based on the file structure I can already recommend the following changes:
  - use common app for the base schema definition
  - use existing apps structure (github/owasp) to define related graphql queries/types (e.g. owasp/graphql/...`)
  - we won't be doing any mutations
  - the main graphql schema entry point could be something like `settings.graphql`

abhayymishraa:
Looks great and organized!!!!

abhayymishraa:
@arkid15r  We still have the ```types``` directory inside the ```graphql``` folder, even though we've migrated the type  to the ```node``` directory. Should we delete it?

frontend/src/pages/ProjectDetails.tsx
```diff
@@ -38,14 +46,21 @@ const ProjectDetailsPage = () => {
...
 
-  if (isLoading)
+  useEffect(() => {
+    if (data) {
+      setRecentReleases(data?.project?.recentReleases)
+      setRecentIssues(data?.project?.recentIssues)
+    }
+  }, [data])
+
+  if (isLoading || loading)
```
arkid15r: Could you update this using a better naming approach?

frontend/src/pages/ProjectDetails.tsx
```diff
@@ -38,14 +46,21 @@ const ProjectDetailsPage = () => {
...
     return (
       <div className="flex min-h-[60vh] items-center justify-center">
         <LoadingSpinner imageUrl="/img/owasp_icon_white_sm.png" />
       </div>
     )
 
-  if (!project)
+  if (!project || error)
```
arkid15r: The additional data retrieving error doesn't mean project doesn't exists.

frontend/src/pages/ProjectDetails.tsx
abhayymishraa: ok ok !!

frontend/src/pages/ProjectDetails.tsx
abhayymishraa: yes actually it mekes sense :100:

frontend/src/pages/ProjectDetails.tsx
```diff
@@ -69,6 +73,16 @@ const ProjectDetailsPage = () => {
       />
     )
 
+  if (error) {
```
arkid15r: Does it mean that we won't show any project data if graphQL query failed?

abhayymishraa: Oh, I see! I thought we might want to show the user a message in case the query fails, like letting them know the backend isn't responding well or there's an issue with GraphQL.

abhayymishraa: btw i am thinking of  it like this maybe a toast will help !!!
if the error from graphql a toast come at right-bottom saying graphql query failed!!
it's just an idea let me know what do you think about it !!!
for reference https://www.chakra-ui.com/docs/components/toast#types

arkid15r (approved):
Let's try it!

frontend/src/pages/ProjectDetails.tsx
arkid15r: Yes, toast fits better here

PR #608, opened 2025-01-25
Update Logic and Migrate SortBy Component from ShadCN to Chakra UI

### LINKED ISSUE #73: Implement Project page sorting

Rajgupta36:
Resolves #73 

## Tasks
- Migrate `sortBy` component from Shadcn to Chakra UI
- Separate buttons for order and attribute sorting
- Update Jest setup to support Chakra UI test cases

----

Media

https://github.com/user-attachments/assets/0bfb3fa7-a733-43ca-9a60-5ca2016a06ed


frontend/src/components/SortBy.tsx
```diff
@@ -1,56 +1,92 @@
...
 
   return (
-    <DropdownMenu open={open} onOpenChange={handleOpenChange}>
-      <DropdownMenuTrigger asChild>
-        <button
-          type="button"
-          className="h-10 w-[100px] justify-between border-2 bg-white text-[#292e36] dark:bg-[#3C3C3C] dark:text-[#D1D5DB]"
+    <div className="flex items-center gap-4">
+      {/* Sort Attribute Dropdown */}
+      <div className="rounded-xl bg-gray-200 px-2 shadow-sm dark:bg-[#323232]">
```
kasya: Can we increase x-padding here a little bit? 
```suggestion
      <div className="rounded-xl bg-gray-200 px-3 shadow-sm dark:bg-[#323232]">
```

frontend/src/components/SortBy.tsx
kasya, line 1: I noticed that sometimes the checkmark next to the sort option does not show up 🤔 Could you look into that?
Screenshot 2025-01-25 at 7 02 58 PM
Screenshot 2025-01-25 at 7 03 26 PM

kasya (changes requested):
Love this! A couple of suggestions and one issue that I found:

frontend/src/components/SortBy.tsx
Rajgupta36: Hi @Kasya,

As I found in the [documentation](https://www.chakra-ui.com/docs/components/select#controlled), when the component renders for the first time, it doesn’t apply the checked icon on selected item. However, when we switch, the icon is applied. Sometimes, due to caching page not rerender, so the checkmark is already applied, which is causing the unexpected behavior. 

We can try forcing the component to reload after every value change (to avoid cache) and add a manual icon next to the selected entity's name.

Rajgupta36: updated
image
image

kasya (approved):
Nice work!

PR #609, opened 2025-01-25
Fix/navbar body click resolved

kayashth1:
On clicking outside the navbar it closes as well as clicking on the nest logo it closes.


kasya (commented):
@kayashth1 Hi! Thanks for your contribution.

Could you elaborate on the issue you're trying to fix here? I'm not sure I understand the problem from the description of this PR. This would help me review and test the changes.

kayashth1:
Hello @kasya , Earlier when someone clicked on the body then the header didn't close but now it is happening and it does not happen when clicking inside the header। It improves UI experience.

Naveen-Pal:
Looks good

kayashth1:
@arkid15r @kasya Could you kindly suggest if there’s anything else I need to do?

kasya (approved):
Nice work! Works great!

PR #610, opened 2025-01-25
Filter inactive projects on API level

### LINKED ISSUE #606: Filter inactive projects on the API request level
We don't need inactive project/chapter in list/search
so instead of using 
`{projects && projects.filter((project) => project.is_active).map(renderProjectCard)}`
filter it on the API request level.

Reference: #422

Naveen-Pal:
Resolves #606


Naveen-Pal:
I am facing an error on running test

  ● Contribute Component › handles pagination for first page

  Unable to find an element with the text: Next Page. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
    
    
probably because of using `export function useSearchPage<T extends object>`
Help would be great.

arkid15r (changes requested):
It seems we can go even deeper and ask algolia not even return those items. For that we need `is_active` [faceting attribute](https://github.com/OWASP/Nest/blob/9fd5d355120cf7474777b5b2c4db1357746f9681/backend/apps/owasp/index/project.py#L46) in Chapter/Project index and use it in `fetchAlgoliaData` query.

arkid15r:
Resolved in https://github.com/OWASP/Nest/pull/1062

Issue #642, opened 2025-01-28
Improve NestBot /news command


arkid15r:
**Is your feature request related to a problem? Please describe.**
Currently we show [pretty basic](https://github.com/OWASP/Nest/blob/main/backend/apps/slack/commands/news.py#L20) /news command output

**Describe the solution you'd like**
Improve the output:
  - use 10 recent items from  [source directory](https://github.com/OWASP/owasp.github.io/tree/main/_posts) (requires some extra work)
  - format it nicely as an output for the `/news` command
  - keep the link to the OWASP website at the bottom

yashgoyal0110:
can i take this issue?

Rajgupta36:
@arkid15r could you assign this issue to me

PR #643, opened 2025-01-28
nestbot news command

yashgoyal0110:
fixes: #642

https://github.com/user-attachments/assets/1210be8f-560c-403a-b41e-fe7e04e80f9d


arkid15r (changes requested):
This works but the idea is to get the contents dynamically w/o the need to update the output manually

yashgoyal0110:
> This works but the idea is to get the contents dynamically w/o the need to update the output manually

You mean fetching news data through some api?

arkid15r:
> > This works but the idea is to get the contents dynamically w/o the need to update the output manually
> 
> You mean fetching news data through some api?

We have an example of getting file contents from GitHub [here](https://github.com/OWASP/Nest/blob/ebbc4c76f2de8916756d40d0e0e1304e1ef76589/backend/apps/slack/commands/staff.py#L8). In this case we need to get contents from multiple files (latest 10 for example).

yashgoyal0110:
> > > This works but the idea is to get the contents dynamically w/o the need to update the output manually
> > 
> > 
> > You mean fetching news data through some api?
> 
> We have an example of getting file contents from GitHub [here](https://github.com/OWASP/Nest/blob/ebbc4c76f2de8916756d40d0e0e1304e1ef76589/backend/apps/slack/commands/staff.py#L8). In this case we need to get contents from multiple files (latest 10 for example).

we are not targeting redirection link on every link.......ryt?

arkid15r:
> > > > This works but the idea is to get the contents dynamically w/o the need to update the output manually
> > > 
> > > 
> > > You mean fetching news data through some api?
> > 
> > 
> > We have an example of getting file contents from GitHub [here](https://github.com/OWASP/Nest/blob/ebbc4c76f2de8916756d40d0e0e1304e1ef76589/backend/apps/slack/commands/staff.py#L8). In this case we need to get contents from multiple files (latest 10 for example).
> 
> we are not targeting redirection link on every news......ryt?

Up to you, sounds like it works too.

yashgoyal0110:
> > > > > This works but the idea is to get the contents dynamically w/o the need to update the output manually
> > > > 
> > > > 
> > > > You mean fetching news data through some api?
> > > 
> > > 
> > > We have an example of getting file contents from GitHub [here](https://github.com/OWASP/Nest/blob/ebbc4c76f2de8916756d40d0e0e1304e1ef76589/backend/apps/slack/commands/staff.py#L8). In this case we need to get contents from multiple files (latest 10 for example).
> > 
> > 
> > we are not targeting redirection link on every news......ryt?
> 
> Up to you, sounds like it works too.


https://github.com/user-attachments/assets/c4c120ba-6901-46a6-9c32-214d34febb9f

backend/pyproject.toml
```diff
@@ -23,6 +23,7 @@ omit = [
 [tool.poetry.dependencies]
 algoliasearch = "^4.13.0"
 algoliasearch-django = "^4.0.0"
+beautifulsoup4 = "^4.12.2"
```
arkid15r: As we already have lxml installed can we handle it with xpath instead of adding a new dependency?

backend/apps/slack/commands/news.py
```diff
@@ -1,12 +1,48 @@
...
 
 from apps.common.constants import NL
 from apps.slack.apps import SlackConfig
 from apps.slack.blocks import markdown
 
 COMMAND = "/news"
+NEWS_URL = "https://owasp.org/news/"
+
+
+def fetch_latest_news():
```
arkid15r: This needs to be cached as we don't want to fetch it every time

arkid15r (approved):
Thanks for addressing this!

yashgoyal0110:
> Thanks for addressing this!

👍✅

PR #660, opened 2025-01-29
fixed bug of handlers and updated the testcases.

### LINKED ISSUE #659: Bug:  Testcase bug  slack/common/handler/*
**Describe the bug**
We have bug in the testcase
```backend/tests/slack/common/handlers/*```
They are not running even if you delete them you can still get coverage and testcases passing
Also it have wrong imports it still works!!!! It should fail!!!!

**To Reproduce**
1. Go to 'backend/tests/slack/common/handlers/*(All files)
2. As you can see all the testcases in that directory have wrong still the testcases worked 
3. Run ```make test```

**Expected behavior**
Basically these testcases that is written isn't executing!!!!

**Screenshots**

Imports: 

Image

Image

Image

Image

Even after the imports were wrong the testcases are passing that makes two choices either the testcase not running or there is bug in the implementation

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04(LTS)
 - Browser Chrome

**Additional context**

abhayymishraa:
Resolves #659 

Add the PR description here.
- [x] Fixed bug 
and improved coverage 

 i teseted and made sure testcases work
Screenshot from 2025-01-29 23-40-38


arkid15r (commented):
Good catch!

I'll add an extra point if you implement a solution that will make sure we avoid this type of problems in future (hint: pre-commit)

abhayymishraa:
> Good catch!
> 
> I'll add an extra point if you implement a solution that will make sure we avoid this type of problems in future (hint: pre-commit)

i was thinking about the same either we can make pre-commit or a workflow , we can use both !!!

abhayymishraa:
Hey @arkid15r 
In this latest commit i have implemeted the following chnages :
- Added a new pre-commit hook that makes sure that testcases file should be ending with ```*_test.py```
(there is a reason for this chnages as the official pre-commit hooks already have a hook to check it  but we are using  files in this format ```*_tests``` which is not the standard and in order to implement it without this we had to make a python file ourselve to check which it not optimal that's why i choose this approach)

- Replaced  ```*_tests.py```  to ```*_test.py```
- Removed constant.py inside test as it was failing in pre-commit and also it is of no use!!(i think so)

After this commit :
I have made sure it will pass for the right file formats:

  -  Screenshot from 2025-01-30 15-23-27
  
i have made sure it fails when it find testcase externsion incorrect :
  
- Screenshot from 2025-01-30 15-23-59

i have made sure all testcase are working with same coverage 
- Screenshot from 2025-01-30 15-09-26

abhayymishraa:
Your Welcome!!!!

PR #668, opened 2025-01-30
Feat/chapter details map

yashgoyal0110:
fixes: #536  

added map to chapter-details page with less initial zoom.

https://github.com/user-attachments/assets/d115a888-53b2-4c7d-91c7-d715dbea9128


frontend/src/components/ChapterMap.tsx
```diff
@@ -64,7 +64,7 @@ const ChapterMap = ({ geoLocData }) => {
     map.addLayer(markerClusterGroup)
 
     if (bounds.length > 0) {
-      map.fitBounds(bounds as L.LatLngBoundsExpression)
+      map.fitBounds(bounds as L.LatLngBoundsExpression, { maxZoom: 10 })
```
arkid15r, line 67: I'm not sure if it'd make more sense to support it as a ChapterMap param, what do you think?

arkid15r (commented):
Looks good

frontend/src/components/ChapterMap.tsx
yashgoyal0110: that can be done, but its not effecting any of the current functionality

arkid15r (approved):
LGTM 👍

PR #672, opened 2025-01-31
test cases for graphql

### LINKED ISSUE #626: Improve test coverage for  backend `graphql`
Image
Image
Image

( part of  #613)

nitinawari:
close #626 
image
image
image

### DIFF
--- a/backend/apps/common/management/commands/purge_data.py
+++ b/backend/apps/common/management/commands/purge_data.py
@@ -14,5 +14,5 @@ class Command(BaseCommand):
         with connection.cursor() as cursor:
             for nest_app in nest_apps:
                 for model in apps.get_app_config(nest_app).get_models():
-                    cursor.execute(f"TRUNCATE TABLE {model._meta.db_table} CASCADE")  # noqa: SLF001
-                    print(f"Purged GitHub {model._meta.verbose_name_plural}")  # noqa: SLF001
+                    cursor.execute(f"TRUNCATE TABLE {model._meta.db_table} CASCADE")
+                    print(f"Purged GitHub {model._meta.verbose_name_plural}")
--- a/backend/apps/common/models.py
+++ b/backend/apps/common/models.py
@@ -17,7 +17,7 @@ class BulkSaveModel(models.Model):
         model.objects.bulk_create((o for o in objects if not o.id), BATCH_SIZE)
         model.objects.bulk_update(
             (o for o in objects if o.id),
-            fields=fields or [field.name for field in model._meta.fields if not field.primary_key],  # noqa: SLF001
+            fields=fields or [field.name for field in model._meta.fields if not field.primary_key],
             batch_size=BATCH_SIZE,
         )
         objects.clear()
--- a/backend/apps/owasp/graphql/nodes/project.py
+++ b/backend/apps/owasp/graphql/nodes/project.py
@@ -16,14 +16,10 @@ class ProjectNode(BaseNode):
 
     recent_issues = graphene.List(IssueNode)
     recent_releases = graphene.List(ReleaseNode)
-    nest_url = graphene.String()
 
     class Meta:
         model = Project
-
-    def resolve_nest_url(self, info):
-        """Resolve project nest URL."""
-        return self.nest_url
+        fields = ()
 
     def resolve_recent_issues(self, info):
         """Resolve project recent issues."""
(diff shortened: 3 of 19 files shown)


nitinawari:
@arkid15r  Sorry for the late update. My exams are ongoing until February 15.

backend/apps/common/management/commands/purge_data.py
```diff
@@ -14,5 +14,5 @@ def handle(self, *_args, **options):
         with connection.cursor() as cursor:
             for nest_app in nest_apps:
                 for model in apps.get_app_config(nest_app).get_models():
-                    cursor.execute(f"TRUNCATE TABLE {model._meta.db_table} CASCADE")  # noqa: SLF001
-                    print(f"Purged GitHub {model._meta.verbose_name_plural}")  # noqa: SLF001
+                    cursor.execute(f"TRUNCATE TABLE {model._meta.db_table} CASCADE")
+                    print(f"Purged GitHub {model._meta.verbose_name_plural}")
```
nitinawari, line 18: This is not related to the PR. It appeared because I added `SLF001` in the Ruff file, as we need it for GraphQL

backend/apps/common/models.py
```diff
@@ -17,7 +17,7 @@ def bulk_save(model, objects, fields=None):
         model.objects.bulk_create((o for o in objects if not o.id), BATCH_SIZE)
         model.objects.bulk_update(
             (o for o in objects if o.id),
-            fields=fields or [field.name for field in model._meta.fields if not field.primary_key],  # noqa: SLF001
+            fields=fields or [field.name for field in model._meta.fields if not field.primary_key],
```
nitinawari, line 20: same for  this

backend/apps/owasp/graphql/nodes/project.py
```diff
@@ -16,9 +16,48 @@ class ProjectNode(BaseNode):
...
+            "releases_count",
+            "stars_count",
+            "subscribers_count",
+            "watchers_count",
+            "languages",
+            "licenses",
+            "topics",
+            "created_at",
+            "updated_at",
+            "released_at",
```
arkid15r: But we don't have a need to expose these fields yet?

arkid15r (commented):
Looks good except the project fields changes

backend/apps/owasp/graphql/nodes/project.py
nitinawari: When I ran the test cases, there was a warning because we didn't explicitly release the field. That's why I released it. Should we continue with the warning without making this change?

arkid15r: I'm not sure what the warning was, could you add it here?
If there must be a field, you could use `id` but we never use it yet.

nitinawari: =============================== warnings summary ===============================
DeprecationWarning: Creating a DjangoObjectType without either the fields or the exclude option is deprecated. Add an explicit `fields = '__all__' `option on DjangoObjectType ProjectNode to use all fields
    super_class.init_subclass_with_meta(**options)

arkid15r: For some reason I don't see it 🤷‍♂️ 
Could you add steps to reproduce it?

nitinawari: This only occurs when we write test cases for` owasp/graphql/nodes/projects.py.`
**Steps to reproduce:**
1. Remove` fields` entity from `meta` class  in the projects.py file.
2. Run the tests

nitinawari: I also tested with only the` id` field, and it's not giving a warning. I know we haven't used it anywhere yet, but we need to expose at least one field. Also, the `id` field is not very sensitive to  expose.

arkid15r: It sounds reasonable, I just can't figure out why I don't see the warning. How do you run the tests?

nitinawari: I’m not sure, maybe our versions are different.
I ran the tests using `make test.`
Okay, I'll commit without the field so you can check the warning in the GitHub checks.
image

nitinawari: I pushed the changes, so now you'll see the warning in the backend test.
https://github.com/OWASP/Nest/actions/runs/13067176604/job/36461423693?pr=672

arkid15r: I fixed that

arkid15r (approved):

PR #683, opened 2025-01-31
Update : sync jobs now run on main repo and main branch only

### LINKED ISSUE #661: Run sync data job in main repository only
Make sure it's not running in forks.

Dishant1804:
Resolves #661 

added a condition that checks whether the repo is the OWASP/Nest or not and also the branch is main or not, this ensures that the sync will only be done in the main repo and main branch and not on the forks, which resolves the issue :)

### DIFF
--- a/.github/workflows/sync-nest-data.yaml
+++ b/.github/workflows/sync-nest-data.yaml
@@ -11,6 +11,7 @@ env:
 jobs:
   update-data:
     name: Run data sync
+    if : github.repository == 'OWASP/Nest' && github.ref == 'refs/heads/main'
     env:
       ANSIBLE_HOST_KEY_CHECKING: False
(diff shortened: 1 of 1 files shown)


Dishant1804:
@arkid15r I've made the necessary changes. Let me know if any improvement/change is needed :)

.github/workflows/sync-nest-data.yaml
```diff
@@ -11,6 +11,7 @@ env:
 jobs:
   update-data:
     name: Run data sync
+    if : github.repository == 'OWASP/Nest' && github.ref == 'refs/heads/main'
```
arkid15r: This should be enough
```suggestion
    if: github.repository == 'OWASP/Nest'
```

arkid15r (approved):
LGTM 👍

arkid15r (approved):

PR #712, opened 2025-02-02
Fetch /gsoc command data from Algolia instead of DB

### LINKED ISSUE #686: Refactor /gsoc command to fetch data from algolia index
Currently we fetch it [from db](https://github.com/OWASP/Nest/blob/9fd5d355120cf7474777b5b2c4db1357746f9681/backend/apps/owasp/models/project.py#L239).

We need to refactor it to get data from algolia similar to [projects](https://github.com/OWASP/Nest/blob/9fd5d355120cf7474777b5b2c4db1357746f9681/backend/apps/owasp/api/search/project.py#L8) with narrowed down searchable attributes list to tags only.

Cache the results using `lru_cache`.

Dishant1804:
Resolves #686 

Made a separate file for gsoc search api that fetches the data from algolia, also the searchable attribute is set to `idx_custom_tags` so that we only get the tag list, which was expected in the issue.


Dishant1804:
@arkid15r I've made necessary changes. Please let me know if any changes are required :)

backend/apps/owasp/api/search/gsoc.py
arkid15r, line 1: You should use existing owasp/api/search/project logic instead.

arkid15r (changes requested):
We already have project search related entity, let's use that instead of adding another similar entity.

backend/apps/owasp/api/search/gsoc.py
```diff
@@ -0,0 +1,25 @@
...
+from algoliasearch_django import raw_search
+
+
+@lru_cache
+def get_gsoc_projects(year, attributes=None):
+    """Return GSoC projects from Algolia."""
+    from apps.owasp.models.project import Project
+
+    query = f"gsoc{year}"
+    searchable_attributes = ["idx_custom_tags"]
```
arkid15r: I think we should extend it to `idx_tags` too.

arkid15r: And `idx_topics`

backend/apps/owasp/api/search/gsoc.py
Dishant1804: @arkid15r So, the `get_gsoc_projects` function needs to be placed in owasp/api/search/project rather than creating separate file for it, correct? Let me know if this seems like the right direction. Correct me if i am not thinking in the direction as you are

backend/apps/owasp/api/search/gsoc.py
Dishant1804: sure i'll extend it to `idx_tags` and `idx_topics` as well

Dishant1804: I've added the `idx_topics` and `idx_tags` to the `searchable_attributes`

backend/apps/owasp/api/search/gsoc.py
Dishant1804: I've moved the `get_gsoc_projects` inside owasp/api/search/project, so that there are no separate files.

arkid15r (approved):
LGTM

Issue #697, opened 2025-02-01
Add downloads attribute to project schema


arkid15r:
**Describe the solution you'd like**
  - Extend [project](https://github.com/OWASP/Nest/blob/9fd5d355120cf7474777b5b2c4db1357746f9681/schema/project.json#L32) schema with an optional `downloads` attribute. 
  - if present it must contain at least 1 item (URL)
  - extend schema
  - add/extend tests for the new attribute

yashgoyal0110:
assign me

PR #713, opened 2025-02-02
download attribute in project schema

yashgoyal0110:
fixes: #697


arkid15r (changes requested):
Could you add more test cases based on the field requirements?

yashgoyal0110:
> Could you add more test cases based on the field requirements?

I tried but how can I apply for array items, as there is no enum for it

arkid15r:
> > Could you add more test cases based on the field requirements?
> 
> I tried but how can I apply for array items, as there is no enum for it

I'd like to have at least the following test cases:
  - empty array
  - non URI value
  - non unique values

yashgoyal0110:
> > > Could you add more test cases based on the field requirements?
> > 
> > 
> > I tried but how can I apply for array items, as there is no enum for it
> 
> I'd like to have at least the following test cases:
> 
> * empty array
> * non URI value
> * non unique values

okay

yashgoyal0110:
> > > > Could you add more test cases based on the field requirements?
> > > 
> > > 
> > > I tried but how can I apply for array items, as there is no enum for it
> > 
> > 
> > I'd like to have at least the following test cases:
> > 
> > * empty array
> > * non URI value
> > * non unique values
> 
> okay

all tests passed


Screenshot 2025-02-03 at 2 17 48 AM

schema/utils/validators.py
```diff
@@ -1,8 +1,32 @@
...
+    # valid domain format check
+    domain_regex = r"^(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9-]{2,}$"
+    if not re.match(domain_regex, parsed_url.netloc):
+        return False
+
+    return True
+
+
 def validate_data(schema, data):
+    # Perform URL validation before schema validation
```
arkid15r: Do we really need this?

arkid15r (changes requested):
Please use alphabetical order for new attributes everywhere.

schema/utils/validators.py
yashgoyal0110: I think yes, If we detect invalid URLs in the downloads list before calling validate(), we can skip the full schema validation and return an error immediately.
This prevents unnecessary processing and speeds up validation.

yashgoyal0110: wdyt?

arkid15r: In this particular case I'd prefer code readability vs validation performance. Out tests/validation speed is not a goal here, at least now.

yashgoyal0110: but currently there's no existing validation for URIs, then where we gonna add validation for it?

arkid15r: We should extend the validator to cover URIs too then. Refer to the [docs](https://python-jsonschema.readthedocs.io/en/stable/) on how to do that.

arkid15r (approved):
Merging this w/o URI validation part, let's address it later.

PR #715, opened 2025-02-02
Buttons tooltips anchor tags migration to chakraUI

### LINKED ISSUE #645: Migrating Buttons and Tooltips to chakraUI
This issue focuses on migrating all the buttons, links and tooltips to chakraUI, also i will modify the test cases (if at all needed) to get the expected result.

The step by step procedure is as follows:
- Migrate all the buttons across the frontend
- Migrate all the anchor tags to Links 
- Migrate all the tooltips to ChakraUI Tooltip

NOTE : All these are mostly independent and other contributors won't change these for the time being,so these won't cause merge conflicts with ongoing issues

Dishant1804:
Resolves #645 

Migrated all the tooltips, buttons and anchor tags to equivalent chakraUI components


Dishant1804:
@arkid15r I've migrated the necessary components mentioned in the issue. Please review it and let me know if any changes are required :)

kasya (dismissed):
Overall great migration! Thank you for updating these!

One issue I noticed:
in the light mode when some buttons in focus state now have this dark border (that is also kind of a pixel off the actual border 🤔 ). Can we get rid of this?

I found two instances of this, but there might be more. Ones that I found are on the User details page and on the issue details modal.
Screenshot 2025-02-02 at 6 49 08 PM
Screenshot 2025-02-02 at 6 49 21 PM

Dishant1804:
> Overall great migration! Thank you for updating these!
> 
> One issue I noticed:
> in the light mode when some buttons in focus state now have this dark border (that is also kind of a pixel off the actual border 🤔 ). Can we get rid of this?
> 
> I found two instances of this, but there might be more. Ones that I found are on the User details page and on the issue details modal.
> Screenshot 2025-02-02 at 6 49 08 PM
> Screenshot 2025-02-02 at 6 49 21 PM
> 

Yess @kasya thank you for suggesting the changes, I'll go to the root of this issue, but for now it seems like the default chakraUI button recipe is causing this focus ring effect, I'll fix that :)

harsh3dev (approved):
LGTM
Please take care of conflicts and comments from other reviewers.
Thanks

abhayymishraa (approved):

frontend/src/components/SortBy.tsx
```diff
@@ -73,28 +75,32 @@ const SortBy = ({
...
-              />
-            )}
-          </button>
           <Tooltip
-            id="sort-order-tooltip"
-            className="rounded-lg bg-white px-1 py-0 text-sm text-gray-600 shadow-md"
-          />
+            content={selectedOrder === 'asc' ? 'Ascending Order' : 'Descending Order'}
+            showArrow
+            positioning={{ placement: 'bottom-start' }}
```
kayashth1, line 81: image
The tooltip should be on the top middle as before. Kindly resolve it

kayashth1: image
The tooltip should be on the top middle as before. Kindly resolve it.

kayashth1 (changes requested):

frontend/src/components/SortBy.tsx
Dishant1804: @kayashth1 just so you know that there is no top-middle in positioning in tooltip, however i'll change the position to top-start for better view.

kayashth1 (approved):

Dishant1804:
> Overall great migration! Thank you for updating these!
> 
> One issue I noticed: in the light mode when some buttons in focus state now have this dark border (that is also kind of a pixel off the actual border 🤔 ). Can we get rid of this?
> 
> I found two instances of this, but there might be more. Ones that I found are on the User details page and on the issue details modal. Screenshot 2025-02-02 at 6 49 08 PM Screenshot 2025-02-02 at 6 49 21 PM

@kasya I've made the suggested changes now there is no focus rings on the mentioned buttons, please review it :)

frontend/src/components/ModeToggle.tsx
abhayymishraa, line 1: Hey, it's not a major change, but in our current behavior, when you hover over the toggle button, a tooltip appears and remains visible even after clicking the button, with the text dynamically updating to "Enable light mode" or "Enable dark mode." However, in this PR, the tooltip disappears when the button is clicked, even if the cursor is still on the toggle button. It’s not a big issue, but if it can be fixed with just a property, it could be helpful.

Apart from that, everything looks good! Great work!

Dishant1804: @abhayymishraa Thank you for addressing this, I dont think that it is that much important as the main purpose of tooltips is just to provide a little context or very short information of what that particular component does in the website, more over this behaviour that you've addressed is very specific to libraries that we use, the react-tooltip was what we were using earlier it had this persistent display support, but chakraUI doesn't have that behaviour. Moreover the user will automatically understand that the button is for toggle once they have clicked it and wont need persistent tooltip in that case

abhayymishraa: @Dishant1804 I get your point, and it’s not a major issue. The main reason I brought it up is that the previous behavior felt a bit smoother since the tooltip didn’t abruptly disappear on click. Users might not need the tooltip after the first interaction, but keeping it visible until the cursor moves away just makes the experience feel more consistent.

That said, if Chakra UI doesn’t support it natively, it’s not a blocker. Just wanted to check if it was an easy fix. Thanks for the explanation!

Dishant1804: @abhayymishraa if you have any workaround suggestion then please suggest, because i think we need to modify the entire component internally as it is given by chakraUI, there is no easy fix to this, and for such an small issue i dont think we should go so much deep that we alter the implementation of the component, as this will just ruin the purpose of using a library.

abhayymishraa: I get that modifying the entire component might not be worth it for such a small issue, and I agree that we shouldn’t override the purpose of using a library. I was just checking if there was a simple workaround.

frontend/__tests__/src/pages/UserDetails.test.tsx
```diff
@@ -37,9 +39,11 @@ const mockUser = {
 const renderWithRouter = (ui: React.ReactElement) => {
   return render(
     <MemoryRouter initialEntries={['/user/testuser']}>
-      <Routes>
-        <Route path="/user/:login" element={ui} />
-      </Routes>
+      <ChakraProvider value={system}>
```
Rajgupta36: Instead of rewriting wrapping code, use the render function from the wrappers/testUtil file.

frontend/src/components/Pagination.tsx
```diff
@@ -88,6 +92,8 @@ const Pagination: React.FC<PaginationProps> = ({
           </React.Fragment>
         ))}
         <Button
+          focusVisibleRing={'none'}
```
Rajgupta36: I didn't find any difference in ui in my system when I removed this prop. Also, if we pass these props to most of the buttons, it might be a good idea to create a component that contains these as default props.
before removing
image
After removing
image
correct me if i am wrong

Rajgupta36 (changes requested):
image
I didn't find any difference when I removed this prop. Also, if we pass these props to most of the buttons, it might be a good idea to create a component that contains these as default props.

frontend/src/components/Pagination.tsx
Dishant1804: yeah i also thought about it, but since there is no noticeable difference here, we can just remove the props wherever necessary and we already have action button, so keeping the props in action button will make much more sense, moreover the focus ring issue was only caused in action buttons, so fixing action button's focus ring will solve the issue, and for further uses we are anyways going to use action button.

frontend/__tests__/src/pages/UserDetails.test.tsx
Dishant1804: Thank you for suggesting that i'll change it, i was aware of testutils wrapper but just forgot to use it here :)

arkid15r:
Just FYI, I've just merged this https://github.com/OWASP/Nest/pull/726

Dishant1804:
> Just FYI, I've just merged this #726

@arkid15r Thank you for informing, our code is compatible with this version, the relevant changes are related to performance which chakraUI has decided to implement on their side which is good as it'll now increase the performance of the tooltip.
image
these are currently relevant to us

frontend/src/components/Pagination.tsx
Dishant1804: @Rajgupta36 i've made the necessary changes please review it :), now the props are in action button which does not show the focus ring, we got rid of it

Dishant1804:
@kasya @Rajgupta36 @arkid15r I've made all the changes suggested by the reviewers, please let me know if we need more changes (if any) :)

Dishant1804:
> image This error appears in both normal and incognito modes. Can anyone verify it, please?

@Rajgupta36 i went deep into this issue and researched about it, here's what i found out:
- The deprecation error we are encountering is due to the browser, the browsers now expect the libraries to use `:state(bg-currentcolor)` instead of `--bg-currentcolor` syntax.
- The deprecation will be solved once the chakraUI library migrates their internal implementation
- Devs at chakraUI have already started working on it we can check it [here](https://github.com/w3c/csswg-drafts/issues/4805)
- Also people using various libraries are facing the same issue we can check out reference [here](https://github.com/csstools/postcss-plugins/issues/1340) , [here](https://github.com/gradio-app/gradio/issues/9345) and [here](https://github.com/Shopify/polaris/issues/12987)

The deprecation will be automatically resolved in upcoming version changes of chakraUI, correct me if i am wrong. I've just presented what i researched about

arkid15r:
@Dishant1804 my understanding we can merge this after resolving the conflicts?

Dishant1804:
@arkid15r i've resolved the conflicts please merge it :)

arkid15r (approved):
LGTM 👍

PR #719, opened 2025-02-02
Added demo_url attribute to project schema

### LINKED ISSUE #692: Add demo attribute to project schema
  - Extend [project](https://github.com/OWASP/Nest/blob/9fd5d355120cf7474777b5b2c4db1357746f9681/schema/project.json#L32) schema with an optional demo attribute. 
  - pick best of different options for the name (demo, demo_website, demo_url, etc)
  - extend schema
  - add/extend tests for the new attribute

kayashth1:
Resolves #692


schema/utils/validators.py
arkid15r, line 1: You shouldn't have changed this file

schema/project.json
```diff
@@ -79,6 +79,11 @@
         "Tools"
       ],
       "type": "string"
+    },
+    "demo_url": {
```
arkid15r: Please move it to its right alphabetical spot

arkid15r (changes requested):

schema/utils/validators.py
kayashth1: Without this URL validation, pytest fails to validate. please guide me on how to tackle it.

arkid15r (approved):

PR #733, opened 2025-02-03
Optimize images for better performance

Rajgupta36:
## Tasks

- [x] Added memoization for better caching 
- [x] Added size attribute 
- [x] Remove unnecessary quality rendering

## Performance Comparison
Before: - Content Length: Between 10,000 to 20,000
After : - Content Length: Between 5,000 to 7,500

## Media Comparison
 size difference 
### Before
Before Image

### After
After Image


aramattamara (approved):
Thank you for optimizing this! LGTM!

arkid15r (approved):

PR #753, opened 2025-02-05
Implemented RepositoryDetails page

### LINKED ISSUE #650: Implement repository details page
Implement repository details page (we can start from project entity). 
It'll mostly be looking like project details page just narrowed down to a single repository data. 
The page should be available via `/projects/<project-key>/repositories/<repository-key>` , e.g. https://nest.owasp.dev/projects/nest/repositories/nest path

As we already know repository key we don't need to hit algolia and can use graphql for this.

Rajgupta36:
Resolves #650 

### Tasks:
- [x] Update node and query for repository and project model
- [x] Create Repository Details page
- [x] Add improvements to Generic Page
- [x] Create `get_repository_query` method in frontend
- [x] Add frontend test cases
- [x] Update backend test cases

----
Media

note :- fix error display to repository not found 

https://github.com/user-attachments/assets/69bb3eca-8886-4285-b1b6-18bb508113e8


frontend/src/pages/RepoDetails.tsx
```diff
@@ -0,0 +1,104 @@
...
+      />
+    )
+
+  const repositoryDetails = [
+    {
+      label: 'Last Updated',
+      value: formatDate(repository.updatedAt),
+    },
+    {
+      label: 'License',
```
kasya: The Frontend looks good to me! 👍🏼 
My only concern was with the `License` field when we don't have anything. Right now it's just an empty line, but what do you think about saying something like `License: unknown`?

kasya (changes requested):
Nice work!!

frontend/src/pages/RepoDetails.tsx
Rajgupta36: Yeah, I think we can apply this to all labels because if value is not present, we should show "Unknown" instead of blank  just like we do with stats

arkid15r (approved):
LGTM 👍 great work!

PR #754, opened 2025-02-05
Added a method to convert block to text and fixed warning 

### LINKED ISSUE #750: Add unified block-to-text converter utility for Slack messages
**Description**
Currently, the Slack command handlers are missing the required `text` field in `chat_postMessage` calls, which is causing warnings in the terminal. Additionally, When using `client.chat_postMessage()`, we need to provide both `blocks` and `text` parameters. Currently, we're missing a unified way to convert block content (that includes emojis, newlines, etc.) to text format, requiring manual text crafting for each specific situation.

Error: 
Image

**Current Behavior**
```python
client.chat_postMessage(channel=conversation["channel"]["id"], blocks=blocks)
```
The above code causes warnings because it's missing the required `text` field, and there's no standard way to extract text from blocks.

**Proposed Solution**
Create a utility function that:
- Accepts the same blocks that we pass to `client.chat_postMessage()`
- Converts block content to text while preserving:
  - Emojis
  - Newlines
  - Other formatting elements
- Can be reused across all command handlers

abhayymishraa:
Resolves #750 

Add the PR description here.

- [x] Added a utility function to convert block into text
- [x] Included emojis, /n 
- [x] Created a extra method to convert <links>   to better viewing/
- [x] Added text field for all slack commands and tested for all commands 

**Screenshot**

#### Response ```text``` from ```client.chat_postMessage```
Screenshot from 2025-02-05 16-51-29

#### Reaponse.data looks like this 
Screenshot from 2025-02-05 16-52-05


backend/apps/slack/utils.py
```diff
@@ -80,3 +81,42 @@ def get_staff_data(timeout=30):
         )
     except (RequestException, yaml.scanner.ScannerError):
         logger.exception("Unable to parse OWASP staff data file", extra={"file_path": file_path})
+
+
+def blocks_to_text(blocks):
```
arkid15r: Could you add tests for the new logic?

abhayymishraa: yes, i'm already working on the testcases , i made pr because there was some bug in my local testcases giving me error

abhayymishraa: i'm already working on the testcases

arkid15r: Converted it to draft.

abhayymishraa: @arkid15r sorry already completed the testcases

abhayymishraa:
- [x] Added testcas for utils
image

arkid15r (approved):
LGTM 👍 

What to improve:
  - better naming
  - never `+` strings in Python

abhayymishraa:
Thank You!!

PR #755, opened 2025-02-05
Add local index names exclusion mechanism

### LINKED ISSUE #677: Avoid algolia index number limitation in dev
We're facing algolia index limit (20 instances) problem. For development environment we want to control what exactly is indexed

Implement enable/disable index functionality for dev environment.
Add a custom `register` decorator to `apps.common.index` that behaves like original `algoliasearch_django.decorators.register` decorator for enabled index models and does nothing for models we don't want to index locally. It should skip indexing for local (development) environment only. Make sure a model is identified by app/model name pair to avoid confusion if same name models exist in different apps.
Create a list of excluded models (all other models are indexed by default). Make sure there is a way to override the list using .env variable (a list of comma separated values with app/models name items to override).

nitinawari:
resolve #677 
image


nitinawari:
@arkid15r ready for review

arkid15r (approved):
Let's try this!

PR #758, opened 2025-02-06
Migrated entity detail page to graphql

### LINKED ISSUE #751: Migrate entity detail pages from Algolia to direct GraphQL queries
**Problem Statement**
Currently, we're using Algolia for both entity list pages (search/ranking/sorting) and entity detail pages. However, for detail pages, we can optimize by fetching data directly from the database via GraphQL instead of hitting Algolia.

**Scope**
Pages to migrate:
- Project details page
- Chapter details page
- Committee details page
- User details page

List pages will continue using Algolia for search functionality.

**Current Implementation**
- All entity data is being fetched from Algolia
- Detail pages don't require Algolia's search/ranking features
- Unnecessary Algolia API calls for direct lookups

**Proposed Changes**
1. Create GraphQL queries for each entity type:

2. Modify detail page components to use GraphQL instead of Algolia:
   - Replace Algolia hooks with GraphQL queries
   - Ensure all necessary data is available via GraphQL
   - Update loading states and error handling

3. Modify the incoming data from the graphql to adapt pages

4. Entity types to update:
   - Projects
   - Chapters
   - Committees
   - Users

**Required Chnages**
- frontend
- backend 
- (tests)

**Benefits**
- Reduced Algolia API usage
- More direct and efficient data fetching
- Lower latency for detail pages
- Reduced costs (fewer Algolia operations)

abhayymishraa:
Resolves #751

Add the PR description here.
- [x] Migrated project detail page to use graphql 
- [x] Added types to the components related to project details 
- [x] logic to ```www-project-``` in the graphql 
- [x] added necessary types to the graphql nodes  user, repository,project,


abhayymishraa:
- [x] Migrated ChapterDetails Page
- [x] Grpahql queries and schema for chapter model

abhayymishraa:
- [x] Removed extra types that of no use 
- [x] Implemented types and change type
- [x] Included only what is important in nodes

abhayymishraa:
- [x] Completed CommitteeDetails Page 
- [x] Addede/fixed testcase

abhayymishraa:
### CommmitteeDetailsPage
Screenshot from 2025-02-07 23-55-36
Screenshot from 2025-02-07 23-55-29

Testcases results
Screenshot from 2025-02-07 22-41-31
Screenshot from 2025-02-07 22-44-21

abhayymishraa:
After Latest Commit :- 
image

abhayymishraa:
This is an experimental implementation of a common place for shared fields. If it works well, we can keep it; otherwise, we can always revert it.

backend/apps/github/graphql/nodes/repository.py
```diff
@@ -31,6 +33,7 @@ class Meta:
             "description",
             "forks_count",
             "key",
+            "latest_release",
```
arkid15r: Do we need to define `latest_release` here too (you created a custom field for that above, right)?

backend/apps/github/graphql/nodes/user.py
```diff
@@ -1,18 +1,47 @@
...
+        """Resolve user created at."""
+        return self.idx_created_at
+
+    def resolve_updated_at(self, info):
+        """Resolve user updated at."""
+        return self.idx_updated_at
+
+    def resolve_url(self, info):
+        """Resolve user URL."""
+        return self.url
```
arkid15r: I believe this could be just a part of `fields` tuple. Do we need to define custom fields that are already parts of a model?

backend/apps/github/graphql/queries/repository.py
```diff
@@ -19,8 +19,9 @@ class RepositoryQuery(BaseQuery):
     def resolve_repository(root, info, project_key, repository_key):
         """Resolve project."""
         try:
+            normalized_key = "www-project-" + project_key
```
arkid15r: Let's not use string concatenation this way 
```suggestion
            normalized_key = f"www-project-{project_key}"
```

backend/apps/github/graphql/queries/repository.py
```diff
@@ -19,8 +19,9 @@ class RepositoryQuery(BaseQuery):
...
             return (
-                Project.objects.get(key=project_key)
+                Project.objects.get(key=normalized_key)
```
arkid15r: Could you inline `normalized_key`?

backend/apps/owasp/graphql/nodes/committee.py
```diff
@@ -0,0 +1,48 @@
...
+        """Resolve project forks count."""
+        return self.owasp_repository.forks_count if self.owasp_repository else 0
+
+    def resolve_stars_count(self, info):
+        """Resolve project stars count."""
+        return self.owasp_repository.stars_count if self.owasp_repository else 0
+
+    def resolve_repositories_count(self, info):
+        """Resolve project repositories count."""
+        return 1  # only one repository per committee
```
arkid15r: If this changes in future this could be a problem. Can we use some sort of `.count()` method here?

backend/apps/owasp/graphql/queries/__init__.py
```diff
@@ -1,7 +1,9 @@
 """OWASP GraphQL queries."""
 
+from apps.owasp.graphql.queries.chapter import ChapterQuery
+from apps.owasp.graphql.queries.committee import CommitteeQuery
 from apps.owasp.graphql.queries.project import ProjectQuery
 
 
-class OwaspQuery(ProjectQuery):
+class OwaspQuery(ProjectQuery, ChapterQuery, CommitteeQuery):
```
arkid15r:
```suggestion
class OwaspQuery(ChapterQuery, CommitteeQuery, ProjectQuery):
```

frontend/src/pages/ChapterDetails.tsx
```diff
@@ -1,39 +1,48 @@
...
+  if (isLoading) {
     return (
       <div className="flex min-h-[60vh] items-center justify-center">
         <LoadingSpinner imageUrl="/img/owasp_icon_white_sm.png" />
       </div>
     )
+  }
 
-  if (!chapter || !chapter.is_active)
+  if ((!chapter && !isLoading) || !chapter?.isActive)
```
arkid15r: We still want to show it with a badge indicating its status.

arkid15r (changes requested):
Great migration! Let's make some improvements where possible

backend/apps/github/graphql/nodes/repository.py
abhayymishraa: somehow i didn't catch that !!

backend/apps/github/graphql/nodes/user.py
abhayymishraa: Some fields, like created_at and updated_at, require transformation before being exposed in GraphQL. If we try to directly use them as fields in the fields tuple, we encounter an issue because they are not directly available in the model in the expected format. Since we cannot modify the model fields directly, we use resolvers to adjust their representation before exposing them via GraphQL.

For contributions_count, I found a bug and will commit a fix. However, contributions_count does not exist as a direct field in the model, so we cannot simply include it in the fields tuple. Instead, we derive this value dynamically using idx_contributions, which aggregates data from RepositoryContributor. This computation requires a resolver since GraphQL fields can only map to actual model attributes unless explicitly overridden.
image

backend/apps/github/graphql/queries/repository.py
abhayymishraa: ok sure . i researched and found concatenation is faster and efficient for these usecases

backend/apps/owasp/graphql/nodes/committee.py
abhayymishraa: didn't quiet get that !! what do i need to do here we can do is 
```
def resolve_repositories_count(self, info):
        """Resolve project repositories count."""
        return 1 if self.owasp_repository else 0
   ```

frontend/src/pages/ChapterDetails.tsx
abhayymishraa: so do i need to implement the badge ?

arkid15r (approved):
LGTM

abhayymishraa:
Thanks @arkid15r

PR #770, opened 2025-02-07
Docker-Multi Stage Build

### LINKED ISSUE #734: Implement docker multi-stage build
Implement multi-stage build for frontend/backend test/staging/production images

Dishant1804:
Resolves #734 

Modified Dockerfiles that previously lacked support for multi-stage builds, now the necessary files support mullti-stage builds


backend/Dockerfile.test
```diff
@@ -26,3 +26,24 @@ COPY settings settings
...
+ENV FORCE_COLOR=1
+ENV PYTHONUNBUFFERED=1
+
+WORKDIR /home/owasp
+
+USER owasp
+
+COPY --from=builder /home/owasp /home/owasp
+
+CMD ["/bin/bash", "-c", "VENV_DIR=$(ls -d /home/owasp/.cache/pypoetry/virtualenvs/*) && $VENV_DIR/bin/pytest"]
```
Dishant1804: @arkid15r i've noticed that the ci/cd backend test is failing because i have removed the command of `poetry run test` from the `test-backend` command in `backend/makefile` and added the command in the the file it self, please help me out with this, should i modify the `ci-cd.yml` file in this case?
The same goes with frontend

yashgoyal0110:
@Dishant1804
Just wanted to know, have you compared build size and build time after making changes?

yashgoyal0110:
@Dishant1804 any update on this, are you still looking into it?

Dishant1804:
@yashgoyal0110 let @arkid15r reply with the correct approach with the ci/cd failure that is happening, then i am more than happy to share the metrics, also i was busy for couple of days as we have personal lives too. I would kindly request you to please have a bit patience :)

backend/Dockerfile.test
Dishant1804: the reason to ask this change was, the poetry should not be installed in the runtime stage as the image size will increase if we do so, which kills the purpose of multi-stage build **(for this particular test file)**. However we can consider an alternative by installing poetry in the `Dockerfile.test`'s runtime stage if we are okay with just one file being a bit large(it will still be fast and performant due to multi stage build)

arkid15r: Yes, we can live with that

Dishant1804:
@arkid15r made the changes, i let the poetry in the test file inorder to run pytest :)

arkid15r (approved):
Let's try this

Issue #746, opened 2025-02-04
Schema validation for URIs


arkid15r:
**Is your feature request related to a problem? Please describe.**
Our current validation for schema doesn't take care of URIs format compliance

**Describe the solution you'd like**
Extend validation based on jsonschema [docs](https://python-jsonschema.readthedocs.io/en/stable/). Look into schema version upgrade options. 

**Additional context**
https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/validators/

yashgoyal0110:
assign me

PR #774, opened 2025-02-07
Add schema URL validation

yashgoyal0110:
fixes: #746


yashgoyal0110:
@arkid15r all test cases are passing in local, then why it is failing in pipeline?


Screenshot 2025-02-08 at 1 03 37 AM

arkid15r:
> @arkid15r all test cases are passing in local, then why it is failing in pipeline?
> Screenshot 2025-02-08 at 1 03 37 AM

It seems the problem is with `blog` -- https://github.com/OWASP/Nest/actions/runs/13206822986/job/36871813084?pr=774

yashgoyal0110:
> > @arkid15r all test cases are passing in local, then why it is failing in pipeline?
> > Screenshot 2025-02-08 at 1 03 37 AM
> 
> It seems the problem is with `blog` -- https://github.com/OWASP/Nest/actions/runs/13206822986/job/36871813084?pr=774

Then why is it passing on local?
How it can be resolved?

arkid15r (changes requested):
I updated the part of code using existing validation package.
Could you make sure we validate all URI fields for both projects/chapters and make all files following the naming convention (see the suffixes `-empty`, `-null`, `-undefined`.

yashgoyal0110:
> I updated the part of code using existing validation package. Could you make sure we validate all URI fields for both projects/chapters and make all files following the naming convention (see the suffixes `-empty`, `-null`, `-undefined`.

yes sure

yashgoyal0110:
@arkid15r what suffixes should we use for non unique and non uri?

arkid15r:
> @arkid15r what suffixes should we use for non unique and non uri?

`non-unique` and `invalid`, we already have those example in the file names.

Make sure to get rid of type mention for the primitive types in the file names (-list, -url, etc) while keeping them for objects (e.g. `sposors-url-null`). `missing` should be replaced with `undefined`, `none` with `null`

yashgoyal0110:
> > @arkid15r what suffixes should we use for non unique and non uri?
> 
> `non-unique` and `invalid`, we already have those example in the file names.
> 
> Make sure to get rid of type mention for the primitive types in the file names (-list, -url, etc) while keeping them for objects (e.g. `sposors-url-null`). `missing` should be replaced with `undefined`, `none` with `null`

noted.

yashgoyal0110:
@arkid15r currently there's no format defined for repository attribute items, shouldn't it be 'uri' only?

arkid15r:
> @arkid15r currently there's no format defined for repository attribute items, shouldn't it be 'uri' only?

did you mean [this](https://github.com/OWASP/Nest/blob/c997e7f829679ca65cbc876c947c13325c2e9f40/schema/project.json#L33)?
use `repository-url-` prefix for testing the definition (while keeping `repositories-` prefix for the field tests)

yashgoyal0110:
> > @arkid15r currently there's no format defined for repository attribute items, shouldn't it be 'uri' only?
> 
> did you mean [this](https://github.com/OWASP/Nest/blob/c997e7f829679ca65cbc876c947c13325c2e9f40/schema/project.json#L33)? use `repository-url-` prefix for testing the definition (while keeping `repositories-` prefix for the field tests)

no i mean this. format 'uri' is not defined here
https://github.com/OWASP/Nest/blob/c997e7f829679ca65cbc876c947c13325c2e9f40/schema/project.json#L199

yashgoyal0110:
> > > > @arkid15r currently there's no format defined for repository attribute items, shouldn't it be 'uri' only?
> > > 
> > > 
> > > did you mean [this](https://github.com/OWASP/Nest/blob/c997e7f829679ca65cbc876c947c13325c2e9f40/schema/project.json#L33)? use `repository-url-` prefix for testing the definition (while keeping `repositories-` prefix for the field tests)
> > 
> > 
> > no i mean this. format 'uri' is not defined here
> > https://github.com/OWASP/Nest/blob/c997e7f829679ca65cbc876c947c13325c2e9f40/schema/project.json#L199
> 
> Look at my previous message.

there's some confusion i think

arkid15r (approved):
LGMT

PR #810, opened 2025-02-11
Routing Algolia Requests through the Backend Instead of Directly from frontend 

### LINKED ISSUE #769: Improve search bar funcionality based on real life requirements
OWASP Nest search/filtering/ranking/sorting is crucial part of the functionality. Finally we're at the point where we understand how it should be implemented.

We need to improve the search bar component tackling multiple problems:
 - search request should go through  the backend instead of direct algolia hits (more control, no unwanted interaction)
 - in-house developed auto-suggestions (again, more control over the behavior, but requires also more work)
 - text filtering for all entities (some work has been done [here](https://github.com/OWASP/Nest/pull/448))

This is combined into a single task to avoid merge conflicts and any misunderstandings leading to incompatibility and eventually more work.

Rajgupta36:
Resolves #769 

Current Implementation :-
- Routing algolia request through backend


backend/settings/urls.py
```diff
@@ -23,6 +24,9 @@
 urlpatterns = [
     path("graphql/", csrf_exempt(GraphQLView.as_view(graphiql=settings.DEBUG))),
     path("api/v1/", include(router.urls)),
+    path(
+        "api/v1/owasp/api/search/algolia", csrf_exempt(algolia_search), name="api-search-algolia"
```
arkid15r: WDYT about moving it into a separate `/idx/` endpoint to cover all index queries. Either `core` or `common` app?

arkid15r: I guess `core`

backend/tests/owasp/api/search/algolia_test.py
```diff
@@ -0,0 +1,74 @@
+import json
+from unittest.mock import Mock, patch
+
+import pytest
+from django.core.cache import cache
+
+from apps.owasp.api.search.algolia import algolia_search
+
+HTTP_200_OK = 200
```
arkid15r: Use `requests.codes`

frontend/src/api/fetchAlgoliaData.ts
```diff
@@ -1,45 +1,34 @@
...
+    const response = await fetch(`${API_URL}/owasp/api/search/algolia`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        indexName,
+        query,
+        page: currentPage,
+        hitsPerPage: indexName === 'users' ? hitsPerPage - 1 : hitsPerPage,
```
arkid15r: This should be on the users related component level

frontend/src/api/fetchAlgoliaData.ts
```diff
@@ -1,45 +1,34 @@
...
-      ...params,
-    }
-    if (filterKey) {
-      request.filters = `idx_key: ${filterKey}`
-    }
-
-    const { results } = await client.search({
-      requests: [request],
+    const response = await fetch(`${API_URL}/owasp/api/search/algolia`, {
+      method: 'POST',
```
arkid15r, line 14: Any reasons why not `GET`?

arkid15r (changes requested):
👍

frontend/src/api/fetchAlgoliaData.ts
Rajgupta36: Thanks for the feedback. before i passed everything from the frontend, the parameter length becomes too long for debugging . i'll change the request method and make the other changes as you suggested

arkid15r: Nah, that was a tricky comment :)
Let's discuss it in Slack.

backend/tests/owasp/api/search/algolia_test.py
Rajgupta36: ohh . thanks very much

frontend/src/hooks/useSearchPage.ts
```diff
@@ -55,7 +55,10 @@ export function useSearchPage<T>({
       params.set('order', order)
     }
 
-    setSearchParams(params)
+    if (searchParams.toString() !== params.toString()) {
+      setSearchParams(params)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Rajgupta36, line 61: if i  add `searchParams` to the dependency array, the issue persists again because when the URL parameters are updated, it triggers this useEffect

backend/tests/core/api/algolia_test.py
```diff
@@ -0,0 +1,72 @@
...
+    [
+        ("projects", "security", 1, 10, MOCKED_SEARCH_RESULTS),
+        ("chapters", "owasp", 2, 20, MOCKED_SEARCH_RESULTS),
+        ("users", "john", 1, 10, MOCKED_SEARCH_RESULTS),
+        ("committees", "review", 1, 10, MOCKED_SEARCH_RESULTS),
+        ("issues", "bug", 1, 10, MOCKED_SEARCH_RESULTS),
+    ],
+)
+@pytest.mark.usefixtures("_clear_cache")
+def test_algolia_search_valid_request(index_name, query, page, hits_per_page, expected_result):
```
arkid15r: Please make it a class

backend/tests/core/utils/match_test.py
```diff
@@ -0,0 +1,148 @@
+from apps.core.utils.params_mapping import get_params_for_index
+
+
+def test_get_params_for_issues():
```
arkid15r: Wrap it into a class too

arkid15r (commented):
LGTM

arkid15r (approved):
Let's migrate, LGTM 👍

PR #821, opened 2025-02-13
fix map and improve it using memoisation

abhayymishraa:
Fixes:

Add the PR description here.
- [x] fix map to use the map 
- [x] optimised using memoisation.
- [x] tested also chapter it will work on both chapter details and chapters 

[Screencast from 13-02-25 05:00:10 PM IST.webm](https://github.com/user-attachments/assets/6950c72d-06e7-4b2f-88ed-9662e45b49c5)


arkid15r (approved):
Works for me, thanks for the fix 👍

abhayymishraa:
it's my mistake i somehow forgot for chapter page : )

PR #822, opened 2025-02-13
Feature/contributing formatting update

moodyadi2006:
This file provides guidelines for setting up the OWASP Nest project, including configuring `.env` files for both the backend and frontend to ensure the application runs correctly. It also highlights the importance of saving `.env` files in **UTF-8 format without BOM (Byte Order Mark)** to prevent unexpected character errors during execution. Most modern text editors, such as VS Code, allow encoding selection when saving a file—choose **"UTF-8"** instead of **"UTF-8 with BOM"** to avoid issues.


arkid15r (approved):
LGTM, thanks for updating this!

moodyadi2006:
You're welcome! Happy to contribute. 😊

PR #824, opened 2025-02-14
Bug fix : handles URL params  correctly

Rajgupta36:
### Task - 
- [x] correct method so that  assure params of url is working 
- [x] reduced unnecessary API calls due to params rerendering

### Media

https://github.com/user-attachments/assets/bdfc12aa-8bc2-49cd-90c8-65b616319db1


arkid15r (approved):
LGTM 👍 
Thanks for fixing this!

PR #837, opened 2025-02-14
Map container is removed

### LINKED ISSUE #786: Map container UI improvement
chapter map container is not looking good because of unnecessary  background.

My suggestion: there should be no background and then it should be positioned accordingly

Image

osamaahmed17:
Resolves #786 
Chapter map container is updated to fix the unnecessary background and bad positioning. The container has been removed to adjust this.

https://github.com/user-attachments/assets/f9c20489-4245-4816-8d74-a0d5885dad81


osamaahmed17:
@arkid15r, feel free to give a review for this pull request.

kasya (approved):
Works great! Thanks!

Issue #723, opened 2025-02-03
Set up Nest documentation


arkid15r:
**Is your feature request related to a problem? Please describe.**
Currently we don't have a centralized docs system.

**Describe the solution you'd like**
Setup `docs/` directory with markdown based mkdocs system.

yashgoyal0110:
Hey @arkid15r, would like to contribute to this

priyankaChauhan2804:
hello, i would love to work on this, can you assign this to me?

PR #841, opened 2025-02-15
Feat/docs setup

yashgoyal0110:
fixes: #723 

- Initialised mkdocs setup for OWASP nest.
- Automated schema documentation process
- added contributing, code-of-conduct and general overview in documentation.


yashgoyal0110:
Hey @arkid15r,
what to do with these errors?

Screenshot 2025-02-16 at 12 34 16 AM

arkid15r:
> Hey @arkid15r, what to do with these errors?
> Screenshot 2025-02-16 at 12 34 16 AM

Make cspell happy -- fix typos or add to custom dict when needed.

arkid15r (commented):
Good work, I'll look into improving it tomorrow. No action required for now.

backend/pyproject.toml
```diff
@@ -43,6 +43,9 @@ pyyaml = "^6.0.2"
 requests = "^2.32.3"
 sentry-sdk = { extras = ["django"], version = "^2.20.0" }
 slack-bolt = "^1.22.0"
+json-schema-for-humans = "^1.3.4"
+mkdocs = "^1.6.1"
+sed = "^0.3.1"
```
arkid15r: We don't need these in staging/production. Please create a separate group `docs`

cspell/custom-dict.txt
```diff
@@ -74,3 +74,10 @@ xdg
 xdist
 xsser
 zsc
+kasya
```
arkid15r: Please keep the ordering consistent

mkdocs.yml
```diff
@@ -0,0 +1,60 @@
+site_name: OWASP NEST Docs
+nav:
+  - Home: index.md
+  - Code of Conduct: code-of-conduct.md
+  - Contributing: contributing.md
```
arkid15r, line 5: Is it possible to use the existing files from their current locations?

generate_schema_docs.sh
arkid15r, line 1: I'd prefer a simple Python script. Later we might need more customization and .sh will be harder to maintain.

mkdocs.yml
```diff
@@ -0,0 +1,60 @@
+site_name: OWASP NEST Docs
```
arkid15r:
```suggestion
site_name: OWASP Nest Documentation
```

arkid15r (changes requested):
We don't need that much emoji in the docs, I'd remove it all. Please also look into these:

mkdocs.yml
yashgoyal0110: I think the best option is to move files to docs folder

mkdocs.yml
```diff
@@ -1,4 +1,4 @@
-site_name: OWASP NEST Docs
+site_name: OWASP NEST Documentation
```
arkid15r: It's OWASP Nest

generate_schema_docs.py
```diff
@@ -0,0 +1,36 @@
+import os
+import subprocess
+
+INPUT_DIR = "./schema"
+OUTPUT_DIR = "./docs/schema"
+
+os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+for schema_file in os.listdir(INPUT_DIR):
```
arkid15r: Use `pathlib` for path related operations.

generate_schema_docs.py
arkid15r, line 1: Wrap this code with `__name__ == "__main__"` check to avoid running upon import.

docs/schema/common.md
arkid15r, line 1: This seems very high-level. It should either generate a proper table for the common entities or be excluded.

mkdocs.yml
arkid15r: Any details on why it's the best option?

docs/code-of-conduct.md
```diff
@@ -0,0 +1,91 @@
...
+    We pledge to act and interact in ways that contribute to an open, welcoming,
+    diverse, inclusive, and healthy community.
+
+## Community Standards
+
+### Positive Behaviors
+
+We encourage behaviors that contribute to a positive environment:
+
+- ✨ Demonstrating empathy and kindness toward other people
```
arkid15r: No emojis in the docs please

arkid15r (changes requested):

yashgoyal0110:
Hey @arkid15r 

MkDocs is structured to serve files from a dedicated docs/ directory. Keeping Markdown files outside docs/ requires workarounds like symbolic links or modifying mkdocs.yml, which can lead to compatibility issues

- Symbolic links don’t always work well on Windows without admin rights.
- Using docs_dir: . can break MkDocs’ expected behavior.
- Copying files before each build adds unnecessary steps

I think keeping both separate should be the approach as text formatting is also different for mkdocs files.

PR #853, opened 2025-02-16
Feature/improved schema definition

moodyadi2006:
I have just added a value check that if it is null or empty then return false and if valid URI then return None. project_test.py in tests contains all the checking for invalid, empty and null URI for link attributes


arkid15r (changes requested):
This PR contains irrelevant files. Please make sure to include only relevant code.

PR #867, opened 2025-02-17
extended the demo schema

### LINKED ISSUE #778: Extend `demo`  project schema attribute to allow multiple URLs
Extend `demo` schema from string to an array. Keep it still optional but if present it must contain at least 1 item.

priyankaChauhan2804:
Resolves #778

This PR updates the demo property in the schema to support both a single string and an array of strings. The property remains optional, but if provided as an array, it must contain at least one item.

**Changes Made**:
Updated the demo property to allow either:
1. A single string (maintaining backward compatibility).
2. An array of strings, ensuring at least one item is present.
3. Applied oneOf to enable flexibility while enforcing valid structure.


schema/project.json
```diff
@@ -171,8 +171,20 @@
     },
     "demo": {
       "description": "Optional URL to the project demo.",
-      "format": "uri",
-      "type": "string"
+      "oneOf": [
```
arkid15r, line 169: Just make it always an array and update tests accordingly.

Issue #870, opened 2025-02-17
Add YAML quality checks for OWASP Schema development


arkid15r:
**Is your feature request related to a problem? Please describe.**
We want to standardize the YAML files look with additional automated checks.

**Describe the solution you'd like**
Add a solution to both pre-commit and CI/CD to check the YAML files (`yamlfmt`, `yamllint` or both)

yashgoyal0110:
Can I work on this?

priyankaChauhan2804:
could i work on this?

Yashgupta9330:
@arkid15r I want to work on this issue

PR #891, opened 2025-02-19
added yaml file checks

yashgoyal0110:
fixes: #870


.pre-commit-config.yaml
```diff
@@ -1,36 +1,43 @@
...
+- repo: https://github.com/djlint/djLint
+  rev: v1.35.2
+  hooks:
+  - id: djlint-reformat-django
+    files: templates/.*.html
+    entry: djlint --reformat
+    types:
+    - html
+
+- repo: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
```
arkid15r: What's your reasoning behind using this instead of https://github.com/google/yamlfmt?

.github/FUNDING.yml
```diff
@@ -1 +1,2 @@
+---
```
arkid15r: I don't think we need those in every yaml file

.github/ISSUE_TEMPLATE/config.yml
arkid15r, line 1: This file should have stayed the same after auto formatting

arkid15r (changes requested):

.pre-commit-config.yaml
yashgoyal0110: not any specific reason, but now i have switched my approach to `https://github.com/google/yamlfmt` and also removed '---' from every file

yashgoyal0110: you can review now

arkid15r: Please use the original google formatter with indent size 2

.pre-commit-config.yaml
```diff
@@ -1,36 +1,43 @@
...
+            files: templates/.*.html
+            entry: djlint --reformat
+            types:
+                - html
+
+    - repo: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
+      rev: 0.2.1
+      hooks:
+          - id: yamlfmt
+            args: [--preserve-quotes, --implicit_start]
```
arkid15r: The options should be in a separate config if possible

.pre-commit-config.yaml
```diff
@@ -1,36 +1,43 @@
...
+            exclude: ".github/ISSUE_TEMPLATE/config.yml"
```
arkid15r: Remove this, my comment was about the formatting style -- not the file exclusion.

.github/ansible/production/nest.yaml
```diff
@@ -1,106 +1,106 @@
...
-    - name: Retart services
-      shell:
-        cmd: docker compose up -d
-
-    - name: Index data
-      async: 1800 # 30 minutes
-      poll: 0
-      shell: |
-        make index-data > /var/log/nest/production/index-data.log 2>&1
+    - name: Deploy Nest to Production
```
arkid15r: Why all these indents?

arkid15r (changes requested):
The goal is to keep YAML files clean and well-formatted.

.pre-commit-config.yaml
yashgoyal0110: i have moved options to separate file and now using google yaml formatter. sorry for the inconvenience

.github/ansible/production/nest.yaml
```diff
@@ -21,7 +21,7 @@
 
     - name: Update backend makefile for the production environment
       shell:
-        cmd: "sed -i 's/\\bnest-backend\\b/production-nest-backend/' ~/backend/Makefile"
+        cmd: sed -i 's/\bnest-backend\b/production-nest-backend/' ~/backend/Makefile
```
arkid15r: I see the the changes that look like auto-formatting. But when I change them back and run `make check` it doesn't auto-format it again.

Can you explain what I'm missing here?

yashgoyal0110: i executed `pre-commit run --all-files` this cmd

arkid15r: `make check` runs `pre-commit run -a` too

yashgoyal0110: even i tried it on my local, you are right maybe some other code dit it while i was trying out few things. I have fixed it

arkid15r (approved):
This may break something so let figure that out

PR #892, opened 2025-02-19
Updated the test cases and Extended the demo Schema

### LINKED ISSUE #778: Extend `demo`  project schema attribute to allow multiple URLs

priyankaChauhan2804:
Resolves #778

**Make demo field consistently use array type**

**Changes**:

1. Modified the schema to make the demo field always use an array type for URLs
2. Updated corresponding test cases to align with the new array-only format
3. Removed the oneOf validator to simplify the schema structure
**Why**:
This change standardizes how demo URLs are handled in the schema, making it more consistent and easier to validate. Previously, the schema allowed for potential confusion with null values. Now, it will always expect an array of URIs, simplifying both the schema and how consumers handle this field.

**Testing**:

Updated demo-null.yaml test case
Verified that existing valid demo URLs continue to work
Confirmed array validation works as expected

### DIFF
--- a/schema/project.json
+++ b/schema/project.json
@@ -166,8 +166,16 @@
     },
     "demo": {
       "description": "Optional URL to the project demo.",
-      "format": "uri",
-      "type": "string"
+      "oneOf": [
+        {
+          "type":"array",
+          "items": {
+            "type": "string",
+            "format": "uri"
+          },
+          "minItems": 1
+        }
+      ]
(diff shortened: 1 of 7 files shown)


arkid15r (approved):
Thanks for improving this 👍

PR #911, opened 2025-02-22
Added test-all command

samyak003:
Added `make test-all` command to run all the tests using one single command instead of running individual test commands.


arkid15r (approved):
LGTM 👍

arkid15r (approved):

Issue #910, opened 2025-02-22
Fix "Sort By" click trigger


arkid15r:
**Describe the bug**
After merging #884 the "sort by" text is [no longer triggers the dropdown](https://github.com/OWASP/Nest/pull/884/files#diff-b1955de76ea3870e960eeb58a0bdf54eb575f26a5581a62d9c89d541adbcf57aR41)

**To Reproduce**
Try clicking on "Sort By" https://nest.owasp.org/projects

**Expected behavior**
It should trigger the dropdown options as it used to do

**Are you going to work on fixing this?**
No.

Raghavg18:
hey @arkid15r can u assign this to me?

RedK22:
I would like to work on this. Can you please assign me?

PR #912, opened 2025-02-22
Fixed sort button not triggering dropdown

RedK22:
Fixes Issue #910 
Fixed the "Sort By" text not triggering dropdown options on click.


kasya (approved):
@RedK22 Works great! Thanks for updating this!

PR #913, opened 2025-02-22
implemented matching model and Added slack models

### LINKED ISSUE #501: Design and implement metadata matching system
Design and implement a way to match community users to project leaders. For leaders we typically have first name and last names (sometimes just partial full name with some comment) so the workflow might need a manual involvement in some cases. But it seems we can easily match exact match cases while keeping "unsure" mappings as suggestions for further human processing.

The `owasp` app models should be modified to include `leaders` m2m relations to `github.user`. The command/job should suggest/populate chapter/committee/project leaders based on a certainty level.

Another area that the system should cover is entity / Slack channel mapping. Same functionality just in a different context.
For this we'll need to create a new model in `slack` app (probably, `Channel`)

I believe in order to achieve the best results the system should suggest matches that need to be verified/approved manually.

This is just an abstract problem description to start a discussion regarding the implementation.

              it's good idea but We don't have any links in the project index, and we can't identify a specific user from name.

_Originally posted by @Rajgupta36 in https://github.com/OWASP/Nest/pull/470#discussion_r1916229870_

Rajgupta36:
Resolves #501 

### Tasks  
- [x] Created Slack models (`Member`, `Channel`, and `Workspace`).  
- [x] Established an M2M relationship between `User` and `Project`, `Chapter`, and `Committee` models.  
- [x] Created a job to populate Slack data using the workspace bot token.  
- [x] Implemented a matching algorithm for user suggestions.


backend/apps/owasp/models/common.py
```diff
@@ -173,6 +173,20 @@ def get_metadata(self):
                 extra={"repository": getattr(self.owasp_repository, "name", None)},
             )
 
+    # M2M
```
arkid15r: I wonder how you ended up w/ this location as the best one for those two new attrs? :)

backend/apps/slack/management/commands/load_slack_data.py
```diff
@@ -0,0 +1,106 @@
...
+            try:
+                cursor = None
+                while True:
+                    response = client.conversations_list(
+                        types="public_channel,private_channel", limit=1000, cursor=cursor
+                    )
+                    self._handle_slack_response(response, "conversations_list")
+
+                    for channel in response["channels"]:
+                        Channel.objects.update_or_create(
```
arkid15r: We prefer using bulk save approach

backend/apps/slack/models/workspace.py
```diff
@@ -0,0 +1,21 @@
...
+class Workspace(TimestampedModel):
+    """Slack Workspace model."""
+
+    class Meta:
+        db_table = "slack_workspaces"
+        verbose_name_plural = "Workspaces"
+
+    slack_workspace_id = models.CharField(verbose_name="Workspace ID", max_length=50, unique=True)
+    name = models.CharField(verbose_name="Workspace Name", max_length=100, default="")
+    bot_token = models.CharField(verbose_name="Bot Token", max_length=200, default="")
```
arkid15r: Storing secrets as plain text is always a bad idea. The bot gave you a hint about that too, btw.
As I haven't worked with `encrypted_fields` I preferred to just extract bot token to env.

arkid15r (approved):
LGTM

Merging this to clear up the way for other PRs and streamline the PR queue. I'll check/improve the actual functionality later.

backend/apps/owasp/models/common.py
Rajgupta36: actually at that time when i created pr we had only project,chapter,committee and event model . all three models except event model  inherited the common module so i thought why not created a common code to reduce line.but i didn't think about future scope . my bad

backend/apps/slack/models/workspace.py
Rajgupta36: Actually, for testing purposes, I’m manually storing the bot token in our database, so we’re not encrypting or decrypting it at the moment.
One possible approach is to store all the data in JSON format, then create a command to encrypt it before saving it to the database. When needed, we can decrypt it and use it.

does this approach sounds good, @arkid15r?

Rajgupta36: As per discussion i created it to support multiple workspace . if we have only OWASP then we can extract bot_token  from env .

PR #859, opened 2025-02-16
Extraction of common object definitions from chapter/projects schema

AbhayTopno:
Resolves #784 

Screenshot 2025-02-17 040658
project.json

Screenshot 2025-02-17 040724
chapter.json

PR #942, opened 2025-02-26
Implementation Of Committee Schema

### LINKED ISSUE #873: Desing and implement OWASP Committee schema

Create committee.json similar to [chapter.json](https://github.com/OWASP/Nest/blob/main/schema/chapter.json) and [project.json](https://github.com/OWASP/Nest/blob/main/schema/project.json). This is a more creative task with no strict requirements -- it's your goal to figure out what should be included in schema. A great to play with real life entity JSON schema.

Some starting points:
  - https://github.com/OWASP/www-committee-chapter
  - https://github.com/OWASP/www-committee-project
  - https://github.com/OWASP/www-committee-education-and-training
  - https://github.com/OWASP/www-committee-events

Dishant1804:
Resolves #873 

Designed and implemented the schema for OWASP Committee and added positive and negative test cases for the same


Dishant1804:
@arkid15r I Have implemented the committee schema with robust tests, Let me know if any changes are required :)

Dishant1804:
Hey @arkid15r the cause of failure of `common/person_test.py` file is now known, the failure image is below

image

There was a test was failing the common schema extraction it part of #859 specifically `name-empty.yaml` file, i have fixed the issue by adding minlength constraint to the person's name in `common.json`, which now expects the name to be minimum of length `3`, in this way we assert that the name is short which passes the case as it was a negative test case.
I'm not sure how this slipped through or was overlooked in the #859  PR, but this PR resolves that issue too.

schema/tests/data/committee/negative/logo-large-empty.yaml
arkid15r, line 1: You don't need to test the logo internal structure as we do that in common.json related tests. You need to test logo for its committee specific rules.

schema/tests/data/committee/positive/optional-requirements.yaml
arkid15r, line 1: optioinal-properties.yaml

schema/tests/committee_test.py
```diff
@@ -0,0 +1,81 @@
...
+from pathlib import Path
+
+import pytest
+import yaml
+
+from tests.conftest import tests_data_dir
+from utils.schema_validators import validate_data
+
+
+def test_positive(committee_schema):
```
arkid15r: Negative on top please

schema/common.json
```diff
@@ -52,36 +52,38 @@
...
-        "github": {
-          "description": "GitHub username",
-          "pattern": "^[a-zA-Z0-9-]{1,39}$",
-          "type": "string"
-        },
-        "name": {
-          "description": "Full name",
-          "minLength": 5,
-          "type": "string"
+        "person": {
```
arkid15r: Do you change anything here?

schema/committee.json
```diff
@@ -0,0 +1,255 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://raw.githubusercontent.com/OWASP/Nest/main/schema/committee.json",
+  "$defs": {
```
arkid15r, line 4: Event/logo definitions are in common.json now.

schema/committee.json
```diff
@@ -0,0 +1,255 @@
...
+      },
+      "required": [
+        "small",
+        "medium",
+        "large"
+      ],
+      "type": "object",
+      "additionalProperties": false
+    },
+    "meeting_minute": {
```
arkid15r:
```suggestion
    "meeting_minutes": {
```

schema/committee.json
```diff
@@ -0,0 +1,255 @@
...
+      "type": "string"
+    },
+    "events": {
+      "description": "List of events organized or sponsored by the committee.",
+      "items": {
+        "$ref": "#/$defs/event"
+      },
+      "type": "array",
+      "uniqueItems": true
+    },
```
arkid15r: I'm not sure if it's something that will be widely used. Let's remove.

schema/committee.json
```diff
@@ -0,0 +1,255 @@
...
+            "$ref": "common.json#/definitions/person"
+          },
+          {
+            "properties": {
+              "role": {
+                "description": "Role of the leader.",
+                "enum": [
+                  "chair",
+                  "vice chair",
+                  "Secretary"
```
arkid15r: Some naming consistency please

schema/committee.json
```diff
@@ -0,0 +1,255 @@
...
+      },
+      "type": "array",
+      "uniqueItems": true
+    },
+    "name": {
+      "description": "The unique name of the committee.",
+      "minLength": 8,
+      "type": "string"
+    },
+    "pitch": {
```
arkid15r: This is mostly for projects, let's use scope/description for that.

schema/committee.json
```diff
@@ -0,0 +1,255 @@
...
+      "description": "A short pitch or tagline summarizing the committee's mission and objectives.",
+      "minLength": 8,
+      "type": "string"
+    },
+    "resources": {
```
arkid15r, line 133: Any examples of use in a real world?

schema/committee.json
```diff
@@ -0,0 +1,255 @@
...
+      "minItems": 1,
+      "type": "array",
+      "uniqueItems": true
+    },
+    "scope": {
+      "description": "Scope and purpose of committee",
+      "items": {
+        "type": "string"
+      },
+      "type": "string"
```
arkid15r: Why both items and type string?

arkid15r (changes requested):
Good work, It needs some improvements:

schema/committee.json
Dishant1804: i forgot to remove items , i'll remove it

schema/common.json
Dishant1804: yes i modified the additional properties of person to true, it was because the leaders are the person who have specific role so i used the person and extended with the enums, which requires additional properties to be true

schema/committee.json
Dishant1804: @arkid15r there were resources md file in the repos that were mentioned in the issue, i tried generalise it here in this property, also attaching the reference to where the resources are : 
[committee-events](https://github.com/OWASP/www-committee-events/blob/master/tab_resources_for_events.md) , [committee-chapters](https://github.com/OWASP/www-committee-chapter/blob/master/tab_resources_for_chapters.md) , [committe-project](https://github.com/OWASP/www-committee-project/blob/master/tab_resources.md)
we can remove it if we don't need it, but in my opinion we should keep it as committees are having the resources section and some committees might have resources in future as well

schema/common.json
arkid15r: I suggest keeping `person` w/o changes and create a committee scoped leader definition that includes person and additional fields you need.

No definition should have additionalProperties set to true.

arkid15r (approved):
It's good enough to merge and improve later. Thanks for mapping committee attributes to schema!

PR #948, opened 2025-02-27
feat: made org and location clickable on user detail page

### LINKED ISSUE #946: ui: improve user detail page
The user detail page is very good. But we can improve it further. See the screenshot at end. I suggest to make `oracle` and `ccextractor` clickable like how it is on GitHub. Then add website or LinkedIn of user. And timezone of user.

- Add clickable links for user orgs
- Add LinkedIn/website link on users page
- Time Zone of user ( a very good feature as our community is distributed around the globe)

- Make the location clickable? (Debatable, not much useful actually but github has it)

**Additional context**
Here is the screenshot -
Image

jaiakash:
Resolves #946

Add clickable org and location on the user details page.

Screenshot 2025-02-27 at 9 12 20 PM


arkid15r (approved):
@jaiakash this works and resolves the issue you suggested.
Do you think we could extend this and cover @github-entity style accounts from user's bio too?

Screenshot 2025-02-27 at 11 57 11 AM

jaiakash:
> @jaiakash this works and resolves the issue you suggested.
> 
> Do you think we could extend this and cover @github-entity style accounts from user's bio too?
> 
> 
> 
> Screenshot 2025-02-27 at 11 57 11 AM
> 
> 

I mean we can but not always that would be github account. Sometime it may not be github org/account. Plus if we are manually parsing text after @ as link then they may be false positive.

arkid15r:
> > @jaiakash this works and resolves the issue you suggested.
> > Do you think we could extend this and cover @github-entity style accounts from user's bio too?
> > Screenshot 2025-02-27 at 11 57 11 AM
> 
> I mean we can 🤔🤔 but not always that would be github org/account.
> 
> Plus if we are manually parsing text after @ as link then there might be false positive.

You're right but it's mostly GH accounts. Would you like to create a separate issue and work on this too?

jaiakash:
> You're right but it's mostly GH accounts. Would you like to create a separate issue and work on this too?

fine no issue. i will send create issue and send PR.

PR #951, opened 2025-02-27
fix: clickable org if there is any @ text in user bio

### LINKED ISSUE #950: Adding clickable links for user bio
Related #948 

Add clickable link for the GS accounts in the bio of the user.

jaiakash:
Resolves #950 

If user bio has any @ text then it will redirect to GitHub.
Screenshot 2025-02-28 at 1 56 25 AM


jaiakash (commented):
Added units tests as well.

frontend/src/pages/UserDetails.tsx
```diff
@@ -150,7 +150,26 @@ const UserDetailsPage: React.FC = () => {
               </div>
             </div>
             <div className="px-6 py-6">
-              {user.bio && <p className="text-lg text-gray-700 dark:text-gray-300">{user.bio}</p>}
+              {user.bio && (
+                <p className="text-lg text-gray-700 dark:text-gray-300">
+                  {user.bio.split(' ').map((word, index) => {
```
kasya: If there's a dot after the link like in this example - the dot is added to the link, which makes it a `404`. 
Screenshot 2025-03-01 at 5 58 41 PM
There could also be a `;!?` or any other thing like that. 
Could you fix this?

kasya (changes requested):

frontend/src/pages/UserDetails.tsx
```diff
@@ -150,7 +150,27 @@ const UserDetailsPage: React.FC = () => {
...
+                    const mentionMatch = word.match(/^@(\w+)/)
```
kasya: This works for the link but it still shows the dot highlighted in the UI. You should be able to fix that by updating both regexes here and at the top. The taking out the correct group item. 
Also, could we move this logic out of the template? It's getting messy.
Screenshot 2025-03-02 at 9 56 30 AM

kasya (changes requested):

jaiakash:
Screenshot 2025-03-03 at 10 14 47 PM

here is the final version
- removes dot if its present at last of username/org name
- decouple the logic outside the UI code

kasya:
> here is the final version
> 
> * removes dot if its present at last of username/org name
> * decouple the logic outside the UI code

Why did you remove the dot? 🤔 What about other punctuation marks?
I thought you would refactor the method to just not include them in links, but not remove them all together - this is still a sentence that should have punctuation marks.
Here's some code for you to start with. I'm not sure if it's 100% covering what we need, but it should help:
```
{user.bio && (
    <p className="text-lg text-gray-700 dark:text-gray-300">
      {user.bio.split(/(\s+)/).map((part, index) => {
        const mentionMatch = part.match(/^@([\w-]+)([.,!?)])?$/);
        if (mentionMatch) {
          return (
            <React.Fragment key={index}>
              <Link
                href={`https://github.com/${mentionMatch[1]}`}
                target="_blank"
                rel="noopener noreferrer"
                className="text-blue-500 hover:underline"
              >
                @{mentionMatch[1]}
              </Link>
              {mentionMatch[2] && <span>{mentionMatch[2]}</span>} {}
            </React.Fragment>
          );
        }
        return <span key={index}>{part}</span>;
      })}
    </p>
  )}
  ```

**Review by kasya — CHANGES_REQUESTED** [review]
Left a suggestion.

**Comment by jaiakash** [comment]
> Why did you remove the dot? 🤔 What about other punctuation marks? I thought you would refactor the method to just not include them in links, but not remove them all together - this is still a sentence that should have punctuation marks. Here's some code for you to start with.

The dot at end of such sentances wont make much difference, that was my assumption. But i think what you are saying is correct, there can be any other puntuation mark as well. 
So i am changing them to include other puntuation marks (of course except @) and show them in the bio.
I hope thats what is expected, right @kasya ?

**Comment by arkid15r** [comment]
> > Why did you remove the dot? 🤔 What about other punctuation marks? I thought you would refactor the method to just not include them in links, but not remove them all together - this is still a sentence that should have punctuation marks. Here's some code for you to start with.
> 
> The dot at end of such sentances wont make much difference, that was my assumption. But i think what you are saying is correct, there can be any other puntuation mark as well. So i am changing them to include other puntuation marks (of course except @) and show them in the bio. I hope thats what is expected, right @kasya ?

I'm not sure about your approach, the point here is that the content must remain the same (event punctuation). The only change is that `@handles` should become links.

**Comment by jaiakash** [comment]
> I'm not sure about your approach, the point here is that the content must remain the same (event punctuation). The only change is that `@handles` should become links.

yes, that's what I intended, to convert them to links. But the punctuation mark (except @) should be as it is.

File: frontend/src/pages/UserDetails.tsx [hunk 3]
```diff
@@ -91,6 +91,30 @@ const UserDetailsPage: React.FC = () => {
     )
   }
 
+  const formattedBio = user?.bio?.split(' ').map((word, index) => {
```
kasya, line 94: Yess! Thank you!

kasya (approved):
Nice!

PR #962, opened 2025-02-28
added focus functionality on home page search bar

### LINKED ISSUE #929: Improve home page search bar behavior
The idea is to have the suggestions dropdown to come again automatically after these steps:
 1. User enters something into the search bar
 1. The suggestions appear
 1. User changes focus to other element (the entered content is still in the search bar)
 1. When user changes focus again to the search bar and it contains some text the suggestions box should appear again automatically (it currently doesn't)

kayashth1:
Resolves #929

## Summary by CodeRabbit

- **New Features**
	- Enhanced the search input behavior to automatically display suggestions when the search field is focused with text entered.
	- Cleared any highlighted suggestions when there is no valid input, providing a smoother and more intuitive search experience for users.
	- Added functionality to remove focus from the search input when the 'Escape' key is pressed.


frontend/src/components/MultiSearch.tsx
```diff
@@ -154,6 +154,14 @@ const MultiSearchBar: React.FC<MultiSearchBarProps> = ({
     setHighlightedIndex(null)
   }
 
+  const handleFocusSearch = () => {
```
kasya, line 159: This works great if you manually change focus by clicking out of the search bar. However, if you press `esc` as the dropdown suggests to close it - clicking on the search bar does not trigger dropdown again.

Can we update this?

kasya (changes requested):
Thank you for working on this! 
One issue I noticed ⬇️

frontend/src/components/MultiSearch.tsx
kayashth1: yes, there was a problem, but now I have corrected this.

kasya (approved):
LGTM!

PR #970, opened 2025-03-01
Adding mkdocs

yashgoyal0110:
fixes: #723


yashgoyal0110:
Hey @arkid15r 
Actually due to some errors i deleted my old fork, and created new fork.
here is the new PR for docs with same changes.
sorry for inconvenience caused

yashgoyal0110:
MkDocs is structured to serve files from a dedicated docs/ directory. Keeping Markdown files outside docs/ requires workarounds like symbolic links or modifying mkdocs.yml, which can lead to compatibility issues

- Symbolic links don't always work well on Windows without admin rights.
- Using docs_dir: . can break MkDocs' expected behavior.
- Copying files before each build adds unnecessary steps

I think keeping both separate should be the approach as text formatting is also different for mkdocs files.

yashgoyal0110:
@arkid15r is something wrong with CI/CD?
why e2e is failing for this PR?

arkid15r:
> MkDocs is structured to serve files from a dedicated docs/ directory. Keeping Markdown files outside docs/ requires workarounds like symbolic links or modifying mkdocs.yml, which can lead to compatibility issues
> 
>     * Symbolic links don't always work well on Windows without admin rights.
> 
>     * Using docs_dir: . can break MkDocs' expected behavior.
> 
>     * Copying files before each build adds unnecessary steps
> 
> 
> I think keeping both separate should be the approach as text formatting is also different for mkdocs files.

The problem here is that we still want readme, code of conduct, and maybe other files to be where they are now due to their special place from GitHub perspective
Screenshot 2025-03-06 at 5 39 44 PM

I don't know what's the best approach here yet. However, we'll unlikely move those files to the docs (copying may still work for us)

yashgoyal0110:
> > MkDocs is structured to serve files from a dedicated docs/ directory. Keeping Markdown files outside docs/ requires workarounds like symbolic links or modifying mkdocs.yml, which can lead to compatibility issues
> > ```
> > * Symbolic links don't always work well on Windows without admin rights.
> > 
> > * Using docs_dir: . can break MkDocs' expected behavior.
> > 
> > * Copying files before each build adds unnecessary steps
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think keeping both separate should be the approach as text formatting is also different for mkdocs files.
> 
> The problem here is that we still want readme, code of conduct, and maybe other files to be where they are now due to their special place from GitHub perspective Screenshot 2025-03-06 at 5 39 44 PM
> 
> I don't know what's the best approach here yet. However, we'll unlikely move those files to the docs (copying may still work for us)

File location is not the only issue, another problem is the way text is formatted for mkdocs file, so if we use same file then how it's gonna work?

yashgoyal0110:
Hey @arkid15r 
any updates on this?

arkid15r (changes requested):
I was going to merge this but decided to update the schema first (using the `generate_schema_docs.py`. I think we need a Makefile target for that with a separate (preferably) or same docker file as for the backend. Let's get it to at least that level, merge and improve incrementally.

yashgoyal0110:
Hey @arkid15r 
I have created a separate docker file and added makefile for schema.

Makefile
```diff
@@ -48,3 +48,6 @@ update-dependencies: \
 update-nest-app-dependencies: \
 	update-backend-dependencies \
 	update-frontend-dependencies
+
+generate-schema-docs:
```
arkid15r: We don't need the compose file for this -- just to be able to build an image.
Please also extend this make target to generating all docs, not the schema update only.

arkid15r (changes requested):
Any solution you deliver should be usable and structured in a way similar to other automations we have. Please create a command that generates the documentation in an isolated environment using properly structured configuration files.

This should work at least locally as the next step is running it as a CI/CD part. I hope you can see the bigger picture now.

Makefile
yashgoyal0110: ok got it.

arkid15r:
@yashgoyal0110 any updates on this? Are you interested in finishing the task?

yashgoyal0110:
> @yashgoyal0110 any updates on this? Are you interested in finishing the task?

Hey @arkid15r ,
Just busy with gsoc and proposal stuff will finish it soon, in the mean time if someone is interested then they may work on it, otherwise i will try to finish it soon

arkid15r (approved):

PR #986, opened 2025-03-03
updated metadata for the static page

abhayymishraa:
- [x] updated html tags to render a static image on the links preview


arkid15r (approved):
Let's try this.

Issue #972, opened 2025-03-02
Move leaders data fetch logic from scraping to github files parsing


arkid15r:
**Is your feature request related to a problem? Please describe.**
Some recent projects/chapters leaders fields are empty.

**Describe the solution you'd like**
Change [current logic](https://github.com/OWASP/Nest/blob/759b86b4cc2c87e3a1bee1d53ec8bc4d5fba5258/backend/apps/owasp/management/commands/owasp_scrape_chapters.py#L66) to getting data directly from GitHub leaders.md files similar to [index.md content](https://github.com/OWASP/Nest/blob/759b86b4cc2c87e3a1bee1d53ec8bc4d5fba5258/backend/apps/owasp/models/common.py#L92).

**Are you going to work on implementing this?**

- [ ] Yes
- [x] No

samyak003:
@arkid15r can i work on this issue?

samyak003:
I don't want a 7 day deadline. Just give me a 2 or 3 day deadline. If I am not able to show any progress then you can assign this issue to someone else.

PR #990, opened 2025-03-03
Changed the logic of fetching leader's data

samyak003:
Fixes: #972


arkid15r (changes requested):
Could you move the logic to [RepositoryBasedEntityMode](https://github.com/OWASP/Nest/blob/6188367c9d485b45257315fa16a31e86fd4e7bb4/backend/apps/owasp/models/common.py#L26)::get_leaders

upd: also some tests would be great 👍

backend/apps/owasp/management/commands/owasp_scrape_chapters.py
arkid15r, line 1: For all entities it's no longer a part of scraping process. It's now within GitHub processing part. We need to structure it properly.

backend/tests/owasp/scraper_test.py
```diff
@@ -108,16 +108,6 @@ def test_initialization_not_found(self, mock_session):
 
         assert scraper.page_tree is None
 
-    def test_get_leaders_no_leaders(self, mock_session):
```
arkid15r, line 111: I believe this test needs to be refactored too based on the new logic and the code location.

backend/apps/owasp/models/common.py
```diff
@@ -146,6 +147,39 @@ def get_top_contributors(self, repositories=()):
             .order_by("-total_contributions")[:TOP_CONTRIBUTORS_LIMIT]
         ]
 
+    def get_leaders(self, repository):
```
arkid15r: Please add tests for the new method.

arkid15r (changes requested):
Looks good, I didn't run it yet. Let's take care of the structural inconsistency first.

arkid15r (approved):
LGTM

PR #1009, opened 2025-03-04
Improves home page chapter map

### LINKED ISSUE #928: Improve home page chapters map
We want all chapters to be shown on the map (in case user wants to explore more) but the default zoom should be on the local/nearest user chapters.

I'm not sure about this feature request feasibility so this requires a bit of research first. We have geo ranking via Algolia index and could use the N nearest chapters to figure out the coordinates for narrowing down the default zoom. All chapters coordinates are available via graphQL and from index too.

Yashgupta9330:
Resolves #928


arkid15r (changes requested):
I don't think this PR addresses the issue. The idea is to render all chapters on the page and set a zoom based on the user's IP address coordinates to show the nearest chapters. When zoomed out all other chapters will be shown too (now we don't have them).

Please also note this issue is about Home page map.

Yashgupta9330:
> I don't think this PR addresses the issue. The idea is to render all chapters on the page and set a zoom based on the user's IP address coordinates to show the nearest chapters. When zoomed out all other chapters will be shown too (now we don't have them).
> 
> Please also note this issue is about Home page map.

I thought it was chapters page map

kasya (commented):
@Yashgupta9330 will you be working on updating this?

Yashgupta9330:
> @Yashgupta9330 will you be working on updating this?

yes i am working on it

frontend/src/components/ChapterMap.tsx
```diff
@@ -81,12 +94,34 @@ const ChapterMap = ({
 
     map.addLayer(markerClusterGroup)
 
-    if (bounds.length > 0) {
-      map.fitBounds(bounds as L.LatLngBoundsExpression, { maxZoom: 10 })
+    // Determine map view based on 6th index (index 5)
+    try {
```
kasya: Could you elaborate a bit more on the logic in this block? I think it works, but I'm just not sure what's going on.
Why do we set view to a sixth chapter? Why not 1st for example? 🤔

frontend/src/pages/Home.tsx
```diff
@@ -55,7 +55,7 @@ export default function Home() {
         indexName: 'chapters',
         query: '',
         currentPage: 1,
-        hitsPerPage: 25,
+        hitsPerPage: 1000,
```
kasya, line 58: Why do we increase this so much? 🤔

kasya (commented):
Looks good! 
Just want to clarify some things! ⬇️

frontend/src/pages/Home.tsx
Yashgupta9330: because I want to show all chapter also previously only 5 chapters was fetching

frontend/src/components/ChapterMap.tsx
Yashgupta9330: I chose the sixth location as the reference point for the default zoom because the data is sorted in a way where the sixth location provides a good central point to cover the surrounding five locations on the map. Zooming to this position ensures that all five prior locations are visible while still maintaining an optimal view of the area. Using the first index would focus too much on a single location, potentially causing some of the other important locations to be out of the initial zoom view.

arkid15r:
@Yashgupta9330 this PR doesn't address the issue -- it has nothing to do w/ the chapters page and for the main page you just increased the number of chapters shown.

I'm closing this out.

arkid15r:
I reworked the logic a bit and how map looks at the main page. I'll file another ticket to improve it on the Chapters page.

arkid15r (approved):
Let's try this

PR #1018, opened 2025-03-05
migrate: footer component to chakra-ui

### LINKED ISSUE #702: Migrating Footer to chakra UI
As we are moving forward with chakra UI migrations, I want to migrate footer component and user card component  to chakra UI

YourBroCode:
Resolves #702 

migrate footer components to the chakra UI.


arkid15r (commented):
@YourBroCode could you check if it makes sense to apply any of coderabbit's comments?

YourBroCode:
> @YourBroCode could you check if it makes sense to apply any of coderabbit's comments?


Checked it. it's suggesting that  we are using Chakra UI so , we could utilise the prop based styling instead of using tailwind classes. we can migrate it component by component in a separate issue section.

arkid15r (approved):
It works for me, let's merge it 👍

PR #1043, opened 2025-03-07
Owasp.org post preview

### LINKED ISSUE #994: Implement owasp.org posts previews
**We want to show news/opinions post previews (title, date, and author info would be enough) on the OWASP Nest main page with links to the original posts on owasp.org site.**  

### Summary of Tasks  

**1. Create the `Post` model** in the `owasp` app  
   - Fields: `title`, `date`, `author`, `author_image`, `url`  
   - Store only essential metadata  

**2. Implement a Django management command (`sync_posts`)**  
   - Fetch markdown files from [OWASP GitHub posts](https://github.com/OWASP/owasp.github.io/tree/main/_posts)
   - Parse YAML to extract metadata  
   - Insert or update posts in the database  

**3. Create an GraphQL endpoint** to fetch recent posts  
   - Returns the 5 most recent posts (sorted by `date`)  
   - Fields: `title`, `date`, `author`, `author_image`, `url`  

**4. Implement the frontend component**  
   - Fetches the latest 5 posts from the API  
   - Displays:  
     - Post title (linked to OWASP.org)  
     - Date  
     - Author  
     - Author image  
   - Responsive design  

**5. Refactor existing functionality (NestBot)**  
 - Migrate to the new data source instead of the current one.

**6. Add tests**  
   - Model tests (validations, constraints)  
   - Management command tests (mocking GitHub API calls)  
   - API endpoint tests (pagination, response validation)  
   - Frontend tests (renders correctly, handles API responses)  

- [ ] Yes
- [x] No

Dishant1804:
Resolves #994

- [x] Post model (Fields: `title`, `date`, `author`, `author_image`, `url`)
- [x] `sync_posts` command
- [x] GraphQL endpoint to fetch recent posts (recent 5 posts by date)  
- [x] Front end with responsive design
- [x] Tests for frontend and backend
- [x] Add post data source to nest bot


backend/apps/owasp/models/post.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class Post(BulkSaveModel, TimestampedModel):
+    """Post model."""
+
+    class Meta:
+        db_table = "owasp_post"
+        verbose_name_plural = "Posts"
+
+    author = models.CharField(verbose_name="Author name")
```
arkid15r:
```suggestion
    author_name = models.CharField(verbose_name="Author name")
```

backend/apps/owasp/models/post.py
```diff
@@ -0,0 +1,33 @@
...
+from django.db import models
+
+from apps.common.models import BulkSaveModel, TimestampedModel
+
+
+class Post(BulkSaveModel, TimestampedModel):
+    """Post model."""
+
+    class Meta:
+        db_table = "owasp_post"
```
arkid15r:
```suggestion
        db_table = "owasp_posts"
```

backend/apps/owasp/models/post.py
```diff
@@ -0,0 +1,33 @@
...
+        verbose_name_plural = "Posts"
+
+    author = models.CharField(verbose_name="Author name")
+    author_image = models.URLField(verbose_name="Author image URL", blank=True)
```
arkid15r:
```suggestion
    author_image_url = models.URLField(verbose_name="Author image URL", blank=True)
```

backend/apps/owasp/models/post.py
```diff
@@ -0,0 +1,33 @@
...
+    date = models.DateTimeField(verbose_name="Publication date")
```
arkid15r:
```suggestion
    published_at = models.DateTimeField(verbose_name="Publication date")
```

backend/apps/owasp/models/post.py
```diff
@@ -0,0 +1,33 @@
...
+    title = models.CharField(verbose_name="Title")
+    url = models.URLField(verbose_name="URL", unique=True)
+
+    def __str__(self):
+        """Return human-readable representation."""
+        return f"{self.title}"
```
arkid15r:
```suggestion
        return self.title
```

backend/apps/owasp/models/post.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @staticmethod
+    def bulk_save(posts, fields=None):
+        """Bulk save posts."""
+        BulkSaveModel.bulk_save(Post, posts, fields=fields)
+
+    @staticmethod
```
arkid15r, line 33: This normally goes into model managers. At least we don't want the hard limit here.

backend/apps/owasp/management/commands/owasp_sync_posts.py
```diff
@@ -0,0 +1,41 @@
...
+import yaml
+from django.core.management.base import BaseCommand
+
+from apps.github.utils import get_repository_file_content
+from apps.owasp.models.post import Post
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        posts_api_url = "https://api.github.com/repos/OWASP/owasp.github.io/contents/_posts"
```
arkid15r: Please use pygithub for GH API interactions when possible. 

Also use `from apps.github.utils import get_repository_file_content`

backend/apps/owasp/management/commands/owasp_sync_posts.py
```diff
@@ -0,0 +1,41 @@
...
+        response = requests.get(posts_api_url, timeout=30)
+        files = response.json()
+        posts = []
+
+        for file in files:
+            if file.get("name", "").endswith(".md"):
+                download_url = file.get("download_url")
+                file_content = get_repository_file_content(download_url)
+
+                if file_content.startswith("---"):
```
arkid15r: Could you unify this with [this](https://github.com/OWASP/Nest/blob/be57118ce74d6c1876521384ead456b7a7d3774f/backend/apps/owasp/models/common.py#L96) case?

Dishant1804: Sure i'll update it

backend/apps/owasp/models/post.py
```diff
@@ -0,0 +1,77 @@
...
+
+from datetime import datetime, timezone
+
+from django.db import models
+from django.utils.dateparse import parse_datetime
+
+from apps.common.models import BulkSaveModel, TimestampedModel
+
+
+class Post(BulkSaveModel, TimestampedModel):
```
arkid15r, line 11: My only concern here is that we don't have a unique key field for a post instance.

arkid15r (changes requested):
This code works well @Dishant1804 -- nicely done 👍 
I updated it with some changes, could you look into fixing the test?

Please keep events/posts separated (I want to try a different layout first). For posts please use 2 column layout with up to 6 posts shown.

frontend/src/pages/Home.tsx
```diff
@@ -131,43 +131,81 @@ export default function Home() {
...
+                      className="hover:underline"
+                      target="_blank"
+                      rel="noopener noreferrer"
+                    >
+                      {post.title}
+                    </a>
+                  </h3>
+                  <h4 className="flex flex-row items-center gap-1 text-sm">
+                    <img
+                      src={post.authorImageUrl}
```
arkid15r: Also we either need to figure out how to crop the user avatar images properly or just don't show them at all. I'm fine with just names
Screenshot 2025-03-12 at 2 13 48 PM

backend/apps/owasp/models/post.py
Dishant1804: > My only concern here is that we don't have a unique key field for a post instance.

Sure we can keep the url field as unique in that way we can achieve the expected results

frontend/src/pages/Home.tsx
Dishant1804: > Also we either need to figure out how to crop the user avatar images properly or just don't show them at all. I'm fine with just names
> Screenshot 2025-03-12 at 2 13 48 PM
> 

I think the names would be fine as the image url is not having query params (i tried that) to modify the size on the go, so lets keep only names that will also look good

Dishant1804:
https://github.com/user-attachments/assets/07a2d5ed-230f-42c8-84a8-5037d17653c6

arkid15r (approved):
Thanks for implementing this, let's try it!

PR #1046, opened 2025-03-07
Improve algolia.py input data validation

### LINKED ISSUE #1036: Improve algolia.py input data validation
Currently we don't handle unexpected input for some index search params like `page`.

Improve input data validation to handle possible SQL injection attempts like

Image

- [ ] Yes
- [x] No

ahmedxgouda:
Resolves #1036

- Created validators.py to contain all the validations.
- Created a validator function for each parameter.
- Used Django `validate_slug` for index_name and query.
- Used exception handling and some other if statements for the tests.
- Returned a `Bad Request` with a status code of `400` if the validation fails.


ahmedxgouda:
Sorry, I merged the upstream by mistake and was trying to revert this commit 😮‍💨

ahmedxgouda:
I think it is ready for reviewing @arkid15r

ahmedxgouda:
These are a lot of merge commits I made. Please review @arkid15r. I put more effort on resolving conflicts than solving the issue.

backend/apps/core/api/algolia.py
```diff
@@ -52,11 +53,17 @@ def algolia_search(request):
...
+        page = data.get("page", 1)
         ip_address = get_user_ip_address(request)
-        limit = int(data.get("hitsPerPage", 25))
-        page = int(data.get("page", 1))
         query = data.get("query", "")
 
+        validation_error = validate_search_params(data)
+
+        if validation_error:
+            return validation_error
```
arkid15r: Use walrus operator for cases like this.

backend/apps/core/validators.py
```diff
@@ -0,0 +1,95 @@
...
+
+from django.core.exceptions import ValidationError
+from django.core.validators import validate_slug
+from django.http import JsonResponse
+
+
+def validate_string_slug(value):
+    """Validate a string value as a slug."""
+    if not value or not isinstance(value, str):
+        return lambda field_name: JsonResponse(
```
arkid15r: I don't think this module should operate with JsonResponse. Just return an error message or None upon validation and process it on the upper level.

backend/apps/core/api/algolia.py
```diff
@@ -52,11 +53,17 @@ def algolia_search(request):
...
         index_name = data.get("indexName")
+
+        limit = data.get("hitsPerPage", 25)
+        page = data.get("page", 1)
         ip_address = get_user_ip_address(request)
-        limit = int(data.get("hitsPerPage", 25))
-        page = int(data.get("page", 1))
         query = data.get("query", "")
 
+        validation_error = validate_search_params(data)
```
arkid15r: The validation should occur as early as possible.

arkid15r (changes requested):
Thanks for improving this! Please look into following when you get a chance:

backend/apps/core/validators.py
```diff
@@ -0,0 +1,75 @@
...
+    if limit_validation_error := validate_limit(limit):
+        return limit_validation_error
+
+    if page_validation_error := validate_page(page):
+        return page_validation_error
+
+    if query_validation_error := validate_query(query):
+        return query_validation_error
+
+    return None
```
arkid15r: I think we should raise a custom exception for failed validation instead of returning an error.

arkid15r: Something like

```python
class ValidationError(Exception):
    """Exception raised for invalid search parameters."""
    def __init__(self, message):
        super().__init__(message)
        self.message = message


def validate_search_params(data):
    """Validate search parameters."""
    index_name = data.get("indexName")
    limit = data.get("hitsPerPage", 25)
    page = data.get("page", 1)
    query = data.get("query", "")

    validate_index_name(index_name)
    validate_limit(limit)
    validate_page(page)
    validate_query(query)


def validate_index_name(index_name):
    if not index_name:
        raise ValidationError("Index name is required.")


def validate_limit(limit):
    if not isinstance(limit, int) or limit <= 0:
        raise ValidationError("Limit must be a positive integer.")


def validate_page(page):
    if not isinstance(page, int) or page < 0:
        raise ValidationError("Page must be a non-negative integer.")


def validate_query(query):
    if not isinstance(query, str):
        raise ValidationError("Query must be a string.")

```

arkid15r: Consider `from django.core.exceptions import ValidationError`

backend/apps/core/validators.py
```diff
@@ -0,0 +1,82 @@
...
+        raise ValidationError(message) from None
+
+
+def validate_limit(limit):
+    """Validate limit."""
+    if not isinstance(limit, int):
+        message = "hitsPerPage must be an integer."
+        raise ValidationError(message)
+
+    max_limit = 100
```
arkid15r: We use 1000 in some cases.

backend/apps/core/api/algolia.py
```diff
@@ -50,19 +52,24 @@ def algolia_search(request):
...
-        page = int(data.get("page", 1))
         query = data.get("query", "")
 
         cache_key = f"{CACHE_PREFIX}:{index_name}:{query}:{page}:{limit}"
         if index_name == "chapters":
             cache_key = f"{cache_key}:{ip_address}"
 
-        result = cache.get(cache_key)
-        if result is not None:
+        if result := cache.get(cache_key):
```
arkid15r: I reverted this for more strict check we had before.

arkid15r (changes requested):
Looks good 👍 
The only step that left to get this merged is adding `validators.py` tests

backend/apps/core/api/algolia.py
ahmedxgouda: I did not review anything BTW. It is a glitch in GitHub pull requests extension in VS Code.

backend/tests/core/api/algolia_test.py
```diff
@@ -88,3 +88,78 @@ def test_algolia_search_invalid_method(self):
 
         assert response.status_code == requests.codes.method_not_allowed
         assert response_data["error"] == "Method GET is not allowed"
+
+    @pytest.mark.parametrize(
```
arkid15r, line 92: You should test validators functions -- create `tests/core/validators_test.py` and test each method with corresponding cases.

Keep some basic tests here too to make sure `algolia_search` is also covered

arkid15r (changes requested):
Good start, let's make it better:

backend/apps/core/api/algolia.py
ahmedxgouda: Why do you think it is more strict?

arkid15r (approved):
LGTM 👍

PR #1059, opened 2025-03-08
Fix tooltip conflicts for duplicate contributors

### LINKED ISSUE #1058: Fix project tooltip issue
> @arkid15r, this issues also persist for contributors section as well.
> 
> https://github.com/user-attachments/assets/26e44fc5-bb6b-4896-9d5d-3c76781a7f60 

 _Originally posted by @yashgoyal0110 in [#997](https://github.com/OWASP/Nest/issues/997#issuecomment-2708190947)_

rishyym0927:
Resolves #1058

This PR resolves an issue where tooltips would behave incorrectly if the same contributor appeared in multiple cards. The fix appends a unique identifier to the tooltip id, ensuring each tooltip instance is distinct.

1. Added a uniqueKey prop to ContributorAvatar.
2. Modified the tooltip id to include uniqueKey.
3. Updated parent component usage to pass a unique identifier.

https://github.com/user-attachments/assets/fb1ff195-e5b5-44c7-8d19-3194e2fce785

### DIFF
--- a/frontend/src/components/ContributorAvatar.tsx
+++ b/frontend/src/components/ContributorAvatar.tsx
@@ -5,7 +5,7 @@ import { Tooltip } from 'components/ui/tooltip'
 
 type ContributorProps = {
   contributor: TopContributorsTypeAlgolia | TopContributorsTypeGraphql
-  uniqueKey: string 
+  uniqueKey: string
 }
 
 const isAlgoliaContributor = (
--- a/frontend/src/components/ContributorAvatar.tsx
+++ b/frontend/src/components/ContributorAvatar.tsx
@@ -5,6 +5,7 @@ import { Tooltip } from 'components/ui/tooltip'
 
 type ContributorProps = {
   contributor: TopContributorsTypeAlgolia | TopContributorsTypeGraphql
+  uniqueKey: string 
 }
 
 const isAlgoliaContributor = (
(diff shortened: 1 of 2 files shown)


frontend/src/components/ContributorAvatar.tsx
```diff
@@ -39,7 +40,7 @@ const ContributorAvatar = memo(({ contributor }: ContributorProps) => {
 
   return (
     <Tooltip
-      id={`avatar-tooltip-${login}`}
+      id={`avatar-tooltip-${login}-${uniqueKey}`}
```
arkid15r, line 43: It seems this composite key should include the project part too. This way we avoid the same bug when same contributor is on the same position within multiple projects on the same page. This case is very low probability but still can happen. You can create a separate issue for that as it requires more effort (index update may be needed)/

arkid15r (approved):
LGTM

frontend/src/components/ContributorAvatar.tsx
rishyym0927: Yes I guess the probability is very less but still it can happen again. Though this implementation will work but as you we need update the indexes and create a more robust unique key. I was thinking of using Math.random() but again ig its not a good practise!!

kasya (approved):

PR #1064, opened 2025-03-09
🔍 Integrate Event Search from GraphQL 

### LINKED ISSUE #1024: Extend main page search bar autocomplete w/ events
After adding events we can now show them in the autocomplete dropdown too. Add events relevant to the user's search query similar to existing entities.

- [ ] Yes
- [x] No

**Additional context**

rishyym0927:
Resolves #1024

- Replaced Algolia event search with a direct GraphQL fetch using GET_MAIN_PAGE_DATA.
- Updated MultiSearchBar to dynamically fetch and display upcoming events alongside existing search results.
- Optimized the search logic to ensure smooth filtering and better performance.
- Ensured that events are properly formatted and clickable, directing users to their respective URLs.
- Events were not indexed in Algolia, so fetching them required a direct GraphQL query.
This improves accuracy by retrieving the latest event data directly from the database.

https://github.com/user-attachments/assets/af60080c-e9e8-4b30-a3bc-40205adb5287

### DIFF
--- a/frontend/src/components/MultiSearch.tsx
+++ b/frontend/src/components/MultiSearch.tsx
@@ -62,8 +62,8 @@ const MultiSearchBar: React.FC<MultiSearchBarProps> = ({
           if (filteredEvents.length > 0) {
             results.push({
               indexName: 'events',
-              hits: filteredEvents,
-              totalPages: 1, 
+              hits: filteredEvents.slice(0, suggestionCount),
+              totalPages: 1,
             })
           }
           setSuggestions(results.filter((result) => result.hits.length > 0))
--- a/frontend/src/components/MultiSearch.tsx
+++ b/frontend/src/components/MultiSearch.tsx
@@ -1,12 +1,15 @@
+import { useQuery } from '@apollo/client'
 import {
   faSearch,
   faTimes,
   faProjectDiagram,
   faBook,
   faUser,
+  faCalendarAlt,
 } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { fetchAlgoliaData } from 'api/fetchAlgoliaData'
+import { GET_MAIN_PAGE_DATA } from 'api/queries/homeQueries'
 import { debounce } from 'lodash'
 import type React from 'react'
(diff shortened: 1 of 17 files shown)


frontend/src/components/MultiSearch.tsx
```diff
@@ -1,12 +1,15 @@
...
   faSearch,
   faTimes,
   faProjectDiagram,
   faBook,
   faUser,
+  faCalendarAlt,
 } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { fetchAlgoliaData } from 'api/fetchAlgoliaData'
+import { GET_MAIN_PAGE_DATA } from 'api/queries/homeQueries'
```
arkid15r: This query fetches a lot of unrelated data. We already use it for the main page too so we can reuse it instead of sending another identical request.

arkid15r (changes requested):
This works but it's really inefficient from the performance perspective. Please improve:

frontend/src/components/MultiSearch.tsx
rishyym0927: Am I allowed to pass the data as props and then using them again?

arkid15r: I guess there should be a preferred way how to handle that in Reach. I'm not very strong on the FE side. Let's get more reviewers for better suggestions here.

rishyym0927: Yeah Sure !!!

arkid15r: @rishyym0927 I think you should go with the solution you suggested if there is no better way to do that. The props work for me 🤷‍♂️

rishyym0927: Yess work in progress

rishyym0927:
Hey @arkid15r, I made all the necessary changes to fix the performance issue !!

arkid15r (approved):
Thanks for addressing this!
Your implementation is the easiest way to make it work but is limited by frontend event title search. If you look at other entity types -- all them use backend/index search. However, we don't have event index so the alternative would be the DB search and GraphQL data exchange.

I'm going to merge it as it seems to satisfy the basic need for event search suggestion.

Issue #1029, opened 2025-03-06
Implement event metadata enrichment process


arkid15r:
**Describe the solution you'd like**
As we've recently added events we want to make the content look nice. For that we need an AI processing to populate event summary and location fields (need to be created). We have examples of doing exactly the same for project/issues -- the task it to add this logic for event. The approach must be the same both logically and structurally (code). This task involves Open AI API interaction as well as graphQL and frontend changes after the new fields are populated.

**Are you going to work on implementing this?**

- [ ] Yes
- [x] No

yashgoyal0110:
Can I work on it?

Dishant1804:
@arkid15r I would like to work on this issue

samyak003:
@arkid15r can I work on this ?

PR #1066, opened 2025-03-09
Added AI processing to events

samyak003:
Fixes: #1029


frontend/src/components/Modal.tsx
```diff
@@ -50,15 +51,17 @@ const Modal: React.FC<ModalProps> = ({
           {title}
         </DialogHeader>
         <Text className="mb-1 text-xs text-gray-700 dark:text-gray-300/60">
-          The issue summary and the recommended steps to address it have been generated by AI
+          The {entityType} summary and the recommended steps to address it have been generated by AI
```
arkid15r: This is unique entirely for different types of modal (issue, event). Pass the full text instead of entity

frontend/src/pages/Home.tsx
```diff
@@ -131,24 +133,31 @@ export default function Home() {
...
-                <div className="mr-4 flex items-center">
-                  <FontAwesomeIcon icon={faCalendar} className="mr-2 h-4 w-4" />
-                  <span>{formatDateRange(event.startDate, event.endDate)}</span>
+                  <h3 className="truncate">{event.name}</h3>
+                </button>
+                <div className="flex flex-wrap items-center text-sm text-gray-600 dark:text-gray-300">
+                  <div className="mr-4 flex items-center">
+                    <FontAwesomeIcon icon={faCalendar} className="mr-2 h-4 w-4" />
+                    <span>{formatDateRange(event.startDate, event.endDate)}</span>
+                  </div>
```
arkid15r: We want to show the suggested location here too.

frontend/src/types/event.ts
```diff
@@ -5,4 +5,5 @@ export type EventType = {
   name: string
   startDate: string
   url: string
+  summary?: string
```
arkid15r, line 8: It seems we missed suggestedLocation

arkid15r (changes requested):
Good work, please look into addressing the bot's and the following comments when you get a chance:

arkid15r (approved):
Let's try this!

PR #1098, opened 2025-03-13
Tooltip collision fix

### LINKED ISSUE #1065: Tooltip Collision for Contributors Across Multiple Projects
**Describe the bug**

**To Reproduce**
The previous fix  ensured unique tooltips when the same contributor appeared in multiple cards. However, this solution does not fully prevent tooltip collisions when the same contributor holds the same position across different projects on the same page. Issue #1059 

Display multiple projects on the same page.
Ensure the same contributor appears in the same position in different projects.
Observe that tooltips may still conflict.

**Expected behavior**

- [x] Yes
- [ ] No

Extend the unique key to include the project identifier.
This may require an index update or changes to how contributor tooltips are assigned.

**Additional Notes**
Since this requires additional effort, this issue is separate from the previous fix.

rishyym0927:
Resolves #1065 

The previous implementation was working but in certain cases it has the possibility of having a collision. Thus added one more variable that is the repo info to make it prone to the collisions!


frontend/src/components/ContributorAvatar.tsx
kasya, line 1: @rishyym0927 thanks for updating this. I'm checking it locally and for all the cases that I saw `repositoryInfo` is actually an empty string. 
So I'm thinking what if we pass in something else to this component from the `Card` to differentiate the project? Could be something connected to the project or to the card itself, like index for example 🤔

kasya (changes requested):

arkid15r (changes requested):
As @kasya mentioned the project related uniqueness is not always guaranteed. It seems the [code](https://github.com/OWASP/Nest/pull/1098/files#diff-36bcdc876b51922fce3da74efdb82197c601fd1b4d68f146843817473a3e172fR36) tries to cover multiple cases and the data is not properly populated for Algolia case.

rishyym0927:
Hey Im looking for more better and viable solution for this

rishyym0927:
Hey @arkid15r , @kasya 

I updated the tooltip id generation to ensure uniqueness when the same contributor appears in multiple projects. Previously, it relied on repositoryInfo, which was often empty, leading to potential duplicate IDs.

- Included projectName in the tooltip id to differentiate contributors across projects.
- If projectName is missing, it falls back to 'unknown-project' to avoid empty values.
- Ensured correct extraction of projectName for GraphQL contributors.

This should prevent conflicts where contributors in the same position across projects could end up with duplicate tooltip IDs.

frontend/src/components/ContributorAvatar.tsx
```diff
@@ -38,9 +38,11 @@ const ContributorAvatar = memo(({ contributor, uniqueKey }: ContributorProps) =>
       ? ` in ${(contributor as TopContributorsTypeGraphql).projectName}`
       : ''
 
+  const projectName = !isAlgolia ? (contributor as TopContributorsTypeGraphql).projectName : null
+
   return (
     <Tooltip
-      id={`avatar-tooltip-${login}-${uniqueKey}`}
+      id={`avatar-tooltip-${login}-${uniqueKey}-${projectName || 'unknown-project'}`}
```
arkid15r, line 45: But this always falls back to `unknown-project` as Algolia response never contains `projectName`.

arkid15r (changes requested):
I suggest look into our backend internals to understand how to fix this properly on the frontend.

arkid15r:
@rishyym0927 any updates on this? Are you going to complete it or we can close it?

rishyym0927:
> @rishyym0927 any updates on this? Are you going to complete it or we can close it?

Yes im working on it!!

Issue #1027, opened 2025-03-06
Clickalbe languages/topics


arkid15r:
**Describe the solution you'd like**
Let's improve the UX and make languages/topics clickable. The click event should redirect user to the search page with the relevant query. Make sure it works for all pages where languages/topics are shown.

The project details page example:
<img width="1199" alt="Image" src="https://github.com/user-attachments/assets/ceb1f6af-b6bd-4d60-80d3-19934f617a8c" />

**Are you going to work on implementing this?**

- [ ] Yes
- [x] No

**Additional context**

YourBroCode:
@arkid15r plz assign this issue to me

yashgoyal0110:
i would like to work on it

KaranNegi20Feb:
Hey @arkid15r I recently merged a frontend issue along with the tests, so maybe i can do this one along with writing some tests for this.

arkid15r:
Resolved in https://github.com/OWASP/Nest/pull/1099

PR #1099, opened 2025-03-13
Fixes issue #1027 Clickalbe tags for languages/topics

KaranNegi20Feb:
Solves issue: Clickalbe languages/topics #1027
Result:

https://github.com/user-attachments/assets/a51beaa1-975c-4bd9-be45-e67f58a415d9


KaranNegi20Feb:
Hey @arkid15r Thank you for assigning me the issue, please review the PR. Thank you.

frontend/src/components/ToggleableList.tsx
```diff
@@ -15,18 +16,23 @@ const ToggleableList = ({
   const [showAll, setShowAll] = useState(false)
 
   const toggleShowAll = () => setShowAll(!showAll)
+  const handleButtonClick = ({ item }) => {
+    navigate(`/projects?q=${encodeURIComponent(item)}`)
+  }
+  const navigate = useNavigate()
```
aramattamara: Please move the navigate definition above because using it before it's initialized can cause errors.

aramattamara (changes requested):
Thanks for this PR! Small changes are needed, but everything else looks good.

frontend/src/components/ToggleableList.tsx
KaranNegi20Feb: Sure @aramattamara

aramattamara (approved):
Checked locally @kasya can be merged.

frontend/src/components/ToggleableList.tsx
KaranNegi20Feb: Hey @aramattamara you may now proceed with the review, thank you.

KaranNegi20Feb:
> Checked locally @kasya can be merged.

Hey @arkid15r could you look into this, Thank you.

kasya (changes requested):
@KaranNegi20Feb Thanks for working on this!
The tags are now clickable and work as expected. I noticed one issue though - when you try to navigate back within the browser it requires at least three clicks before you're actually redirected to the prev. page. If you unfold that one you'll notice there are 2 identical extra items in history:
Screenshot 2025-03-14 at 6 03 08 PM

Can we fix that? It might be pretty frustrating for the end user 🤔

KaranNegi20Feb:
> @KaranNegi20Feb Thanks for working on this! The tags are now clickable and work as expected. I noticed one issue though - when you try to navigate back within the browser it requires at least three clicks before you're actually redirected to the prev. page. If you unfold that one you'll notice there are 2 identical extra items in history: Screenshot 2025-03-14 at 6 03 08 PM
> 
> Can we fix that? It might be pretty frustrating for the end user 🤔

Hey @kate, this issue persists from the https://nest.owasp.org/projects?q=shell, u may try this one and check it it

KaranNegi20Feb:
> > @KaranNegi20Feb Thanks for working on this! The tags are now clickable and work as expected. I noticed one issue though - when you try to navigate back within the browser it requires at least three clicks before you're actually redirected to the prev. page. If you unfold that one you'll notice there are 2 identical extra items in history: Screenshot 2025-03-14 at 6 03 08 PM
> > Can we fix that? It might be pretty frustrating for the end user 🤔
> 
> Hey @kate, this issue persists from the https://nest.owasp.org/projects?q=shell, u may try this one and check it it

Even if i send the previous history as a prop so that it can navigate back, it still wont bcoz of the issue from the search bar in projects, maybe we should create some diffrent issue for that,
You can also see the same behaviour here http://localhost:3000/community/users

KaranNegi20Feb:
ANOTHER SUGGESTION:
maybe instead of self we can open a new tab for these tags when clicked.

kasya (approved):
@KaranNegi20Feb Thanks for pointing that out! Let's merge this one in then and address that issue separately!

PR #1111, opened 2025-03-14
fixed cluster marker text color

### LINKED ISSUE #1091: Improve chapter map look
The white numbers are not that visible in the dark theme. I think we can do better than this:

- [ ] Yes
- [x] No

**Additional context**

Image

nipunh:
Resolves #1091 

Fixed dark color for cluster marker text. I have used #4b5563 color as it was default dark color for the map cluster group text. If required we can also use #000 (black)

Screenshot 2025-03-14 at 6 36 44 PM
Screenshot 2025-03-14 at 6 45 30 PM


arkid15r (approved):
LGTM 👍

PR #1112, opened 2025-03-15
Feature: scroll to top button

### LINKED ISSUE #614: Feature: scroll to top button
Sometimes, when users scroll down long pages, it becomes tedious to manually scroll back to the top. This issue can negatively impact the user experience, especially on mobile devices or pages with significant content.

Implement a "Scroll to Top" button that becomes visible when the user scrolls down a certain distance. Clicking this button would smoothly scroll the page back to the top. The button should be styled minimally and placed in the bottom-right corner of the viewport for easy access.

1.Adding a "Back to Top" link in the footer, though it requires users to scroll all the way down.
2. Implementing infinite scrolling to reduce the need to scroll back to the top, which may not be suitable for all use cases.

**Additional context**
- The button should be responsive and not obstruct other page elements.
- It can include animations (e.g., fade-in/fade-out on visibility).

Li-ChangC:
Resolves #614

This PR adds a responsive and animated "Scroll to Top" button. The button smoothly scrolls the user back to the top of the page, improving navigation on longer pages.

### Features ###
- Becomes visible after scrolling down a certain distance
- Positioned in the bottom-right corner for easy access
- Smooth scroll animation for better user experience
- Responsive and compatible with dark mode
- Does not obstruct other page elements

https://github.com/user-attachments/assets/dd9a1ac4-d087-46e9-b11f-06790cbd43e3


frontend/src/pages/Chapters.tsx
arkid15r, line 1: I suggest adding it at a single place instead of making it a part of each page. We don't want to maintain additions of the scroll up feature for future pages.

frontend/src/components/ScrollToTop.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    window.addEventListener('scroll', throttledScroll)
+    return () => window.removeEventListener('scroll', throttledScroll)
+  }, [handleScroll])
+
+  return (
+    <button
+      onClick={scrollToTop}
+      aria-label="Scroll to top"
+      className={`duration-400 fixed bottom-4 right-4 z-50 flex h-11 w-11 items-center justify-center rounded-full bg-owasp-blue bg-opacity-70 text-white shadow-lg transition-all hover:scale-105 hover:bg-opacity-100 active:scale-100 dark:bg-opacity-30 dark:text-slate-300 hover:dark:bg-opacity-50 ${
```
arkid15r: Could you keep it within acceptable max width?

arkid15r (changes requested):
Thanks for addressing this, I checked the new functionality and it works locally for me. Please look into the following when you get a chance:

frontend/src/pages/Chapters.tsx
Li-ChangC: Thanks for the suggestion, fixed now.

frontend/src/components/ScrollToTop.tsx
Li-ChangC: Thanks for the suggestion, fixed now.

arkid15r (commented):
It works for me locally 👍 

Could you add some tests for the new functionality?

kasya (approved):
LGTM! Thanks for adding this!

PR #1123, opened 2025-03-16
Add support for getting public IP address in local environment

kasya:
Added support for getting IP address for local environment. This helps working with components that require checking user IP address (like chapter's map, for example).


arkid15r (approved):
LGTM

PR #1196, opened 2025-03-28
fixed the modal window where the close button was focused earlier

### LINKED ISSUE #1193: Fix modal window
When it opens it has a focus on the close button. I don't think it's something we want.
Make sure the close button is not in focus.

- [ ] Yes
- [x] No

**Additional context**

Image

codic-yeeshu:
Resolves #1193

- Previously, when opening the modal, the close button was automatically focused. After this fix, it will only be highlighted when hovered. Here is a recording clip for reference.
- 

https://github.com/user-attachments/assets/bf35949c-d8cc-49e4-be34-2ade00046d71

### DIFF
--- a/frontend/src/components/Modal.tsx
+++ b/frontend/src/components/Modal.tsx
@@ -41,7 +41,7 @@ const Modal: React.FC<ModalProps> = ({
       >
         <DialogCloseTrigger
           asChild={false}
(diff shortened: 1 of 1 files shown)


arkid15r (approved):
LGTM 👍

Issue #1203, opened 2025-03-29
Improve main page recent chapters/projects component


arkid15r:
**Describe the solution you'd like**
Improve the look in general:
 - Make sure chapter/project leaders now use the recently introduces [LeadersList](https://github.com/OWASP/Nest/blob/main/frontend/src/components/LeadersList.tsx) component
 - Wrap names with [TruncatedText](https://github.com/OWASP/Nest/blob/main/frontend/src/components/TruncatedText.tsx) to have a consistent look

**Are you going to work on implementing this?**

- [ ] Yes
- [x] No

**Additional context**

<img width="1315" alt="Image" src="https://github.com/user-attachments/assets/fdb28dbb-9b73-4c6b-9f4c-067e6416192c" />

codic-yeeshu:
Could you please assign it to me?

MohammadIqbal0:
I want to resolve this, can you assign this to me?

arkid15r:
> I want to resolve this, can you assign this to me?

Please start from `good first issue` labeled tasks.

yashgoyal0110:
Can i work on it?

PR #1207, opened 2025-03-29
use leaders-list for main page

yashgoyal0110:
fixes: #1203


arkid15r (changes requested):
We want to have a consistent looks for recent chapters/projects on the main page.
For achieving that the contents need to be truncated. It's not the case at the moment:

Screenshot 2025-03-30 at 7 14 48 PM

Also make sure to remove extra spacing near commas. For future I suggest focusing on PRs quality in general.

yashgoyal0110:
> We want to have a consistent looks for recent chapters/projects on the main page. For achieving that the contents need to be truncated. It's not the case at the moment:
> 
> Screenshot 2025-03-30 at 7 14 48 PM
> Also make sure to remove extra spacing near commas. For future I suggest focusing on PRs quality in general.

Do we need to truncate every leader name or truncation is only needed when leader name div reaches available width?

yashgoyal0110:
Hey @arkid15r 
I have made the changes and made truncate component more flexible to to receive both elements and text props for wider usecase

arkid15r (approved):
LGTM

Issue #342, opened 2025-01-05
Implement NestBot block message template system


arkid15r:
**Is your feature request related to a problem? Please describe.**
Currently NestBot commands and event handlers use `markdown` method for rendering user facing content. The formatting process is time consuming and somewhat cumbersome.

**Describe the solution you'd like**
The idea is to implement a some sort of template system that would allow writing the content in text files and convert it into properly formatted markdown before sending to a user.
It should recognize formatting for `\n`, `<@user_id>`, `<#{OWASP_GSOC_CHANNEL_ID}>` and similar values.

bhagyashree980:
please this issue assign to me

yashpandey06:
My Approach:

1. Template Directory Setup:

Create a templates/ directory containing template files with pre-written message content, including dynamic placeholders (e.g., {{ user_id }}, {{ channel_id }}).



2. Template Engine:

Use a template engine like Jinja2 to render dynamic content. (Alternative engines like Mako or Chameleon can also be considered, but Jinja2 is widely used and well-supported.)



3. Rendering Process:

The template engine will process the templates, replacing placeholders with dynamic data, and output properly formatted content ready to be sent to Slack.

yashpandey06:
Was working on it but got a little occupied with some work .
Apologies for theates intimation 
Please re assign it @arkid15r

nitinawari:
Can i work on this

nitinawari:
@arkid15r  we have  this issue in to-do ?

arkid15r:
> [@arkid15r](https://github.com/arkid15r) we have this issue in to-do ?

Put it out from backlog for you, thank you @nitinawari

Prateekshit73:
Hi @arkid15r, I’d like to work on this! My approach:
- Create a `templates/` directory for text-based templates (e.g., `welcome.txt` with placeholders like `{{user_id}}`).
- Use Jinja2 to render dynamic content into markdown, supporting `\n`, `<@user_id>`, `<#channel_id>`, etc.
- Integrate into NestBot’s message handling with minimal changes.

Could you assign this to me? Thanks!

Li-ChangC:
Hi @arkid15r, I’d love to take on this issue. Could you please assign me this task? Thanks!

arkid15r:
[...]

PR #1211, opened 2025-03-30
Add unit tests for Slack command handlers

Li-ChangC:
This PR partially addresses #342 by adding unit tests for Slack command handlers (such as /jobs, /news, and /sponsor). These tests improve our coverage and help catch potential issues early. Further work on #342  will follow in subsequent PRs.


arkid15r (approved):
Thanks for adding these tests @Li-ChangC 

While they are quite useful my [suggestion](https://github.com/OWASP/Nest/issues/342#issuecomment-2734951686) was mostly about testing the entire produced output for some commands in order to make sure that the future refactored code produces the same output.

I'm merging this 👍

PR #1224, opened 2025-03-30
 Tooltip Position and Enhance Hover Effect for Contributor Avatars"

### LINKED ISSUE #1223: Fix Tooltip Position and Enhance Hover Effect for Contributor Avatars
Image

The tooltip is currently appearing at the top, which causes it to overlap with other elements.  I want the tooltip to:

Appear below the contributor’s icon instead of above.
Not collide with other contributor icons or text.

Adding a hover effect, where:

The tooltip appears when hovering.
The icon slightly enlarges or moves out for a better visual effect.

srinjoy933:
Resolves #1223


kasya (approved):
LGTM!

arkid15r (approved):

PR #1281, opened 2025-04-03
Added endpoint to generate csrf tokens

abhayymishraa:

- [x] Added endpoint ```/csrf``` for the genration of csrf 
- [x] Added utility 
- [x] Fixed testcase  

Add the PR description here.


abhayymishraa:
@coderabbitai  could you review this pr for any improvements and vulnurability! Thank you

abhayymishraa:
i think 
2. Token Expiry: Consider adding an expiration time to the CSRF token cookie to limit the window of possible attacks:
```
// Add max-age or expires attribute
const expiryMinutes = 60; // 1 hour
document.cookie = `csrftoken=${data.csrftoken}; path=/; SameSite=Lax; max-age=${expiryMinutes * 60}`
```

We can do that but if we not specify expiry minutes then it will be treated as session key by default. So it depends on weather we want to specify expiry time or use it as a default session key

arkid15r (approved):
LGTM 👍

PR #1338, opened 2025-04-10
Added sentry config and global-error component

Rajgupta36:
Task 
- [x]  Added global-error component
- [x]  Added sentry config for nextjs
- [x]  Added related test case

Media
Screenshot from 2025-04-11 00-33-40


arkid15r (approved):
Okay, let's try it. I'm pretty sure we'll need to improve the config later.

PR #1414, opened 2025-04-23
Added Metadata for dynamic pages and static pages

abhayymishraa:

Add the PR description here.

image


abhayymishraa:
@coderabbitai re review please

frontend/src/server/fetchCsrfTokenServer.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (!response.ok) {
+      throw new Error(`Failed to fetch CSRF token: ${response.status} ${response.statusText}`)
+    }
+
+    const data = await response.json()
+    return data.csrftoken
+  } catch (error) {
+    if (error) {
+      throw error
+    }
```
arkid15r: Could you explain this logic?

frontend/src/app/organizations/[organizationKey]/layout.tsx
```diff
@@ -0,0 +1,36 @@
+import { Metadata } from 'next'
+import React from 'react'
+import { GET_ORGANIZATION_DATA } from 'server/queries/organizationQueries'
```
Comment (arkid15r) [inline frontend/src/app/organizations/[organizationKey]/layout.tsx]: The existing queries are too heavy for layout purposes -- they return a lot of information that is not going to be used at all.

frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/layout.tsx
```diff
@@ -0,0 +1,38 @@
+import { Metadata } from 'next'
+import React from 'react'
+import { GET_REPOSITORY_DATA } from 'server/queries/repositoryQueries'
```
Comment (arkid15r) [inline frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/layout.tsx]: Same here -- you only need name/description but request all this [data](https://github.com/OWASP/Nest/blob/17304a8b8be46f7d5b6db1d480f72e241c11128a/frontend/src/server/queries/repositoryQueries.ts#L3)

arkid15r (changes requested):
It's a working PoC from my point of view. I have a couple of questions though:
 - I didn't get the whole point of the new `_DOCKER` env variables -- it all works after removing them (I might be missing something as I don't have much experience with Next.js/SSR)
 - Can we somehow test the new functionality?
 
 Please also look into optimizing the following:

abhayymishraa:
> It's a working PoC from my point of view. I have a couple of questions though:
> 
> * I didn't get the whole point of the new `_DOCKER` env variables -- it all works after removing them (I might be missing something as I don't have much experience with Next.js/SSR)
> * Can we somehow test the new functionality?
> 
> Please also look into optimizing the following:


for the first point :
When running the project locally (without Docker),
 the Next.js server (running on port 3000) sends API requests to http://localhost:8000/graphql because the backend is also running locally.
But when we run everything inside Docker containers,
 the Next.js app is now running inside its own container (next-frontend), and the backend is in another container (backend).
In Docker, containers can't use localhost to talk to each other.
 Instead, they must use each other’s container names to connect.
So inside Docker, the Next.js app should send requests to http://backend:8000/graphql (using the backend container's name) instead of localhost.
If we incorrectly use localhost inside a container, it will just point back to itself (the Next.js app), not the backend, and cause problems like missing metadata when using server-side rendering (SSR).
it all worked for you because we used try catch to catch it all but in a docker setup whatever i explained above should be followed to get metadata, you must checked it on the static routes try dynamic routes and then check the metadata.

for the testing purpose i will look forward whatever i can do to test this new functionality.

frontend/src/app/organizations/[organizationKey]/layout.tsx
Comment (abhayymishraa) [reply frontend/src/app/organizations/[organizationKey]/layout.tsx]: Yes i will work on that and make separate query for the metadata

frontend/src/server/fetchCsrfTokenServer.ts
abhayymishraa: Actually what in order to access the csrf token on server inside docker setup we can't just send the request to localhost:3000 because inside docker it can contact other containers through container_name

frontend/src/utils/helpers/apolloClientServer.ts
arkid15r, line 1: Shouldn't this be under `server/` directory?

arkid15r (changes requested):
Okay, after reading your comments and taking another look into this the actual problem seem to be in a bit different dimension. It look lime Next.js server side URLs need to be set differently. We don't need another environment (`docker` is not in the same dimension w/ local/staging/production).

I updated code to what at least works locally for me and removed all try/catch blocks (this is really bad practice, think twice before using this approach again).

Now we need to fix e2e tests, I guess it's about CSRF but I'm not really sure.

frontend/src/utils/helpers/apolloClientServer.ts
abhayymishraa: yes, i think both fetchcsrftoken and apolloserverclient will be under server/ directory

abhayymishraa:
> Okay, after reading your comments and taking another look into this the actual problem seem to be in a bit different dimension. It look lime Next.js server side URLs need to be set differently. We don't need another environment (`docker` is not in the same dimension w/ local/staging/production).
> 
> I updated code to what at least works locally for me and removed all try/catch blocks (this is really bad practice, think twice before using this approach again).
> 
> Now we need to fix e2e tests, I guess it's about CSRF but I'm not really sure.

I completely agree that it’s better not to use try-catch if it can be avoided. However, at that point, I had two options: either add a new dependency to the project or use try-catch to keep things simple. Since my main goal was to quickly check if the metadata works correctly for dynamic components, I decided to use try-catch temporarily without adding extra dependencies. Now that you have suggested to not use i will keep that in my mind, I plan to add the playwright-ssr dependency and properly fix the setup.

Thank you :)

arkid15r (approved):
I believe we still have inconsistency on what should and shouldn't belong with `server` directory. Let's address it later as another OWASP Snapshot time coming soon and we want to have the metadata feature released before that.

Great work, thank you @abhayymishraa !

PR #1455, opened 2025-05-04
Fix top contributors

### LINKED ISSUE #1426: Improve About page top contributors section

On the [about page ](https://nest.owasp.org/about) we don't want to include project leaders into Top Contributors block. In order to do that we need to extend graphql top contributors [endpoint](https://github.com/OWASP/Nest/blob/bacb8084b0d82e99f7187ae9c2665bf4fb6330e0/backend/apps/github/graphql/queries/repository_contributor.py#L21) to accept `excluded_users` optional param. If supplied the endpoint returned result shouldn't contain the specified users. The usernames to exclude are listed [here](https://github.com/OWASP/Nest/blob/bacb8084b0d82e99f7187ae9c2665bf4fb6330e0/frontend/src/app/about/page.tsx#L27).

- [ ] Yes
- [x] No

shining-bluemoon-11:
### RESOLVE #1426 

- [x] Improve About page top contributors section

In frontend , applied the filtering logic on the about page and 
in graphql endpoint added [excluded_usernames] parameter that allows to fetch all contributors  by explicitly passing the [excluded_usernames]

Screenshot 2025-05-04 195528


shining-bluemoon-11:
@arkid15r  done 
i learned so many things by solving just this single issue , i know it is not that big , but i learned a lot while resolving this 
if any further improvement needed , do let me know : ) 
Thanks for the opportunity , looking forward for more ! 
Thankyou

arkid15r (changes requested):
The bot's comment you marked as resolved is indeed correct. Your backend new logic remains unused as you filter top contributors on the frontend. The idea is to make backend return the correct (filtered) contributor set when requested (using the new `excluded_usernames` param).

shining-bluemoon-11:
Ok , I will update it

arkid15r (approved):
LGTM

@shining-bluemoon-11 @Dishant1804 thanks for taking care of this 👍

PR #1494, opened 2025-05-17
Fix backend trivy scan

### LINKED ISSUE #1479: Fix backend trivy scan
Take care of [JWT token false positive](https://github.com/OWASP/Nest/actions/runs/14958908834/job/42018396017#step:2:376) (exclude from scanning or suppress) and remove the temporary `continue-on-error: true` workarounds

ahmedxgouda:
Resolves #1479


arkid15r (commented):
I think I fixed it in 63f8bd8c261302db775f78e987e164475fd53937

arkid15r (approved):
LGTM

PR #1495, opened 2025-05-18
Improve docker cache mounts

ahmedxgouda:
- Improved docker builds by adding cache mounts to apk, pip, and npm.
- Improved docker builds in CI/CD by using gha cache.
- Tried to add cache mounts in CI/CD by a workaround.


ahmedxgouda:
@coderabbitai Suggest a way to inject the cache came from restoring cache step to the docker builder

ahmedxgouda:
I should stop using the VS Code GUI for commiting

ahmedxgouda:
It was not that hard, just reverted to the old commit before GHA

arkid15r (changes requested):
@ahmedxgouda it looks very promising!

Could you take care of the [schema image](https://github.com/OWASP/Nest/blob/main/schema/docker/Dockerfile.test) too?

Please also keep `FORCE_COLOR` where possible.

arkid15r (approved):
Alright, let's try this live!

PR #1510, opened 2025-05-22
Refactor Slack app commands/events

arkid15r:
Slack app commands/events code refactoring.


backend/tests/apps/common/utils_test.py
```diff
@@ -14,6 +15,22 @@
...
+        ("text", "expected"),
+        [
+            ("", ""),
+            ("A", "a"),
+            ("A1B2C3", "a1_b2_c3"),
+            ("AppHomeOpened", "app_home_opened"),
+            ("Contribute", "contribute"),
+            ("Gsoc", "gsoc"),
+            ("SomeViewWithNumbers123", "some_view_with_numbers123"),
+            ("TeamJoin", "team_join"),
```
kasya, line 31: Let's also add a test case for when the text is already in a snake case

backend/apps/slack/events/event.py
```diff
@@ -69,48 +75,36 @@ def handler(self, event, client, ack):
...
-            event (dict): The Slack event payload.
-            client (WebClient): The Slack WebClient instance.
-
-        """
-        try:
-            conversation = self.open_conversation(client, event.get("user"))
-            client.chat_postMessage(
-                channel=conversation["channel"]["id"],
-                text=":warning: An error occurred processing your request.",
+        if ephemeral_message := self.ephemeral_message:
```
ahmedxgouda, line 101: I think this will be always false

backend/apps/slack/events/event.py
```diff
@@ -199,8 +193,16 @@ def get_template_file_name(self):
             str: The template file name in snake_case.
 
         """
-        file_name = "".join(
-            [f"_{c.lower()}" if c.isupper() else c for c in self.__class__.__name__]
-        ).lstrip("_")
+        return f"events/{convert_to_snake_case(self.__class__.__name__)}.jinja"
 
-        return f"events/{file_name}.jinja"
+    def get_user_id(self, event) -> str:
```
ahmedxgouda, line 198: Does it deserve to create a function for this? We just used it once here and its logic is really simple.

backend/apps/slack/events/team_join.py
```diff
@@ -65,24 +62,14 @@ def get_context(self, event):
...
-            logger.exception("Slack API error in team_join handler: %s", e.response["error"])
-            raise
-        except Exception:
-            logger.exception("Error in team_join handler for user: %s", user_id)
-            raise
+        Returns:
+            str: The user ID.
+
+        """
+        return event["user"]["id"]  # User object is returned (other events return just the ID)!
```
ahmedxgouda, line 75: Same as above, why didn't we use it in the file?

backend/apps/slack/events/event.py
arkid15r: This part works for contribute/gsoc member_joined_channel events.

backend/apps/slack/events/event.py
arkid15r: In order to consolidate the logic and avoid code duplication for team_join handler i added this simple method. As for that event the user_id is returned in a different way (see the comment).

backend/apps/slack/events/team_join.py
arkid15r: To re-use the event handler instead of repeating the code w/ this small difference.

ahmedxgouda: I mean, you implemented this method, but did not use it for now. At least, we should use it here.
image

backend/apps/slack/events/event.py
ahmedxgouda: Yeah, it is good when we use it in the future.

ahmedxgouda: I just see it from low-level perspective, defining a function and calling it is more costly for the CPU than just evaluating the code in it.

arkid15r: This is a web-app w/ 2 languages (both Python/JS are not the best from performance point of view) and many different frameworks. I'm not against performant code however I don't think we should optimize it on that level.

backend/apps/slack/events/team_join.py
arkid15r: That's a valid point, updated everywhere.

arkid15r:
> > LGTM
> 
> This is funny. Me approving your PR

Thanks for reviewing 👍

kasya (changes requested):
This looks solid 👍🏼 
One request ⬇️

backend/tests/apps/common/utils_test.py
arkid15r: Done.

kasya (approved):

PR #1513, opened 2025-05-23
message model and message fetching

### LINKED ISSUE #1475: Implement Slack Data Sync for Channel Content Gathering
The first goal is to implement the data sync logic that will gather and persist content from selected Slack channels via the Slack API.

Dishant1804:
Resolves #1475 

- [x] new message model
- [x] command to sync slack messages across the conversations of workspace


backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -0,0 +1,215 @@
...
+                    oldest=oldest,
+                )
+                self._handle_slack_response(response, "conversations_history")
+                messages_data = response.get("messages", [])
+
+                messages = [
+                    message
+                    for message_data in messages_data
+                    if (
+                        message := self._create_message_from_data(
```
arkid15r: Similar to other data syncing this has to use `Message.update_data`

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -0,0 +1,215 @@
...
+                "slack_message_id": message_ts,
+                "conversation": conversation,
+                "text": message_data.get("text", ""),
+                "timestamp": message_ts,
+                "is_thread": is_actual_thread_parent,
+            }
+
+            if include_threads and is_actual_thread_parent:
+                try:
+                    thread_response = client.conversations_replies(
```
arkid15r: We should split this somehow, maybe sync replies after we're fully done w/ "main" conversations.

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -0,0 +1,215 @@
...
+            )
+
+        except SlackApiError as e:
+            self.stdout.write(
+                self.style.ERROR(
+                    f"Failed to fetch messages for {conversation.name}: {e.response['error']}"
+                )
+            )
+
+    def _create_message_from_data(
```
arkid15r, line 245: And this seems to belong with `Message.from_slack`

backend/apps/slack/models/message.py
```diff
@@ -0,0 +1,84 @@
...
+            message = Message(slack_message_id=slack_message_id, conversation=conversation)
+
+        thread_replies_list = data.get("thread_messages", [])
+        if thread_replies_list:
+            parent_text = data.get("text", "")
+            combined_text_parts = [parent_text]
+            combined_text_parts.extend(
+                msg.get("text", "") for msg in thread_replies_list if msg.get("text")
+            )
+            message.text = "\n\n".join(filter(None, combined_text_parts))
```
arkid15r: We must keep the messages atomic -- let's save thread replies as messages with a foreign key to a parent.

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -0,0 +1,215 @@
...
+                        combined_text_list.extend(reply_texts)
+
+                        message_dict["text"] = "\n\n".join(filter(None, combined_text_list))
+
+                except SlackApiError:
+                    logger.warning("Failed to fetch thread replies")
+
+            return Message.update_data(message_dict, save=False)
+
+        except KeyError:
```
arkid15r: Why do we need this?

backend/apps/slack/models/message.py
```diff
@@ -0,0 +1,84 @@
...
+
+class Message(TimestampedModel):
+    """Slack Message model."""
+
+    class Meta:
+        db_table = "slack_messages"
+        verbose_name_plural = "Messages"
+        unique_together = ("conversation", "slack_message_id")
+
+    conversation = models.ForeignKey(
```
arkid15r, line 28: Please also make sure we have `author` information stored (FK to `Member`).

arkid15r (changes requested):
It's a great working prototype, I tested it locally and it does the job. Please modify it based on the following suggestions:

backend/apps/slack/management/commands/slack_sync_messages.py
Dishant1804: https://api.slack.com/methods/conversations.replies I went through the docs while implementing the fetching and found out that the replies are just normal messages with same `thread_ts` , so syncing it along with the main conversation will retain the context of the conversation, which in turn will help in generating better responses
 Also splitting will introduce 2 passes (one for main messages and other for fetching all the thread replies)

arkid15r: Another reason why replies should be stored as regular messages too. I think we can keep it tied together, however `Message::update_data` method should consume fully fetched message data -- you'll probably have to save parent messages first.

Dishant1804: Sure I'll look into it :)

backend/apps/slack/models/message.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+    slack_message_id = models.CharField(verbose_name="Slack Message ID", max_length=50)
+    text = models.TextField(verbose_name="Message Text", blank=True)
+    timestamp = models.DateTimeField(verbose_name="Message Timestamp", blank=True)
+
+    def __str__(self):
+        """Human readable representation."""
+        text_display_limit = 50
+        text_preview = (
+            self.text[:text_display_limit] + "..."
```
arkid15r: Try `common::utils::truncate()`

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -0,0 +1,306 @@
...
+
+        if replies_to_save:
+            batch_size = 1000
+            for i in range(0, len(replies_to_save), batch_size):
+                batch = replies_to_save[i : i + batch_size]
+                Message.bulk_save(batch)
+
+    def _create_message_from_data(
+        self,
+        client: WebClient,
```
arkid15r, line 247: `client` seems to be unused here.

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -0,0 +1,306 @@
...
+
+            slack_user_id = message_data.get("user") or message_data.get("bot_id")
+            if not slack_user_id:
+                return None
+
+            try:
+                author = Member.objects.get(
+                    slack_user_id=slack_user_id, workspace=conversation.workspace
+                )
+            except Member.DoesNotExist:
```
arkid15r, line 275: Can we create a user instead?

arkid15r (changes requested):
Looks good, a couple more ideas to consider:

backend/apps/slack/management/commands/slack_sync_messages.py
Dishant1804: I think we are already syncing the members of the workspace -- so if the member is able to message he/she will de facto be a part of workspace and they will already be synced as we are syncing the members as well in `slack_sync_data` [here](https://github.com/OWASP/Nest/blob/085ee6218df12c71aebc55c8fc40462ec18012d4/backend/apps/slack/management/commands/slack_sync_data.py#L102)
Creating a user will be redundant here

arkid15r: Consider a case when user joins after the `slack_sync_data` is finished but before the messages sync. It's also important to remember that those 2 sync will not always go together.

Dishant1804: This is such a great case I missed it completely 
Thanks for explaining I'll implement it

backend/apps/slack/models/message.py
```diff
@@ -0,0 +1,114 @@
...
+    parent_message = models.ForeignKey(
+        "self",
+        on_delete=models.CASCADE,
+        related_name="thread_replies",
+        null=True,
+        blank=True,
+    )
+    slack_message_id = models.CharField(verbose_name="Slack Message ID", max_length=50)
+    text = models.TextField(verbose_name="Message Text", blank=True)
+    timestamp = models.DateTimeField(verbose_name="Message Timestamp", blank=True)
```
arkid15r:
```suggestion
    created_at = models.DateTimeField(verbose_name="Created at", blank=True)
```

backend/apps/slack/models/message.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __str__(self):
+        """Human readable representation."""
+        text_preview = truncate(self.text, 50)
+        return f"{text_preview}"
```
arkid15r:
```suggestion
        return truncate(self.text, 50)
```

backend/apps/slack/models/message.py
```diff
@@ -0,0 +1,114 @@
...
+        is_thread_reply: bool = False,
+        parent_message: "Message | None" = None,
+    ) -> None:
+        """Update instance based on Slack message data."""
+        self.slack_message_id = message_data.get("ts", "")
+        self.conversation = conversation
+        self.text = message_data.get("text", "")
+        self.author = author
+
+        self.is_thread_parent = not is_thread_reply and bool(
```
arkid15r: Do you need explicit `bool()` here?

backend/apps/slack/models/message.py
```diff
@@ -0,0 +1,114 @@
...
+            message_data.get("reply_count", 0) > 0
+        )
+        if is_thread_reply and parent_message:
+            self.parent_message = parent_message
+
+        if ts := message_data.get("ts"):
```
arkid15r: I'm not sure about constraints here -- the timestamp (create_at) is not nullable and I don't think a message can come w/o ts? Let me know if I'm missing something.

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -0,0 +1,312 @@
...
+            try:
+                latest_reply = (
+                    Message.objects.filter(
+                        conversation=conversation,
+                        parent_message=parent_message,
+                    )
+                    .order_by("-timestamp")
+                    .first()
+                )
+                oldest_ts = latest_reply.timestamp.timestamp() if latest_reply else None
```
arkid15r: This is why it should be named `created_at`

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -0,0 +1,312 @@
...
+        client: WebClient,
+        message_data: dict,
+        conversation: Conversation,
+        *,
+        is_thread_reply: bool,
+        parent_message: Message | None = None,
+    ) -> Message | None:
+        """Create Message instance using from_slack pattern."""
+        try:
+            if message_data.get("subtype") in ["channel_join", "channel_leave", "bot_message"]:
```
arkid15r:
```suggestion
            if message_data.get("subtype") in {"channel_join", "channel_leave", "bot_message"}:
```

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -0,0 +1,312 @@
...
+                batch = replies_to_save[i : i + batch_size]
+                Message.bulk_save(batch)
+
+    def _create_message_from_data(
+        self,
+        client: WebClient,
+        message_data: dict,
+        conversation: Conversation,
+        *,
+        is_thread_reply: bool,
```
arkid15r: The default value missing?

backend/apps/slack/models/message.py
Dishant1804: Yes you are correct the message will not come w/o timestamp -- it is expected from the response

backend/apps/slack/models/message.py
Dishant1804: No we dont need it, we can do without it -- I'll remove it

PR #1514, opened 2025-05-24
Improve roadmap visualization

### LINKED ISSUE #1361: Refactor the Roadmap Section on About Page to Enhance Visual Appeal and Structure
### **Is your feature request related to a problem? Please describe.**  
The current roadmap section on the About page looks fairly simple and lacks a structured or visually engaging layout. It doesn't effectively highlight the project’s milestones or planned features, which may reduce its impact and clarity for users or contributors.

### **Describe the solution you'd like**  
Redesign the roadmap with a more structured and visually appealing layout. This will make it more interactive and easier to understand at a glance.

upd: we now have milestones data available and can use OWASP Nest milestones for the roadmap component content -- for this we need to extend [resolve_recent_milestones](https://github.com/OWASP/Nest/blob/a5810974cd3d4ac9f46f4aa531a7faa5edea3ffd/backend/apps/github/graphql/queries/milestone.py#L15) method w/ filtering by project ability.

### **Are you going to work on implementing this?**  
- [x] Yes  
- [ ] No  

### **Additional context**  

Image

ahmedxgouda:
Resolves #1361

I preferred to add the milestone description in this case. I also removed a lot of non-necessary frontend code
image


frontend/src/app/about/page.tsx
```diff
@@ -159,20 +159,38 @@ const About = () => {
...
-                >
-                  {item.title}
-                </Link>
-              </li>
+          <div className="grid gap-4">
+            {projectMetadata.recentMilestones.map((milestone, index) => (
+              <div
+                key={index}
+                className={`flex items-center gap-4 overflow-hidden rounded-lg bg-gray-200 p-6 dark:bg-gray-700 ${
+                  index % 2 === 0 ? '' : 'ml-8'
```
kasya: Can we update the layout to always have the same alignment for all items? I understand the idea behind it, but for now I'm thinking maybe we could simplify this. 
Screenshot 2025-05-24 at 12 32 07 PM

frontend/src/app/about/page.tsx
```diff
@@ -159,20 +159,38 @@ const About = () => {
...
+                <div className="flex-1">
+                  <Link
+                    href={milestone.url}
+                    target="_blank"
+                    className="inline-block hover:underline"
+                  >
+                    <h3 className="mb-2 text-xl font-semibold text-blue-400">{milestone.title}</h3>
+                  </Link>
+                  <p className="text-gray-600 dark:text-gray-300">{milestone.body}</p>
+                  <div className="mt-2 flex items-center gap-4">
```
kasya: Also, can we remove progress block for now? I like the idea, but with how milestone completion is calculated - this might be confusing. It might decrease significantly after showing 95% for example. So I think it would be better to not show any specific completion numbers.

frontend/src/app/about/page.tsx
```diff
@@ -159,20 +159,38 @@ const About = () => {
...
+                  index % 2 === 0 ? '' : 'ml-8'
+                }`}
+              >
+                <div className="flex-1">
+                  <Link
+                    href={milestone.url}
+                    target="_blank"
+                    className="inline-block hover:underline"
+                  >
+                    <h3 className="mb-2 text-xl font-semibold text-blue-400">{milestone.title}</h3>
```
kasya: And instead of showing the progress bar we could add an icon here to any items that have a percentage (only if it's > 0) that will show that the milestone is in progress. 
I think something like `user-gear` [icon](https://fontawesome.com/icons/user-gear?f=classic&s=solid) should be good 👍🏼  Would be good to add a tooltip to it that will say something like `In progress` or `Started`

kasya (changes requested):
Love this! 
A couple of requests ⬇️ :

arkid15r (commented):
LGTM

arkid15r (approved):

PR #1515, opened 2025-05-25
improved breadcrumbs by introducing page layout component

### LINKED ISSUE #1425: Improve breadcrumbs component
We have pretty simple and straight-forward breadcrumbs component. However it lacks some flexibility due to its simplicity.

Improve the breadcrumbs component to display each project/chapter/user/etc page H1 header content, e.g. for OWASP Nest it should be `Home > Projects > OWASP Nest` instead of `Home > Project > Nest`. Another example -- https://nest.owasp.org/members/arkid15r should show `Home > Members > Arkadii Yakovets` instead of `Home > Members > Arkid15r`.

Implement changes for all pages and add a couple of simple tests.

- [ ] Yes
- [x] No

**Additional context**

Image

codic-yeeshu:
Resolves #1425 

- Improved breadcrumbs by introducing a Page Layout component.
- Instead of showing elements from the URL, we now use props as breadcrumb items.
- Also modified test cases that were affected by the changes.


codic-yeeshu:
Hey @arkid15r ! I've looked at the nitpick comments by the code bot, and I don't think those changes should be considered for this PR. However, if changes are required, please let me know.

frontend/src/components/BreadCrumbs.tsx
```diff
@@ -3,16 +3,20 @@
...
+import _ from 'lodash'
 import Link from 'next/link'
-import { usePathname } from 'next/navigation'
-import { capitalize } from 'utils/capitalize'
 
-export default function BreadCrumbs() {
-  const homeRoute = '/'
-  const pathname = usePathname()
-  const segments = pathname.split(homeRoute).filter(Boolean)
+interface BCrumbItem {
```
kasya: Can we come up with a more clear naming here? For example `BreadCrumbItem` - it does not seem to be conflicting with the imported component 🤔

frontend/src/app/chapters/[chapterKey]/page.tsx
```diff
@@ -60,15 +61,22 @@ export default function ChapterDetailsPage() {
...
-      geolocationData={chapter}
-      is_active={chapter.isActive}
-      socialLinks={chapter.relatedUrls}
-      summary={chapter.summary}
-      title={chapter.name}
-      topContributors={topContributors}
-      type="chapter"
-    />
+    <PageLayout
+      bcItems={[
```
Comment (kasya) [inline frontend/src/app/chapters/[chapterKey]/page.tsx]: Same here - can we update this for a clearer naming? If it's conflicting with anything we can do something like `crumbItem`.

frontend/src/components/PageLayout.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import BreadCrumbs from 'components/BreadCrumbs'
+
+interface BreadcrumbItem {
+  title: string
+  href: string
```
kasya: I think `path` is more suitable here than `href`

frontend/src/components/BreadCrumbs.tsx
kasya, line 1: Also noticed how different this component looks like in different themes.
Light theme has some weird background color.
Can we get rid of it?
Screenshot 2025-06-01 at 1 23 30 PM
Screenshot 2025-06-01 at 1 23 37 PM

frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx
```diff
@@ -106,20 +107,32 @@ const RepositoryDetailsPage = () => {
...
-      summary={repository.description}
-      title={repository.name}
-      topContributors={topContributors}
-      topics={repository.topics}
-      recentMilestones={repository.recentMilestones}
-      type="repository"
-    />
+    <PageLayout
+      bcItems={[
+        { title: 'Organizations', href: '/organizations' },
```
Comment (kasya) [inline frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx]: Can we utilize `usePathname`  from `next/navigation` to dynamically get the path, UpperCase it and use here for title and a path itself for `href`(should be named path)? Instead of hardcoding this.

kasya (changes requested):
Works good! 👍🏼 
Left some requests for naming, styling - and just general improvements.

codic-yeeshu:
Thank you for the feedback, @kasya . I’ll make the updates asap!.

frontend/src/components/BreadCrumbs.tsx
codic-yeeshu: updated

frontend/src/app/chapters/[chapterKey]/page.tsx
Comment (codic-yeeshu) [reply frontend/src/app/chapters/[chapterKey]/page.tsx]: updated

frontend/src/components/PageLayout.tsx
codic-yeeshu: updated

frontend/src/components/BreadCrumbs.tsx
codic-yeeshu: updated

frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx
Comment (codic-yeeshu) [reply frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx]: updated

codic-yeeshu:
Hey @kasya, @arkid15r 

SonarQube is blocking the merge due to repeated test logic.
Can we move the shared logic to a common utility file? I can push a commit for it, as I’ve already tested it locally.

Let me know your thoughts!

kasya (changes requested):
Left a suggestion!

kasya (approved):
Looked a bit more into this and realized it won't be possible to wrap it just once in the Root Layout. 

This works great right now so let's keep it as is 👍🏼

frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx
```diff
@@ -106,20 +107,29 @@ const RepositoryDetailsPage = () => {
...
-      topContributors={topContributors}
-      topics={repository.topics}
-      recentMilestones={repository.recentMilestones}
-      type="repository"
-    />
+    <PageLayout
+      breadcrumbItems={[
+        {
+          title: repository.name,
+          path: `/organizations/${organizationKey}/repositories/${repositoryKey}`,
```
Comment (kasya) [inline frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx]: Actually, I just realized this was never addressed. 
Can we use a path from `usePathname()` here instead? This could be a potential issue if we change the structure at some point or rename something.

frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx
```diff
@@ -106,20 +107,29 @@ const RepositoryDetailsPage = () => {
...
-      stats={RepositoryStats}
-      summary={repository.description}
-      title={repository.name}
-      topContributors={topContributors}
-      topics={repository.topics}
-      recentMilestones={repository.recentMilestones}
-      type="repository"
-    />
+    <PageLayout
+      breadcrumbItems={[
```
Comment (kasya) [inline frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx]: Another point: passing an Array with just one object is a bit unclear for readability. I understand you probably made it this way for easier merging them in `PageLayout` but could you update this to pass in an object instead?

Comment (codic-yeeshu) [reply frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx]: Totally valid. Just updating it.

frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx
Comment (codic-yeeshu) [reply frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx]: Yeah right, I can take this logic to PageLayout.

frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx
Comment (kasya) [inline frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx:1]: I noticed this when going over OWASP projects, but the issue is probably with how you build BreadCrumb items in `PageLayout`. 
The OWASP Amass org (this is just an example):
when on the organization page
Screenshot 2025-06-13 at 6 57 54 PM
But when we are getting to a repo level the org name gets messed up since you just parse the URL.
Screenshot 2025-06-13 at 6 52 53 PM

So to make this work we'll have to switch back to using Array of objects to pass in to `breadcrumbItems` and pass in the whole path starting from `organization` level, so 4 objects, each with path and title. We'll only need to do this for this page, since it's the only one that has two nested levels.
That will also require updating logic in `PageLayout` for generating Breadcrumb to accommodate an array of objects. 

For all other pages passing just one item is enough - the title and the rest will be parsed from the url (since they are all 1 level deep)

kasya (changes requested):
This is much better! I noticed one issue though for repositories pages. This is a unique use case but we'll need to update logic to make this work.

frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx
Comment (codic-yeeshu) [reply frontend/src/app/organizations/[organizationKey]/repositories/[repositoryKey]/page.tsx:1]: Understood @kasya, let me update that.

codic-yeeshu:
Hey @kasya ! I've created a cleaner [PR](https://github.com/OWASP/Nest/pull/1677) to address this issue. The previous one had several conflicts, so I’ve consolidated all the fixes and enhancements into this new PR for clarity.

Issue #1424, opened 2025-04-26
Implement Sponsor component


arkid15r:
**Describe the solution you'd like**
Currently we have this layout on the main page

<img width="1269" alt="Image" src="https://github.com/user-attachments/assets/cd493282-db5f-4296-adb9-82e6c0eb1076" />

**Describe alternatives you've considered**
The idea is to extract it to a component and reuse on the main page (keeping the look as is) and add it at the bottom of each project details, chapter details and repository details page. The component should be configurable for each case. The target link should be unique for each project/chapter and look like this: https://owasp.org/donate/?reponame=www-project-nest&title=OWASP+Nest where reponame is `www-project-` prefixed project key (`www-chapter-` for chapters).

**Are you going to work on implementing this?**

- [ ] Yes
- [x] No

yashgoyal0110:
@arkid15r 
Assign please

rishyym0927:
@arkid15r Can I work on this issue?

arkid15r:
@rishyym0927 any progress towards PR?

rishyym0927:
Hey @arkid15r sorry for the delay but Im working on it!!

M-ayank2005:
Hey @arkid15r,
I would like to work on this issue. Kindly assign it to me.
Thanks.

M-ayank2005:
Hi @arkid15r,
I've created a new pull request with the requested changes.Please have a look when you get a chance .Thanks

PR #1536, opened 2025-05-30
Add sponsor section component

M-ayank2005:
This PR adds a contextual sponsor section to the details pages for certain entity types.

Resolves: #1424

**Changes Made :**

- Updated DetailsCardProps interface: Added a new key property.
- Donation URL generation: Created a utility function to format the donation URL based on the entity type.
- Conditional rendering: The sponsor section is only shown for: Projects , Chapters and Repositories

Sponsor section UI:

- Includes explanatory text
- Features a donation button with contextual text based on the entity type

**Placement:** The sponsor section appears at the bottom of the relevant details pages.


frontend/src/components/CardDetailsPage.tsx
```diff
@@ -47,6 +49,24 @@ const DetailsCard = ({
...
+      repoPrefix = 'www-project-'
+    } else if (type === 'chapter') {
+      repoPrefix = 'www-chapter-'
+    } else if (type === 'repository') {
+      repoPrefix = ''
+    }
+
+    const encodedTitle = encodeURIComponent(`OWASP ${title || ''}`)
+    return `https://owasp.org/donate/?reponame=${repoPrefix}${key}&title=${encodedTitle}`
+  }
```
kasya: Using a map here would make more sense and and would be a cleaner approach:
```suggestion
  const getDonationUrl = () => {
    if (!key) return 'https://owasp.org/donate/'
  
    const prefixMap = {
      project: 'www-project-',
      chapter: 'www-chapter-',
      repository: '',
    }
  
    const encodedTitle = encodeURIComponent(`OWASP ${title || ''}`)
    return `https://owasp.org/donate/?reponame=${prefixMap[type] || ''}${key}&title=${encodedTitle}`
}
```

frontend/src/types/card.ts
```diff
@@ -40,6 +40,7 @@ export interface DetailsCardProps {
   geolocationData?: GeoLocDataGraphQL | null
   heatmap?: JSX.Element
   is_active?: boolean
+  key?: string
```
kasya, line 43: @M-ayank2005 You added this new property here, but I don't see any key being passed on any page 🤔 
Without it - links don't work as they should.

kasya (changes requested):
Thanks for working on this. You'll need to fix some issues:

kasya:
Could you also add tests to cover this new component added to project, chapter and repo pages?

frontend/src/components/CardDetailsPage.tsx
M-ayank2005: Yeah , using map would be better and cleaner approach.

frontend/src/components/CardDetailsPage.tsx
```diff
@@ -51,19 +51,16 @@ const DetailsCard = ({
 }: DetailsCardProps) => {
   const getDonationUrl = () => {
     if (!key) return 'https://owasp.org/donate/'
```
kasya, line 53: This is still not addressed. No key is present (ever) so all links only direct to `https://owasp.org/donate/`
You need to update the backend to pass in this data for a key.

kasya (changes requested):

frontend/src/components/CardDetailsPage.tsx
M-ayank2005: @kasya  Will it work as expected just  by passing key in DetailsCard props?

kasya: @M-ayank2005 depends on what you expect to have in that `key` 🤔 . But I assume it should.
The thing is - you need to actually work on passing it in in this same PR.

kasya:
@M-ayank2005 Hi! Any updates here?

Issue #711, opened 2025-02-02
OWASP Project Health Dashboard


arkid15r:
### **OWASP Project Health Dashboard Development Plan**

The **Project Health Dashboard** will provide internal visibility into the health and status of OWASP projects, including both new and older projects, with a focus on monitoring key health metrics and automated evaluations. This tool will help identify projects that require attention, such as those with low activity, leadership issues, or non-compliance, and will not be publicly available.

#### **1. Define Health Metrics and Indicators**
   - **Goal:** Define key metrics and indicators to assess the health of a project.
   - **Tasks:**
     - Identify key health metrics (e.g., number of contributors, open vs. closed issues, release frequency, etc.).
     - Develop project health indicators, such as:
       - **Releases**: Number of releases, stability, frequency.
       - **Integration with OWASP.org**: Regular updates to OWASP site with project content.
       - **Project age**: Age of the project, with considerations for mature projects.
       - **Last release**: Last commit, commit hash, and the stars received.
       - **Activity metrics**: Number of stars, forks, issues, pull requests, issues needing help.
       - **Compliance checks**: Ensuring projects adhere to OWASP funding policies.
     - Metrics for **Project Vitality**: Track growth over time in terms of contributors, issues, PRs, and activity.

#### **2. Identify and Flag Projects with Health Issues**
   - **Goal:** Automatically identify projects with potential health issues.
   - **Tasks:**
     - Flag **old projects** with no recent releases or updates.
     - Identify **new projects that failed to launch** based on lack of activity (e.g., no commits or releases in a set period).
     - Highlight projects with **fewer than two leaders** or contributors.
     - Track projects with **non-compliant funding policies**.
     - Perform automated checks to evaluate if a project’s **health metrics** fall below set thresholds.

#### **3. Create Project Health Evaluation Criteria**
   - **Goal:** Develop a system to evaluate the health of projects based on their activity and compliance.
   - **Tasks:**
     - Establish health thresholds for various metrics:
       - Number of releases within a defined timeframe.
       - Integration level with OWASP.org (ensure the project has up-to-date content and documentation).
       - Frequency and stability of releases (e.g., multiple releases in the last year).
       - Number of open issues versus closed issues, and the turnaround time for pull requests.
       - Contributor growth and activity trends.
     - Track **project age** to differentiate between young and mature projects.
     - Flag projects with **issues needing help** or long unaddressed open issues.

#### **4. Dashboard Layout & Design**
[...]

PR #1550, opened 2025-06-03
Create a health evaluation criteria

### LINKED ISSUE #1540: Create a health evaluation criteria
- Create a script to populate project health metrics fields.
- Create a script to populate the score of the project health.

ahmedxgouda:
Resolves #1540

- Implemented two scripts to update the metrics and the score of the projects health.
- Implemented the functionality in two scripts instead of one for simplicity and separation of concerns. Also, the requirements may change in the future, so no reason to update all fields of the metrics to calculate new score.
- Updated the models with new properties.
- Implemented the tests.
- Added targets to Makefile.

### DIFF
--- a/backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
+++ b/backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
@@ -11,6 +11,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         forward_fields = {
+            "age_days": 6.0,
             "contributors_count": 6.0,
             "forks_count": 6.0,
             "is_funding_requirements_compliant": 5.0,
@@ -22,7 +23,6 @@ class Command(BaseCommand):
             "total_releases_count": 6.0,
         }
         backward_fields = {
-            "age_days": 6.0,
             "last_commit_days": 6.0,
             "last_pull_request_days": 6.0,
             "last_release_days": 6.0,
--- a/backend/tests/apps/owasp/management/commands/owasp_update_project_health_metrics_scores_test.py
+++ b/backend/tests/apps/owasp/management/commands/owasp_update_project_health_metrics_scores_test.py
@@ -8,7 +8,7 @@ from apps.owasp.management.commands.owasp_update_project_health_metrics_scores i
 from apps.owasp.models.project_health_metrics import ProjectHealthMetrics
 from apps.owasp.models.project_health_requirements import ProjectHealthRequirements
 
-EXPECTED_SCORE = 58.0
+EXPECTED_SCORE = 52.0
 
 
 class TestUpdateProjectHealthMetricsScoreCommand:


backend/apps/owasp/management/commands/owasp_update_project_health_metrics_score.py
kasya, line 1: Could you update the file name to say `...health_metrics_scores.py` (plural)?

backend/apps/owasp/management/commands/owasp_update_project_health_metrics.py
```diff
@@ -0,0 +1,61 @@
...
+            "unanswered_issues_count": "unanswered_issues_count",
+            "unassigned_issues_count": "unassigned_issues_count",
+        }
+        updated_count = errors_count = 0
+        for project in projects:
+            try:
+                self.stdout.write(
+                    self.style.NOTICE(f"Updating metrics for project: {project.name}")
+                )
+                metrics = ProjectHealthMetrics.objects.get_or_create(project=project)[0]
```
kasya: We'd like to keep track of changes throughout the time. So we'd need to just add new records with this job and not update existing record. With this current approach we'll always have just one metrics record (current state) for a project.

Could you update this logic?

backend/apps/owasp/management/commands/owasp_update_project_health_metrics.py
```diff
@@ -0,0 +1,61 @@
...
+
+                # Update metrics based on requirements
+                for metric_field, project_field in field_mappings.items():
+                    value = getattr(project, project_field)
+                    setattr(metrics, metric_field, value)
```
kasya: This could be done inline:
```suggestion
                    setattr(metrics, metric_field, getattr(project, project_field))
```

backend/apps/owasp/management/commands/owasp_update_project_health_metrics.py
```diff
@@ -0,0 +1,61 @@
+"""A command to update OWASP project health metrics."""
+
+from django.core.management.base import BaseCommand
+
+from apps.owasp.models.project import Project
+from apps.owasp.models.project_health_metrics import ProjectHealthMetrics
+
+MINIMUM_LEADERS = 3
```
aramattamara: Could you add a comment on what it is, and why is 3?

backend/apps/owasp/management/commands/owasp_update_project_health_metrics.py
```diff
@@ -0,0 +1,61 @@
...
+                    self.style.NOTICE(f"Updating metrics for project: {project.name}")
+                )
+                metrics = ProjectHealthMetrics.objects.get_or_create(project=project)[0]
+
+                # Update metrics based on requirements
+                for metric_field, project_field in field_mappings.items():
+                    value = getattr(project, project_field)
+                    setattr(metrics, metric_field, value)
+
+                is_leaders_compliant = project.leaders_count >= MINIMUM_LEADERS
```
kasya: Could you move this into the model as a property? Similar to how we have `is_funding_requirements_compliant`.
We could call it `is_leaders_requirements_compliant`

backend/apps/owasp/management/commands/owasp_update_project_health_metrics.py
```diff
@@ -0,0 +1,61 @@
...
+                metrics.is_project_leaders_requirements_compliant = is_leaders_compliant
+                metrics.save()
+                updated_count += 1
+            except (AttributeError, ValueError, TypeError):
+                self.stdout.write(self.style.ERROR(f"Error updating project {project.name}"))
+                errors_count += 1
```
aramattamara: The information of the exception is missing from the error message. I suggest
```py
except (AttributeError, ValueError, TypeError) as e:
    ... f"Error updating project {project.name}: {e}"))
``` 
so we know what exactly was wrong.

aramattamara: Also, you shouldn't catch python errors AttributeError, ValueError, TypeError, which typically means a bug is in code.
I suggest not catching them and just letting the whole command fail, printing exception stack trace to the output automatically.

But you also don't want half the ProjectHealthMetrics to be saved, and then get an error on some of them, and some of them is not saved. For that, I suggest keeping them all in a list and then save all together:
```py
to_save = []
for project in projects:
   ...
   to_save.append(metrics)

ProjectHealthMetrics.objects.bulk_update(to_save)
```
This will save them all or none of them, not partially.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_score.py
```diff
@@ -0,0 +1,53 @@
...
+                    if metric_value >= requirement_value:
+                        score += weight
+
+                metric.score = score
+                metric.save(update_fields=["score"])
+                self.stdout.write(
+                    self.style.SUCCESS(f"Updated score for project {metric.project.name}: {score}")
+                )
+
+            except (AttributeError, ValueError, TypeError) as e:
```
aramattamara: The same as the previous comment, don't catch attribute errors.

backend/apps/owasp/models/project.py
```diff
@@ -140,6 +142,18 @@ def is_documentation_type(self) -> bool:
...
+    def is_funding_requirements_compliant(self) -> bool:
+        """Indicate whether project is compliant with funding requirements."""
+        if self.repositories_count == 0:
+            return True
+
+        compliant_repos = self.repositories.filter(
+            is_funding_policy_compliant=True,
+        ).count()
+
+        return compliant_repos == self.repositories_count
```
aramattamara: Probably simpler
```suggestion
        """Indicate whether all repos in project are compliant with funding requirements."""
        return not self.repositories.filter(is_funding_policy_compliant=False).exists()
```

backend/apps/owasp/models/project.py
```diff
@@ -182,6 +218,18 @@ def pull_requests(self):
...
+    def pull_requests_count(self) -> int:
+        """Return count of pull requests."""
+        return self.pull_requests.count()
+
+    @property
+    def pull_request_last_created_at(self) -> datetime.datetime | None:
+        """Return last created pull request."""
+        if pull_request := self.pull_requests.order_by("-created_at").first():
+            return pull_request.created_at
+        return None
```
aramattamara: Try:
```suggestion
return self.pull_requests.aggregate(Max("created_at"))["created_at__max"]
```
just check it works correct on empty `self.pull_requests`

aramattamara (dismissed):
Thanks for working on this. You'll need to fix some issues and make improvements.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics.py
ahmedxgouda: Actually it should be 2, it was stated in the description of the original idea: #711
image

backend/apps/owasp/models/project.py
ahmedxgouda: Yeah it worked.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_score.py
```diff
@@ -0,0 +1,52 @@
...
+        }
+        to_save = []
+        for metric in metrics:
+            # Calculate the score based on requirements
+            self.stdout.write(
+                self.style.NOTICE(f"Updating score for project: {metric.project.name}")
+            )
+            requirements = ProjectHealthRequirements.objects.get(level=metric.project.level)
+
+            score = 0.0
```
kasya: The idea is to have multiple ProjectHealthMetrics per project, per date so that we can then visualize health score change in time. This is why the parent model is a `TimestampModel`.

We should change `score` on the model to be nullable, instead of having a `0.0` default value. And then here we should only calculate new metrics score for those projects that have `null` value in score (meaning un-scored records only). This is a better approach because `0.0` technically could be a valid value.

Not sure if that's clear. Let me know if you have more questions about how to update this.

kasya (changes requested):
Thanks @ahmedxgouda ! 
I started doing this PR a few days ago and you've already addressed some of the concerns I had, but I still have a couple of things ⬇️

backend/apps/owasp/management/commands/owasp_update_project_health_metrics.py
ahmedxgouda: Yes, this is true, but isn't too much for a single line?
We could rename the variable value to a more meaning name like project_value. I think this is more readable.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_score.py
ahmedxgouda: It is clear, thanks! I updated the logic.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
```diff
@@ -0,0 +1,52 @@
...
+            requirements = ProjectHealthRequirements.objects.get(level=metric.project.level)
+
+            score = 0.0
+            for field, weight in weight_mapping.items():
+                metric_value = getattr(metric, field)
+                requirement_value = getattr(requirements, field)
+
+                if metric_value >= requirement_value:
+                    score += weight
+
```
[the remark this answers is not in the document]
ahmedxgouda: I think this maybe a good suggestion

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
```diff
@@ -0,0 +1,54 @@
...
+    def handle(self, *args, **options):
+        metrics = ProjectHealthMetrics.objects.filter(score__isnull=True)
+        weight_mapping = {
+            "age_days": (6.0, -1),
+            "contributors_count": (6.0, 1),
+            "forks_count": (6.0, 1),
+            "last_release_days": (6.0, -1),
+            "last_commit_days": (6.0, -1),
+            "open_issues_count": (6.0, 1),
```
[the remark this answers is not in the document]
ahmedxgouda: I don't know if this is correct @kasya

backend/apps/owasp/management/commands/owasp_update_project_health_metrics.py
kasya: @ahmedxgouda the main idea is that there's no point in creating a variable if you only use it once on the next line. 
If you feel like it is needed for readability - you can keep it as is. But I'm pretty sure @arkid15r won't like this either 😅

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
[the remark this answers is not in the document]
kasya: This is an interesting point! @ahmedxgouda technically, rabbit is correct - larger number of open issues should decrease the score 🤔 . What do you think though? I'd love to hear your reasoning about points for this.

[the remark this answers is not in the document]
ahmedxgouda: I was thinking that if a project has a fewer number of open issues, it means that it is less active, but if this number grows big, it will be an issue.
So, if a project has less number of open issues than the requirement will increase the score, but what if this project has nearly no open issues like 3, 2, 1, or even 0! I think this should have low score. Maybe we can add two thresholds for the open issues count, one for large numbers and one for very small numbers of open issues count.
What do you think @kasya?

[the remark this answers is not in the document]
ahmedxgouda: I'd update it with rabbit suggestion for now to make it simple.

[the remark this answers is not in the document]
kasya: @ahmedxgouda Thanks for sharing!  My only concern with this logic is that 0 issues might mean that the project just does not use this functionality at all to track any issues/tickets (maybe they use some external tool like Jira). Not sure how often that's the case inside OWASP projects, but could definitely be a thing. 

We can keep it as is for now 💯

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
```diff
@@ -0,0 +1,60 @@
...
+                metric_value = getattr(metric, field)
+                requirement_value = getattr(requirements, field)
+
+                if (metric_value >= requirement_value and direction == 1) or (
+                    metric_value <= requirement_value and direction == -1
+                ):
+                    score += weight
+
+            # Evaluate compliance with funding and leaders requirements
+            if metric.is_funding_requirements_compliant:
```
kasya: Why are we hardcoding this? Can we use general approach for these (mapping)?

backend/apps/owasp/models/project_health_metrics.py
```diff
@@ -62,6 +63,52 @@ class Meta:
         verbose_name="Unassigned issues", default=0
     )
 
+    @property
+    def age_days(self) -> int:
+        """Calculate project age in days."""
+        if self.created_at:
+            return (timezone.now() - self.created_at).days
+        return 0
```
kasya: Could you update all of these into a single statement?
```suggestion
        return (timezone.now() - self.created_at).days if self.created_at else 0
```

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
```diff
@@ -0,0 +1,60 @@
...
+from apps.owasp.models.project_health_metrics import ProjectHealthMetrics
+from apps.owasp.models.project_health_requirements import ProjectHealthRequirements
+
+
+class Command(BaseCommand):
+    help = "Update OWASP project health metrics score."
+
+    def handle(self, *args, **options):
+        metrics = ProjectHealthMetrics.objects.filter(score__isnull=True)
+        weight_mapping = {
```
kasya: The sum of these now equals 90. Is that intentional?
I'd assume it should be a `100`.

backend/tests/apps/owasp/management/commands/owasp_update_project_health_metrics_test.py
```diff
@@ -0,0 +1,88 @@
...
+        # Set the leaders count to meet compliance
+        mock_project.leaders_count = 2
+        mock_metrics_instance.is_project_leaders_requirements_compliant = True
+
+        # Execute command
+        with patch("sys.stdout", new=self.stdout):
+            call_command("owasp_update_project_health_metrics")
+
+        # Verify command output
+        assert "Evaluating metrics for project: Test Project" in self.stdout.getvalue()
```
kasya, line 75: Can we also check that records were created with the proper data?
If you can not check actual records created, can we at least check that `ProjectHealthMetrics.bulk_save()` has been called with expected data?

kasya (dismissed):
Left a couple of comments and questions ⬇️

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
ahmedxgouda: It was hardcoded because the fields had different weights, now they have the same weights. I will work on it. Thanks!

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
ahmedxgouda: Yes it is 90, but we have two fields that are not in the mapping which are `is_project_leaders_requirements_compliant` and `is_funding_requirements_compliant`. Each has a weight of 5. So, the overall score is 100.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
```diff
@@ -0,0 +1,64 @@
...
+                requirements_cache[level] = ProjectHealthRequirements.objects.get(level=level)
+
+            requirements = requirements_cache[level]
+            score = 0.0
+            for field in forward_fields:
+                if getattr(metric, field) >= getattr(requirements, field):
+                    score += 6.0
+            for field in backward_fields:
+                if getattr(metric, field) <= getattr(requirements, field):
+                    score += 6.0
```
arkid15r: This approach removes flexibility your code used to have -- a custom weight per metric. Please revert it to the previous approach.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
```diff
@@ -0,0 +1,64 @@
...
+
+            # Evaluate compliance with funding and leaders requirements
+            if metric.is_funding_requirements_compliant:
+                score += 5.0
+            if metric.is_project_leaders_requirements_compliant:
+                score += 5.0
```
arkid15r: These fields also need to be in the same mapping/config of metrics/weights.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
ahmedxgouda: I thought It was suggested by Kate here: https://github.com/OWASP/Nest/pull/1550#discussion_r2141308779
Maybe I got it wrong.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
ahmedxgouda: Okay, but these are not parts of the ProjectHealthRequirements model, so we need an additional if statement

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
```diff
@@ -0,0 +1,66 @@
...
+            "is_funding_requirements_compliant": 5.0,
+            "is_leader_requirements_compliant": 5.0,
+            "open_pull_requests_count": 6.0,
+            "recent_releases_count": 6.0,
+            "stars_count": 6.0,
+            "total_pull_requests_count": 6.0,
+            "total_releases_count": 6.0,
+        }
+        backward_fields = {
+            "age_days": 6.0,
```
arkid15r: Shouldn't this be in `forward_fields`?

backend/apps/owasp/management/commands/owasp_update_project_health_requirements.py
```diff
@@ -84,59 +92,41 @@ class Command(BaseCommand):
...
+                "owasp_page_last_update_days": 0,
+                "recent_releases_count": 0,
+                "recent_releases_time_window_days": 0,
+                "stars_count": 0,
+                "total_pull_requests_count": 0,
+                "total_releases_count": 0,
+                "unanswered_issues_count": 0,
+                "unassigned_issues_count": 0,
+            },
+        )
```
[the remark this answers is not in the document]
ahmedxgouda: Maybe a good suggestion from rabbit @arkid15r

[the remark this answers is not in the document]
arkid15r: Let's add tests for that instead? 
It's unexpected when project has unrecognizable level -- it should be fixed on their side. Technically we shouldn't even calculate a score for them.

backend/apps/owasp/management/commands/owasp_update_project_health_metrics_scores.py
ahmedxgouda: I assumed that young projects should have more score than old ones

arkid15r: Let's assume opposite for now -- the older project the better score.

backend/apps/owasp/management/commands/owasp_update_project_health_requirements.py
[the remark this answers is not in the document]
ahmedxgouda: You mean, we should assume that we don't have projects with unknown level?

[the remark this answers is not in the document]
arkid15r: We shouldn't have those. But if we do we can just skip them instead of building the whole defensive system w/ default level requirements.

arkid15r (approved):

PR #1566, opened 2025-06-08
Fix Recent Releases Card: Show Tag as Fallback and Correct Release URLs

### LINKED ISSUE #1537: Improve recent releases card
**Describe the bug**
For recent releases from external organizations (not from OWASP GitHub org) the card is missing some information. 

**To Reproduce**
For example https://nest.owasp.org/organizations/factionsecurity/repositories/faction#recent-releases

**Expected behavior**
- It should contain a readable name for all release items (if title is absent we should use the tag).
- The URL has to lead to the organization's repository release page (it currently leads to Nest main page)

- [ ] Yes
- [x] No

**Screenshots**

Image

rishyym0927:
This PR improves the Recent Releases card component with the following changes:

- Fixes #1537 

- Ensures that each release item displays a readable name: if the release name is absent, the tagName is shown as a fallback.

- Fixes the release link: if the url is missing or invalid, the link now correctly points to the organization's repository release page using the organization, repository, and tag.

- No changes are required in the backend as long as all necessary fields (name, tagName, url, organizationName, repositoryName) are present in the API response.

- These changes address issues where releases from external organizations were missing information or had incorrect links. No unrelated code was modified.


frontend/src/components/RecentReleases.tsx
```diff
@@ -67,11 +67,17 @@ const RecentReleases: React.FC<RecentReleasesProps> = ({
                   <h3 className="flex-1 overflow-hidden text-ellipsis whitespace-nowrap font-semibold">
                     <Link
                       className="text-blue-400 hover:underline"
-                      href={item?.url || '/'}
+                      href={
+                        item?.url && item?.url !== '/'
+                          ? item.url
+                          : item.organizationName && item.repositoryName && item.tagName
+                            ? `/organizations/${encodeURIComponent(item.organizationName)}/repositories/${encodeURIComponent(item.repositoryName)}/releases/tag/${encodeURIComponent(item.tagName)}`
```
kasya: We don't have any pages at that address 🤔 :
Screenshot 2025-06-07 at 9 20 04 PM

I believe this should lead to the release page on GitHub

kasya (changes requested):
Thanks for working on this.
One *big* issue I found ⬇️

frontend/src/components/RecentReleases.tsx
rishyym0927: Working to fix this bug

kasya:
@rishyym0927 Hi! Any updates on this one?

rishyym0927:
@kasya  can you clarify one thing, that if the url is not present then what should I use in replacement of that or how to deal with it? Also what if the tag name or title name both are missing how can we deal with it then? shouldn't we make these fields mandatory at the backend part ?

kasya:
> @kasya can you clarify one thing, that if the url is not present then what should I use in replacement of that or how to deal with it? Also what if the tag name or title name both are missing how can we deal with it then? shouldn't we make these fields mandatory at the backend part ?

@rishyym0927 the item you work with doesn't have a `url` property on it. You need to update the backend to pass it in.

I would recommend you to check how links for releases work on the [organization page](http://localhost:3000/organizations/OWASP#recent-releases). It's the same component.
Also, maybe look at [GraphQL](http://localhost:8000/graphql/) to check what data is available on the backend

kasya:
Also, tests are failing now. Please run `make check` and `make test` after addressing this and before pushing changes 👌🏼

rishyym0927:
image


Hey @kasya  as I can see that the graphql query return the url in recentReleases, so it does have the url property in it. What goes wrong is I gues that sometimes the when url link is not received then it basically redirect to home page of the nest where as we shoudl redirect it not found page. IF im correct then can I mplement this solution cause I dont think there is any need of making changes in hte backend as of now

> @rishyym0927 the item you work with doesn't have a `url` property on it. You need to update the backend to pass it in.
> 
> I would recommend you to check how links for releases work on the [organization page](http://localhost:3000/organizations/OWASP#recent-releases). It's the same component. Also, maybe look at [GraphQL](http://localhost:8000/graphql/) to check what data is available on the backend

Hey @kasya,
Thanks for the suggestion!

I checked the GraphQL query, and it looks like the recentReleases does return the url property. From what I observed, the issue seems to occur when the url is not present for a particular release — in such cases, it redirects to the home page of Nest. Ideally, we should redirect to a 404 or a "Not Found" page instead.

If my understanding is correct, would it make sense for me to implement this behavior on the frontend? I believe this might be sufficient for now, without requiring changes on the backend — but happy to discuss further if you think otherwise.

Let me know what you think!

rishyym0927:
I implemented a new solution which works pretty well, even if the url fiield is missing @kasya

kasya:
> 
> Hey @kasya, Thanks for the suggestion!
> 
> I checked the GraphQL query, and it looks like the recentReleases does return the url property. From what I observed, the issue seems to occur when the url is not present for a particular release — in such cases, it redirects to the home page of Nest. Ideally, we should redirect to a 404 or a "Not Found" page instead.
> 
> Let me know what you think!

@rishyym0927 have you compared the data on those 2 pages? The data that is sent to the project page releases does have the `url` property on it. 
Screenshot 2025-06-15 at 12 52 19 PM

But you don't have it in the objects you work with on the repo page.
Screenshot 2025-06-15 at 12 52 29 PM

GraphQL only shows you what's available - not necessarily all of it was sent to the frontend. And that's exactly the case for this page.

rishyym0927:
> > Hey @kasya, Thanks for the suggestion!
> > I checked the GraphQL query, and it looks like the recentReleases does return the url property. From what I observed, the issue seems to occur when the url is not present for a particular release — in such cases, it redirects to the home page of Nest. Ideally, we should redirect to a 404 or a "Not Found" page instead.
> > Let me know what you think!
> 
> @rishyym0927 have you compared the data on those 2 pages? The data that is sent to the project page releases does have the `url` property on it. Screenshot 2025-06-15 at 12 52 19 PM
> 
> But you don't have it in the objects you work with on the repo page. Screenshot 2025-06-15 at 12 52 29 PM
> 
> GraphQL only shows you what's available - not necessarily all of it was sent to the frontend. And that's exactly the case for this page.

Hey @kasya,
Thank you so much for the clarification and the screenshots — that really helped me understand the gap better.

You're absolutely right — I hadn’t compared the data between the project page and the repo page closely enough. I now see that while the GraphQL query can return the url, it isn’t being passed to the frontend in the repo page context, which is likely why the redirection issue occurs there.

Apologies for the oversight on my part. I’ll dig into this further

kasya (dismissed):
I cleaned this up a bit (removed prev.added  logic for working with dates that was not needed, we have a `formatDate()` util that takes care of that).

Also pushed missing part with getting `url` for a release in repository query. 
This works fine now! 👌🏼

arkid15r (approved):

PR #1569, opened 2025-06-08
Fix Authentication related bugs

### LINKED ISSUE #1555: Next auth follow up tasks
Here is a list of changes we need to make in order to keep it in a good shape:
 - update CI/CD workflows with valid env names (frontend auth related envs for staging/production)
 - implement it in a way it'd work for contributors w/o GitHub application set up
 - make the UserMenu look consistent w/ existing dropdown menu (Community)
 - fix e2e tests and figure out whether it's possible to keep the `turbo` mode for tests

Rajgupta36:
Resolves #1555 

tasks
- [x] Handle empty github credentials
- [x] fix unit and e2e tests


Rajgupta36:
@coderabbitai review it

frontend/.env.example
```diff
@@ -10,6 +10,6 @@ NEXT_PUBLIC_SENTRY_DSN=
 NEXT_SERVER_CSRF_URL=http://backend:8000/csrf/
 NEXT_SERVER_DISABLE_SSR=false
 NEXT_SERVER_GRAPHQL_URL=http://backend:8000/graphql/
-NEXT_SERVER_GITHUB_CLIENT_ID=
+NEXT_PUBLIC_GITHUB_CLIENT_ID=
```
arkid15r: Are we going to have this on the client (user browser)?

My understanding this convention works as `NEXT_SERVER_` -- for frontend server part and `NEXT_PUBLIC_` -- for frontend client part.

frontend/__tests__/unit/pages/Login.test.tsx
```diff
@@ -4,6 +4,7 @@ import { useRouter } from 'next/navigation'
 import { useSession, signIn } from 'next-auth/react'
 import { render } from 'wrappers/testUtil'
 import LoginPage from 'app/login/page'
+import { isAuthEnable } from 'utils/constants'
```
arkid15r:
```suggestion
import { isAuthEnabled } from 'utils/constants'
```

frontend/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -2,25 +2,36 @@ import { gql } from '@apollo/client'
...
+}
+
+const authOptions = {
+  providers,
   session: {
     strategy: 'jwt' as const,
   },
   callbacks: {
     async signIn({ account }) {
+      if (!isAuthEnable() && account?.provider === 'github') {
```
Comment (arkid15r) [inline frontend/src/app/api/auth/[...nextauth]/route.ts]: This needs to be narrowed down to `isGitHubAuthEnabled`

frontend/src/utils/constants.ts
```diff
@@ -123,3 +124,7 @@ export const userAuthStatus = {
   LOADING: 'loading',
   UNAUTHENTICATED: 'unauthenticated',
 }
+
+export const isAuthEnable = () => {
+  return Boolean(GITHUB_CLIENT_ID)
```
arkid15r: Let's check both client_id/secret exist.

arkid15r (changes requested):
My main question here is -- how do we make sure that frontend secrets are not available on the client side (user's browser)?

frontend/src/utils/constants.ts
Rajgupta36: for this i need to make it client side variable . that's vulnerable otherwise we can't access server side variable into client side

arkid15r: Fair enough, would make sense on server side only.

.github/workflows/run-ci-cd.yaml
```diff
@@ -504,8 +502,10 @@ jobs:
           echo "NEXT_PUBLIC_SENTRY_DSN=${{ secrets.VITE_SENTRY_DSN }}" >> frontend/.env
           echo "NEXT_SERVER_CSRF_URL=${{ secrets.NEXT_SERVER_CSRF_URL }}" >> frontend/.env
           echo "NEXT_SERVER_GRAPHQL_URL=${{ secrets.NEXT_SERVER_GRAPHQL_URL }}" >> frontend/.env
+          echo "NEXT_SERVER_GITHUB_CLIENT_ID=${{ secrets.GITHUB_CLIENT_ID }}" >> frontend/.env
+          echo "NEXT_SERVER_GITHUB_CLIENT_SECRET=${{ secrets.GITHUB_CLIENT_SECRET }}" >> frontend/.env
 
-      - name: Build frontend image
+          - name: Build frontend image
```
arkid15r: This looks wrong and causes CI/CD failure.

Rajgupta36:
> Does this address
> 
> * make the UserMenu look consistent w/ existing dropdown menu (Community)
>   ?

sorry i forgot to address. i updated it.

arkid15r (approved):

PR #1577, opened 2025-06-10
Implement GraphQL resolvers for project health metrics

### LINKED ISSUE #1539: Flag projects with health issues - Project Health Dashboard first task
**Flag projects with health issues:**
1. ​Old projects with no recent releases or updates.
2. ​New projects that have no recent commits or
3. releases in a set period.
4. ​Projects with fewer than two leaders or
5. contributors (we can call them.
6. ​Projects with non-compliant funding policies.

ahmedxgouda:
Resolves #1539

- Implemented a resolver that returns unhealthy projects with filtering options.
- Appended a health field to the `ProjectNode` that returns the metrics of the project.
- Updated the `ProjectHealthMetrics` and `Project` models with new fields and properties which will be helpful for filtering.
- Implemented the tests.

### DIFF
--- a/backend/apps/owasp/graphql/nodes/project.py
+++ b/backend/apps/owasp/graphql/nodes/project.py
@@ -39,9 +39,11 @@ class ProjectNode(GenericEntityNode):
     @strawberry.field
     def health_metrics(self, limit: int = 30) -> list[ProjectHealthMetricsNode]:
         """Resolve project health metrics."""
-        return ProjectHealthMetrics.objects.filter(project=self).order_by("-nest_created_at")[
-            :limit
-        ]
+        return ProjectHealthMetrics.objects.filter(
+            project=self,
+        ).order_by(
+            "-nest_created_at",
+        )[:limit]
 
     @strawberry.field
     def issues_count(self) -> int:
--- a/backend/apps/owasp/models/project_health_metrics.py
+++ b/backend/apps/owasp/models/project_health_metrics.py
@@ -23,7 +23,6 @@ class ProjectHealthMetrics(BulkSaveModel, TimestampedModel):
     contributors_count = models.PositiveIntegerField(verbose_name="Contributors", default=0)
     created_at = models.DateTimeField(verbose_name="Created at", blank=True, null=True)
     forks_count = models.PositiveIntegerField(verbose_name="Forks", default=0)
-
     is_funding_requirements_compliant = models.BooleanField(
         verbose_name="Is funding requirements compliant", default=False
     )


backend/apps/owasp/graphql/nodes/project.py
```diff
@@ -34,6 +36,11 @@
 class ProjectNode(GenericEntityNode):
     """Project node."""
 
+    @strawberry.field
+    def health(self) -> list[ProjectHealthMetricsNode]:
```
arkid15r: I recommend using full names, e.g. `health_metrics` in this case.

backend/apps/owasp/graphql/nodes/project.py
```diff
@@ -34,6 +36,11 @@
...
+        """Resolve project health metrics."""
+        return ProjectHealthMetrics.objects.filter(project=self).order_by("-nest_created_at")
+
```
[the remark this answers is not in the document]
arkid15r: @ahmedxgouda could you look into bot's suggestions and apply/resolve them?
This one makes sense and the reasonable limit could be `30`.

backend/apps/owasp/graphql/nodes/project_health_metrics.py
```diff
@@ -0,0 +1,55 @@
...
+    @strawberry.field
+    def last_pull_request_days(self) -> int:
+        """Resolve last pull request age in days."""
+        return self.last_pull_request_days
+
+    @strawberry.field
+    def last_release_days(self) -> int:
+        """Resolve last release age in days."""
+        return self.last_release_days
+
```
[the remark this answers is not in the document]
arkid15r: @ahmedxgouda have you checked if this suggestion is valid?

backend/apps/owasp/graphql/queries/project_health_metrics.py
```diff
@@ -0,0 +1,82 @@
...
+        *,
+        is_contributors_requirement_compliant: bool | None = None,
+        is_funding_requirements_compliant: bool | None = None,
+        has_no_recent_commits: bool | None = None,
+        has_recent_releases: bool | None = None,
+        is_leader_requirements_compliant: bool | None = None,
+        limit: int = 20,
+        has_long_open_issues: bool | None = None,
+        has_long_unanswered_issues: bool | None = None,
+        has_long_unassigned_issues: bool | None = None,
```
arkid15r: I'm really not sure if we need these set of filters supported.

backend/apps/owasp/graphql/queries/project_health_metrics.py
arkid15r, line 1: I don't think we need this file at all (if we extend the project endpoint to return project related metrics. Or if we prefer to keep this one it should just contain a simple endpoint to return the project related metrics (probably 30 most recent once to show the data on the FE).

backend/apps/owasp/models/project_health_metrics.py
```diff
@@ -23,6 +23,18 @@ class Meta:
...
+    has_no_recent_commits = models.BooleanField(
+        verbose_name="Has no recent commits", default=False
+    )
+    has_long_open_issues = models.BooleanField(verbose_name="Has long open issues", default=False)
+    has_long_unanswered_issues = models.BooleanField(
+        verbose_name="Has long unanswered issues", default=False
+    )
+    has_long_unassigned_issues = models.BooleanField(
+        verbose_name="Has long unassigned issues", default=False
+    )
```
arkid15r: Let's keep it as is, we already have all data we need for the front-end part.

arkid15r (changes requested):
I suggest to keep it simple while focusing on the working BE/FE PoC. For now we just need a project health metrics endpoint (I think having a separate endpoint would be more beneficial as it may be quite a lot of data -- a bunch of metric values for the recent 30 days).

Please also look into addressing the bot's comments.

backend/apps/owasp/graphql/queries/project_health_metrics.py
ahmedxgouda: Actually we maybe need it. There are some projects that already have high scores, but they have issues in these filters (I tested it on my side). Also, these filters were stated in the original idea's description.
For the bot's suggestions, most of them were before the last force-push (i.e. when the two branches are stacked). I checked the it's comments after the last changes, and I didn't find them correct at all. The limit's suggestion maybe before the force-push, so I didn't look at, but I'll update it.

backend/apps/owasp/graphql/queries/project_health_metrics.py
ahmedxgouda: It is for the dashboard -as I remember in the original idea's description- there should be an internal dashboard that are not visible in public that shows the unhealthy projects with filtering options. If you want me to revert this, I will do it.

backend/apps/owasp/models/project_health_metrics.py
ahmedxgouda: Okay, if you still don't want these filters, I will revert these changes.

backend/apps/owasp/graphql/queries/project_health_metrics.py
arkid15r: Ah, I see. Yes, we haven't planned that step out yet. We may be just okay w/ the Django admin interface offerings -- not sure yet. So yes, for now we should focus on user visible parts -- extending the project details page w/ the health related data.

backend/apps/owasp/graphql/nodes/project_health_metrics.py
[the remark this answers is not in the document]
ahmedxgouda: No, it is not. Even strawberry type does not have this attribute.

backend/tests/apps/owasp/models/project_health_metrics_test.py
```diff
@@ -82,3 +96,24 @@ def test_count_defaults(self, field_name, expected_default):
...
+    def test_handle_days_calculation(self, field_name, expected_days):
+        """Should return correct days for date fields."""
+        metrics = ProjectHealthMetrics()
+        metrics.created_at = self.FIXED_DATE
+        metrics.last_committed_at = self.FIXED_DATE
+        metrics.last_released_at = self.FIXED_DATE
+        metrics.owasp_page_last_updated_at = self.FIXED_DATE
+        metrics.pull_request_last_created_at = self.FIXED_DATE
+
+        assert getattr(metrics, field_name) == expected_days
```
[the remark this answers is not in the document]
ahmedxgouda: @arkid15r What do you think of this? It is related to the previous PR, but idk if this is correct

backend/tests/apps/owasp/graphql/nodes/project_health_metrics_test.py
```diff
@@ -0,0 +1,69 @@
...
+            ("last_release_days", int),
+            ("open_issues_count", int),
+            ("open_pull_requests_count", int),
+            ("owasp_page_last_update_days", int),
+            ("stars_count", int),
+            ("recent_releases_count", int),
+            ("unanswered_issues_count", int),
+            ("unassigned_issues_count", int),
+        ],
+    )
```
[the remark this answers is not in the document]
ahmedxgouda: Score could be None, so it is not possible to include within this test because the test checks if it is not None. What do you think? @arkid15r

arkid15r (changes requested):
As we don't have any model changes -- could you clean up the migrations?

arkid15r (approved):
Merging this, let's unblock https://github.com/OWASP/Nest/pull/1633

PR #1588, opened 2025-06-12
Rate limiting for messages sync

### LINKED ISSUE #1585: Implement Slack API rate limit handling mechanism
Based on https://github.com/OWASP/Nest/pull/1513 review

For message sync (and probably other Slack API calls heavy features) we need to have automatic rate limit handling mechanism. I'd extend it to error handling/retry where 429 error would be just on of the cases.

- [ ] Yes
- [x] No

**Additional context**
https://api.slack.com/apis/rate-limits

Dishant1804:
Resolves #1585 

- [x] added rate limiting and necessary delays to counter rate limits


arkid15r (approved):
I tried this code and it works. However, you can easily spot the wrong implementation approach when you see the same code in 3 places (actually more than one is enough -- DRY).

I'm merging this due to the time constaints we have. Please create an issue to unify the error handling

arkid15r (approved):

PR #1603, opened 2025-06-13
Update login page route

Rajgupta36:
Update auth route

### DIFF
--- a/frontend/src/app/api/auth/[...nextauth]/route.ts
+++ b/frontend/src/app/api/auth/[...nextauth]/route.ts
@@ -11,6 +11,7 @@ import {
 } from 'utils/credentials'
 
 const providers = []
+
 if (IS_GITHUB_AUTH_ENABLED) {
   providers.push(
     GitHubProvider({
--- a/frontend/src/components/LoginPageContent.tsx
+++ b/frontend/src/components/LoginPageContent.tsx
@@ -9,11 +9,11 @@ import { useSession, signIn } from 'next-auth/react'
 import { FC, useEffect } from 'react'
 import { userAuthStatus } from 'utils/constants'
 
-type Props = {
-  isGitHubAuthEnabled: boolean
+type LoginPageContentProps = {
+  readonly isGitHubAuthEnabled: boolean
 }
 
-const LoginPageContent: FC<Props> = ({ isGitHubAuthEnabled }) => {
+const LoginPageContent: FC<LoginPageContentProps> = ({ isGitHubAuthEnabled }) => {
   const { status } = useSession()
   const router = useRouter()
 


kasya (approved):
This works great 👍🏼 

Separate request though - could we hide the `Sign in` button if there's no `NEXT_SERVER_GITHUB_CLIENT_ID` in the `.env`? This way it will be safe to deploy to prod before we set up this whole thing for production.

Rajgupta36:
> This works great 👍🏼
> 
> Separate request though - could we hide the `Sign in` button if there's no `NEXT_SERVER_GITHUB_CLIENT_ID` in the `.env`? This way it will be safe to deploy to prod before we set up this whole thing for production.

We want but we can't . It's a security concern if we expose our github credentials on client side .

arkid15r:
> > This works great 👍🏼
> > Separate request though - could we hide the `Sign in` button if there's no `NEXT_SERVER_GITHUB_CLIENT_ID` in the `.env`? This way it will be safe to deploy to prod before we set up this whole thing for production.
> 
> We want but we can't . It's a security concern if we expose our github credentials on client side .

Why do you need to expose it on the client side? Why can't you have a similar flag indicating the github app presence on the server. It doesn't sound like impossible task to me (like the way you put it in your response).

The problem here is that we can't show the sign in button to users until it's all ready in prod. The alternative here is to use the feature branch.

Rajgupta36:
> > > This works great 👍🏼
> > > Separate request though - could we hide the `Sign in` button if there's no `NEXT_SERVER_GITHUB_CLIENT_ID` in the `.env`? This way it will be safe to deploy to prod before we set up this whole thing for production.
> > 
> > 
> > We want but we can't . It's a security concern if we expose our github credentials on client side .
> 
> Why do you need to expose it on the client side? Why can't you have a similar flag indicating the github app presence on the server. It doesn't sound like impossible task to me (like the way you put it in your response).
> 
> The problem here is that we can't show the sign in button to users until it's all ready in prod. The alternative here is to use the feature branch.


You're right. I didn't mean to imply it's impossible. we can definitely handle this by introducing a new flag that's scoped to both the client and server side. This would let us hide the sign-in button until everything is ready in production.

Rajgupta36:
Screenshot from 2025-06-16 23-51-39
when auth is disabled

arkid15r:
@Rajgupta36 could you test the new code @kasya added?

Rajgupta36:
> @Rajgupta36 could you test the new code @kasya added?

amazing ,It's  working

arkid15r (approved):
LGTM
@kasya @Rajgupta36 thanks for adding this!

PR #1621, opened 2025-06-16
Saving entire body of messsages

### LINKED ISSUE #1613: Save entire message body during for message sync
We're going to face the new Slack API limits soon

We want to save the entire message body during the message sync process.

- [ ] Yes
- [x] No

Dishant1804:
Resolves #1613 

- [x] messages whole body saved
- [x] thread replies whole body saved


backend/apps/slack/models/message.py
```diff
@@ -53,6 +54,7 @@ def from_slack(
         self.has_replies = message_data.get("reply_count", 0) > 0
         self.slack_message_id = message_data.get("ts", "")
         self.text = message_data.get("text", "")
+        self.message_body = message_data
```
arkid15r: It's already a message. This means we will face something like `message.message_body`.
```suggestion
        self.raw_data = message_data
```

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -322,18 +322,12 @@ def _create_message_from_data(
...
-                message_data.get("files"),
-                message_data.get("blocks"),
-            ]
-        ):
+        if not [
+            message_data.get("text"),
+            message_data.get("attachments"),
+            message_data.get("files"),
+            message_data.get("blocks"),
+        ]:
```
arkid15r: What do you mean by this?

arkid15r (changes requested):

backend/apps/slack/management/commands/slack_sync_messages.py
Dishant1804: I meant this
```
if not any([
    message_data.get("text"), 
    message_data.get("attachments"), 
    message_data.get("files"), 
    message_data.get("blocks")
]):
    return None
```

But I just removed any as well while removing the other code of `channel_join`, etc. which is a mistake from my end

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -322,11 +322,7 @@ def _create_message_from_data(
         parent_message: Message | None = None,
     ) -> Message | None:
         """Create Message instance using from_slack pattern."""
-        if message_data.get("subtype") in {
-            "channel_join",
-            "channel_leave",
-            "bot_message",
-        } or not any(
+        if not any(
```
arkid15r: One of the major goals here is to store all messages we can. What kind of messages do you skip here with this `if` and slack user / bot user related `if` below?

Also what messages don't have the author field?

We should avoid skipping any content.

arkid15r (changes requested):

backend/apps/slack/management/commands/slack_sync_messages.py
Dishant1804: Yeah so let's just remove that check, because I'm not sure what are other types of messages than these 4 
So it would be better if we just remove the `if` check

backend/apps/slack/models/message.py
```diff
@@ -20,11 +20,15 @@ class Meta:
 
     created_at = models.DateTimeField(verbose_name="Created at")
     has_replies = models.BooleanField(verbose_name="Has replies", default=False)
+    is_bot = models.BooleanField(verbose_name="Is bot", default=False)
```
arkid15r: This is a `member`'s attribute, not `message`'s

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -322,79 +322,89 @@ def _create_message_from_data(
...
-                message_data.get("files"),
-                message_data.get("blocks"),
-            ]
-        ):
-            return None
-
         try:
-            if not (slack_user_id := (message_data.get("user") or message_data.get("bot_id"))):
-                return None
+            slack_user_id = message_data.get("user")
```
arkid15r: Use just one try/except for member creation where just member data be different depending on user/bot case.

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -322,79 +322,89 @@ def _create_message_from_data(
...
+                                self.stdout.write(
+                                    self.style.WARNING(f"Failed to fetch bot data for {bot_id}.")
+                                )
+                                break
 
             return Message.update_data(
                 data=message_data,
                 conversation=conversation,
-                author=author,
+                author=author if isinstance(author, Member) else None,
```
arkid15r:
```suggestion
                author=author,
```

arkid15r (changes requested):

backend/apps/slack/models/message.py
Dishant1804: I think we should keep it this will help us to filter out the bot messages during the retrieval

arkid15r: You can do it by excluding message authors with `is_bot` set to True.

backend/apps/slack/management/commands/slack_sync_messages.py
arkid15r, line 1: It seems you still have 2 separate blocks -- 1 for user and another for bot. It's the same Member entity just with different data. Handle it within a single try/except block.

PR #1633, opened 2025-06-18
Implement charts for health metrics

### LINKED ISSUE #1622: Add Project Health Visualization to Project Page
Create a new section on the **Project page** to visualize project health metrics. This section should appear at the bottom of the page, with the total project score displayed prominently at the top of the page.

The health section should include multiple charts visualizing trends and activity over the **last 30 days**. 

---

#### 1. Issues Health Trend
- **Metrics:** `open_issues_count`, `unassigned_issues_count`
- **Chart type:** Line chart with two lines
- **Purpose:** Highlights issue volume and triage responsiveness over time

#### 2. Unanswered Issues Trend
- **Metrics:** `open_issues_count`, `unanswered_issues_count`
- **Chart type:** Line chart with two lines
- **Purpose:** Shows proportion of issues that are going unanswered

#### 3. Activity Trend
- **Metric:** `open_pull_requests_count`
- **Chart type:** Line chart
- **Purpose:** Reflects repository activity and engagement

#### 4. Community Size 
- **Metrics:** `stars_count`, `forks_count` (two separate charts)
- **Chart type:** Line chart over time
- **Purpose:** Demonstrates community interest and adoption

#### 5. Days Since Last Commit / Last Release
- **Derived from:** `last_committed_at`, `last_released_at` (two charts)
- **Chart type:** Radial progress bar
- **Purpose:** Shows recency of activity using status color thresholds:
  - 🟢 Less than 3 days
  - 🟡 3–20 days
  - 🔴 More than 20 days

---

### Technical Considerations

- Use a charting library such as [ApexCharts](https://apexcharts.com/) for rendering visualizations
- Charts should be responsive and consistent with existing UI styling
- Data should represent daily snapshots from the last 30 days
- Reuse chart components where appropriate to avoid duplication

---

### Notes

- Additional metrics such as contributors, compliance status, or other indicators may be added later
- The project health score should be displayed clearly at the top of the Project page for quick reference

- [ ] Yes
- [x] No

ahmedxgouda:
Resolves #1622

https://github.com/user-attachments/assets/69aded0f-f4d0-4203-b567-f4a69cd52bed


frontend/src/components/HealthMetricsCharts.tsx
arkid15r, line 1: Let's consolidate the issues related charts -- have a single chart w/ all data (total issues, unassigned, unanswered) instead of

Screenshot 2025-06-18 at 2 48 32 PM

frontend/src/components/GeneralCharts.tsx
arkid15r, line 1: I suggest splitting this component on per chart type basis.

frontend/src/components/HealthMetricsCharts.tsx
```diff
@@ -0,0 +1,104 @@
...
+              name: 'Unanswered Issues',
+              data: data.map((item) => item.unansweredIssuesCount),
+            },
+          ]}
+          labels={labels}
+          icon={faQuestionCircle}
+        />
+      </div>
+      <LineChart
+        title="Activity Trend"
```
arkid15r:
```suggestion
        title="Pull Requests"
```

frontend/src/components/HealthMetricsCharts.tsx
```diff
@@ -0,0 +1,104 @@
...
+              name: 'Forks',
+              data: data.map((item) => item.forksCount),
+            },
+          ]}
+          labels={labels}
+          icon={faCodeFork}
+        />
+      </div>
+      <div className="grid grid-cols-1 gap-4 md:grid-cols-2">
+        <RadialChart
```
arkid15r: We want to show a single value here -- the latest one (e.g. how long since today the last commit was. Same for the last release.

arkid15r (changes requested):
Great PoC 👍 
We need to play w/ the style a bit and have enough data to visualize.

You can start with addressing any of these:

frontend/src/components/CardDetailsPage.tsx
```diff
@@ -199,6 +201,7 @@ const DetailsCard = ({
               <RepositoriesCard repositories={repositories} />
             </SecondaryCard>
           )}
+        {type === 'project' && <HealthMetricsCharts data={healthMetricsData} />}
```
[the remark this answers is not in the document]
kasya: I actually came here to suggest this myself 😆 
@ahmedxgouda can we render this dynamically based on if there's Metrics data? I know we'll probably have it, but there might be a case when we already have a project, but haven't synced their data and haven't calculated the score yet. 
This looks weird on the FE in that case:
Screenshot 2025-06-18 at 6 23 02 PM

frontend/src/components/GeneralCharts.tsx
```diff
@@ -0,0 +1,85 @@
...
+          },
+        }}
+        series={series}
+        height={450}
+      />
+    </SecondaryCard>
+  )
+}
+
+const RadialChart: React.FC<{
```
kasya: Can we do [Gradient radial chart](https://apexcharts.com/react-chart-demos/radialbar-charts/gradient/) instead?
Currently this one if so multiple series, but this should work with a single number (number of days since last release/commit). 
Or we could try a [Semi-circle one](https://apexcharts.com/react-chart-demos/radialbar-charts/semi-circle-gauge/)

Basically it should get from green to red based on number of days.

frontend/src/components/GeneralCharts.tsx
```diff
@@ -0,0 +1,85 @@
...
+  title: string
+  icon?: IconProp
+  labels: string[]
+  series: number[]
+}> = ({ title, series, icon, labels }) => {
+  const colors = series.map((days) => {
```
kasya: I saw your comment in today's update about colors for dark theme - I'm pretty sure we'll be able to pass in values dynamically based on the theme.
We could get the theme with `import { useTheme } from 'next-themes'` and then:
```suggestion
  const { theme } = useTheme();
  const colors = series.map((days) => {
```

And then render different colors based on that value.

frontend/src/components/GeneralCharts.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    return `${(total / w.globals.series.length).toFixed(0)} days`
+                  },
+                },
+              },
+            },
+          },
+        }}
+        series={series}
+        height={450}
+        type="radialBar"
```
kasya: You would then need to add key here to trigger re-render of the component on theme change:
```suggestion
        type="radialBar"
        key={theme}
```

frontend/src/components/GeneralCharts.tsx
kasya: I did however plan to update requirements for this last chart, but you're so fast to implement stuff 😅

There should actually be just 1 breakpoint in this chart where before it - it's good, green.. and after - it's red and bad'. But since the path from green to red goes through orange - we'll still have it there. The numbers for that breakpoint are different based on the project level.
For example (totally random numbers here!), for Flagship level project there should be at least 1 release a month. So, if they are pass 30 days since last release - it should show days number and the chart at the red. 

Does that make sense? I'm not sure if I was able to explain it well. 
Let me know in Slack if you need me to clarify this. I can try to draw a picture :)

frontend/src/components/GeneralCharts.tsx
kasya: Same goes for another chart to make colors work with the theme 👍🏼

kasya (commented):
Nice work!!! I know it's still in draft, but I also have a few comments 👍🏼

frontend/src/components/HealthMetrics.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <div className="relative mt-4 flex w-full items-center">
+        <div className="flex-grow border-t border-gray-300"></div>
+        <h3 className="flex items-center gap-2 text-lg font-semibold text-gray-700 dark:text-gray-200">
+          <FontAwesomeIcon icon={faHeart} />
+          Health Metrics
+        </h3>
+        <div className="flex-grow border-t border-gray-300"></div>
+      </div>
+      <h4 className="mb-4 text-center text-lg font-semibold text-gray-700 dark:text-gray-200">
+        Score: {data[0].score}
```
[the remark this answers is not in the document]
ahmedxgouda: The length of data is checked when rendering the component in the DetailsCard component.

frontend/src/components/LineChart.tsx
```diff
@@ -0,0 +1,33 @@
...
+  labels?: string[]
+  icon?: IconProp
+}> = ({ title, series, labels, icon }) => {
+  return (
+    <SecondaryCard title={title} icon={icon}>
+      <Chart
+        options={{
+          xaxis: {
+            categories: labels,
+          },
```
kasya, line 49: Can we remove the toolbar on all of these?

```suggestion
          },
          chart: {
            toolbar: {
              show: false,
            },
          },
```

frontend/src/components/CardDetailsPage.tsx
```diff
@@ -191,6 +191,9 @@ const DetailsCard = ({
             <RecentReleases data={recentReleases} showAvatar={showAvatar} showSingleColumn={true} />
           </div>
         )}
+        {type === 'project' && healthMetricsData.length > 0 && (
```
kasya: Could you move this down, to be after the repositories block?

frontend/src/components/GradientRadialChart.tsx
```diff
@@ -0,0 +1,107 @@
...
+  let checkpoint = requirement
+  let showRed = false
+
+  if (days > requirement) {
+    checkpoint = days
+    showRed = true
+  }
+
+  if (checkpoint === 0) {
+    checkpoint = 1 // Avoid division by zero
```
kasya: This could be done inline:
```suggestion
    checkpoint = days || 1
    showRed = true
  }
```

frontend/src/components/LineChart.tsx
kasya, line 49: Also, can we make the stroke smooth? 
```suggestion
          },
          stroke: {
                curve: 'smooth'
              },
```
To compare (regular vs smooth):
Screenshot 2025-06-20 at 12 16 16 PM
Screenshot 2025-06-20 at 12 18 32 PM

frontend/src/components/LineChart.tsx
```diff
@@ -0,0 +1,33 @@
...
+        }}
+        series={series}
+        height={450}
```
kasya: Also, can we make them a bit narrower? I don't expect the line chrts we use to have a big range of numbers shown, so it should be fine with smaller charts:
```suggestion
        height={200}
```
Here's 200:
Screenshot 2025-06-20 at 12 21 12 PM

frontend/src/components/LineChart.tsx
kasya, line 1: Can we add support for dark theme? Right now the details of charts are not really visible in a dark theme 🤔 
Screenshot 2025-06-20 at 6 23 36 PM

frontend/src/components/HealthMetrics.tsx
```diff
@@ -84,21 +86,23 @@ const HealthMetricsCharts: React.FC<{ data: HealthMetricsProps[] }> = ({ data })
...
       </div>
       <div className="grid grid-cols-1 gap-4 md:grid-cols-2">
-        <RadialChart
+        <GradientRadialChart
           title="Days Since Last Commit"
           icon={faCodeCommit}
-          series={data.map((item) => item.lastCommitDays)}
-          labels={labels}
+          days={data[0]?.lastCommitDays ?? 0}
+          requirement={data[0]?.lastCommitDaysRequirement ?? 0}
```
kasya: While reviewing this, I noticed that it always returns 365 for any project I test with. This behavior is expected based on the current setup - it’s not a but - but we’ll need to adjust it to better reflect reality. Naturally, the number of releases and commits should differ between Flagship and Incubator projects.

I'll sync with Ark to clarify what the expected values should be, and once we have that direction, I’ll create a separate task to update the logic accordingly. 👍🏼

kasya (changes requested):
Nice work!! Left some requests!

kasya:
@ahmedxgouda I also just pushed some updates for Radial Chart styling. I realized that gradient fill works not as I expected it to, so switched that to use static gradient background image instead. 
Could you review those changes? 

Also, I realized that it would be super helpful to have a "legend" of what green and red mean on this scale. Could you try adding something like this to the radial chart (`Active` and `Stale`)?
image

Also, if we could pluralize the `day(s)` in the caption that would be great. So that it would say `1 day` and `5 days`. 
Screenshot 2025-06-20 at 7 15 04 PM
We already have a utility for that 👍🏼

frontend/src/components/LineChart.tsx
kasya: Also, this should not say `undefined` on the tooltip 👀 just noticed this @ahmedxgouda

ahmedxgouda: I wonder why it is saying that 🤔. I will look at it, thanks!

ahmedxgouda: What project is this? It is okay on my side. I tested multiple projects including Nest 🤔 @kasya 
image

frontend/src/components/GradientRadialChart.tsx
```diff
@@ -0,0 +1,121 @@
...
+              shadeIntensity: 0.5,
+              stops: stops,
+              colorStops: colorStops,
+            },
+          },
+        }}
+        series={[requirement ? normalizedDays : 0]}
+        height={400}
+        type="radialBar"
+      />
```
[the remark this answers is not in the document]
ahmedxgouda: It is not available in the Chart component!

frontend/src/components/GradientRadialChart.tsx
```diff
@@ -0,0 +1,121 @@
...
+            gradient: {
+              shade: theme,
+              type: 'horizontal',
+              shadeIntensity: 0.5,
+              stops: stops,
+              colorStops: colorStops,
+            },
+          },
+        }}
+        series={[requirement ? normalizedDays : 0]}
```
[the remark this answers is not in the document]
ahmedxgouda: No, a project should have requirements that are not 0

frontend/src/components/LineChart.tsx
ahmedxgouda: Maybe it happened when Next made a fast reload post your changes.

frontend/src/components/HealthMetrics.tsx
```diff
@@ -0,0 +1,106 @@
...
+} from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import React from 'react'
+import type { HealthMetricsProps } from 'types/healthMetrics'
+import GradientRadialChart from 'components/GradientRadialChart'
+import LineChart from 'components/LineChart'
+
+const HealthMetrics: React.FC<{ data: HealthMetricsProps[] }> = ({ data }) => {
+  const openIssuesCountArray = data.map((item) => item.openIssuesCount)
+  const labels = data.map((item, index) => `Day ${index + 1}`)
```
kasya: We need to update resolvers for all of these data points that we use on charts to also return the date. So this should be not just a number, but a date and number of Open issues at that date. So that it would show the date on the tooltip and on the x axis instead of just Day 1 etc.
Same for other data points.
Screenshot 2025-06-21 at 12 31 25 PM

frontend/src/components/HealthMetrics.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <div className="relative mt-4 flex w-full items-center">
+        <div className="flex-grow border-t border-gray-300"></div>
+        <h3 className="flex items-center gap-2 text-lg font-semibold text-gray-700 dark:text-gray-200">
+          <FontAwesomeIcon icon={faHeart} />
+          Health Metrics
+        </h3>
+        <div className="flex-grow border-t border-gray-300"></div>
+      </div>
+      <h4 className="mb-4 text-center text-lg font-semibold text-gray-700 dark:text-gray-200">
+        Score: {data[0].score}
```
kasya: Can we remove this whole divider and add a score at the top of the page. It should be something like a circle of different background color (based on the score, again from green to red) and the score number. 
We want it to be more prominent at the top of the page. It could also be clickable to scroll to the metrics charts lower on the page 🤔

frontend/public/img/gradient.png
kasya, line 1: Could you delete this file if we no longer use it?

frontend/src/components/GradientRadialChart.tsx
```diff
@@ -0,0 +1,129 @@
...
+    showRed = true
+  }
+  // Ensure checkpoint is at least 1 to avoid division by zero
+  checkpoint = checkpoint || 1
+  // Normalize days based on the checkpoint
+  const normalizedDays = (days / checkpoint) * 100
+  const normalizedRequirement = (requirement / checkpoint) * 100
+
+  if (showRed) {
+    const orangeStop = normalizedRequirement / 2
```
kasya: I had this case when with 5 days requirement it still showed all green even though technically it should have showed at least orange. 
Screenshot 2025-06-21 at 11 24 59 AM
I can see we only set orange if `showRed` is truthy. Can we change this logic? Also, maybe we should start showing orange at 75% instead of 50%? I'm not sure, but would love to see how that looks with setups when project is almost at stale.

kasya (changes requested):
@ahmedxgouda left a couple more comments!
I've also fixed the issue with extra space in firefox and will push the changes shortly.

frontend/src/components/HealthMetrics.tsx
ahmedxgouda: Another chart again or just a circle? @kasya

kasya: @ahmedxgouda I meant that clicking it should scroll to these charts on the project page. Maybe be the first one `Issues trend`? 🤔 We can update that later on, not necessarily with this PR 👍🏼

frontend/src/types/healthMetrics.ts
```diff
@@ -1,4 +1,5 @@
 export type HealthMetricsProps = {
+  createdAt: string
```
kasya, line 2: Great call!

frontend/src/components/GradientRadialChart.tsx
kasya, line 1: It's not exactly working as it should regarding the gradient. This is why I feel like static  image worked better. Let me add some screenshots to maybe help me explain.
Here are 2 different projects charts for releases:
Screenshot 2025-06-21 at 5 54 38 PM
Screenshot 2025-06-21 at 5 54 49 PM
Based on the second one - the first one should already have yellow/orange color showing on the fill (I marked approx. spot on the full circle chart).

Same for these two - looking at releases at the some spot - the commits chart should have some yellow-ish shade already:
Screenshot 2025-06-21 at 5 52 01 PM

Is this possible with the stops passed in? If not, we'll have to go back to static image.

frontend/src/components/HealthMetrics.tsx
```diff
@@ -3,32 +3,24 @@ import {
...
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import React from 'react'
 import type { HealthMetricsProps } from 'types/healthMetrics'
 import GradientRadialChart from 'components/GradientRadialChart'
 import LineChart from 'components/LineChart'
 
 const HealthMetrics: React.FC<{ data: HealthMetricsProps[] }> = ({ data }) => {
   const openIssuesCountArray = data.map((item) => item.openIssuesCount)
-  const labels = data.map((item, index) => `Day ${index + 1}`)
+  const labels = data.map((item) => {
```
kasya, line 16: Thanks for adding dates, but they are showing backwards  😅 I think it should be Jun 20th, and then Jun 21st. Could you order them by date (preferably on the backend)?
Screenshot 2025-06-21 at 5 49 53 PM

frontend/src/components/CardDetailsPage.tsx
```diff
@@ -49,11 +50,27 @@ const DetailsCard = ({
   geolocationData = null,
   repositories = [],
 }: DetailsCardProps) => {
+  let scoreStyle = 'bg-green-200 text-green-800'
```
kasya: Can we do it as a circle instead on the right side of the page. It should specify that this is a health score and show a number.
Here's a very rough sketch of what I had in mind!
image

frontend/src/components/LineChart.tsx
kasya, line 1: Also, for line charts - is there a way to customize y axes? I don't think it makes sense to show `.5` on these -  there can not be half a pull request or half a star.
Screenshot 2025-06-21 at 6 10 38 PM

kasya (changes requested):
We are almost there! 💪🏼 
Left a few more requests ⬇️

frontend/src/components/GradientRadialChart.tsx
ahmedxgouda: Actually, I don't feel the static image is a good idea.
Let me explain the logic further. We determined two checkpoints, one for the gradient to start going to orange and the other to start going to red. The fill of the gradient and the checkpoints are determined based on mathematical calculations. For example, the orange checkpoint is when the days hits 3/4 of the requirement. The red checkpoint is when the days exceeded the requirement. Any other case, the bar still green.
The bars are scaled based on their requirements. It is not reasonable to compare between two bars with different requirements, because they have different scales. The first two that you provided, not just they have different scales, the second one's scale is twice as the first one's. For the ending checkpoint, we determined that it is either the days or the requirement (just the bigger). So, here we can say that why the second two do not have the same color at the same bar position. They actually don't have the same scale! the second has a scale of 6 -which is the days- and the first has a skill of 5 - which is the requirement.
So, we can say, to resolve the issue that you stated. We need like a formula to get a fixed end checkpoint which does not depend on conditions, although this is so accurate.
The logic can be improved within time, we are close to the ideal approach. I don't think it is a good idea to just stick with the static image. You know, it is static, and may fail to represent the value accurately.
Sorry for this long text, just wanted to explain my thoughts. I'd like to hear your opinion of that of course. Thanks! @kasya

kasya: Oooh, I finally got where the confusion is coming from for both of us! We were looking at this chart differently.

As I see it - even the full fill circle should not have 6 days as a scale as you mentioned above - it should still just be the requirement (so, 5). But if the project is passed that (say, 10 days) it should just show red, meaning that the project is stale and passed the requirement.

So, to simplify this, the days since last release should not affect the scale of the chart - it should only change the percentage of the "filled" part of the chart.
And that makes more sense to me, because for example, if we look at these 2 charts above - the one for commits has 2 days and 5 requirement. It's filled to approx. where 2 would be on that scale. But then on the left you have the same requirement (5) but the scale is different. I find it confusing tbh. I would that expect them to be different only if the requirement is different, because that's what represents the chart - number of days from active to stale.

Does that makes sense? what do you think?

kasya: @ahmedxgouda we are discussing this with @arkid15r! Please do not work on any updates until further notice :) It might be we keep it this way.

frontend/src/components/LineChart.tsx
```diff
@@ -0,0 +1,50 @@
...
+            toolbar: {
+              show: false,
+            },
+            foreColor: color,
+          },
+          tooltip: {
+            theme: theme,
+          },
+          xaxis: {
+            categories: labels,
```
kasya, line 37: Let's add `tickAmount` property here to only show maximum of 10 or 15 labels for x axis. 
```suggestion
          xaxis: {
            categories: labels,
            tickAmount: 10,
```
I tested this out and showing all 30 won't work with our chart size (dates are random hardcoded dates):
Screenshot 2025-06-22 at 6 29 45 PM
Screenshot 2025-06-22 at 6 29 30 PM

frontend/src/components/GradientRadialChart.tsx
kasya: @ahmedxgouda we discussed this with Ark and decided to keep it as is for now 👍🏼 
The only issue I noticed is with the gradient state when the project is very much pass their requirement. 
Example:
Screenshot 2025-06-22 at 5 48 35 PM
Since the gradient is just changing color horizontally - this becomes visible the more red we have.
So we thought maybe it's better to show only red color for the whole bar if days are over 300% of requirement.
So, for example above with requirement of 5 days, if the  days are 15 - this should be full red:
Screenshot 2025-06-22 at 6 39 51 PM

I don't expect this to happen too often, but we should consider this case as well.

kasya (changes requested):

PR #1646, opened 2025-06-21
feature: DJango Ninja and drf cleanup

### LINKED ISSUE #1601: Migrate REST API to Django ninja
TBD

- [ ] Yes
- [x] No

abhayymishraa:
Resolves  #1601 

- [x] Migrated from drf to django ninja 
 

Add the PR description here.


backend/apps/owasp/api/chapter.py
```diff
@@ -1,28 +1,29 @@
...
-# ViewSets define the view behavior.
-class ChapterViewSet(viewsets.ReadOnlyModelViewSet):
-    """Chapter view set."""
 
-    queryset = Chapter.objects.all()
-    serializer_class = ChapterSerializer
+@router.get("/", response=list[ChapterSchema])
+def get_chapters(request: HttpRequest) -> list[ChapterSchema]:
+    """Get all chapters."""
+    return Chapter.objects.all()
```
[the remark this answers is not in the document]
abhayymishraa: @arkid15r  should we apply pagination ?

backend/apps/github/api/issue.py
```diff
@@ -1,29 +1,30 @@
...
 
 from apps.github.models.issue import Issue
 
+router = Router()
+
 
-# Serializers define the API representation.
-class IssueSerializer(serializers.HyperlinkedModelSerializer):
-    """Issue serializer."""
+class IssueSchema(BaseModel):
```
arkid15r: Why base on this instead of `ninja::Schema`?

backend/apps/github/api/issue.py
```diff
@@ -1,29 +1,30 @@
...
+    url: str
 
-# ViewSets define the view behavior.
-class IssueViewSet(viewsets.ReadOnlyModelViewSet):
-    """Issue view set."""
 
-    queryset = Issue.objects.all()
-    serializer_class = IssueSerializer
+@router.get("/", response=list[IssueSchema])
+def get_issue(request: HttpRequest) -> list[IssueSchema]:
```
arkid15r: This should be plural and probably `list_issues`.

backend/apps/github/api/issue.py
```diff
@@ -1,29 +1,30 @@
...
+    """Get all issues."""
```
arkid15r:
```suggestion
    """Get issues."""
```

backend/apps/owasp/api/chapter.py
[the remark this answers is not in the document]
arkid15r: Yes, this is a must have one.

backend/apps/github/api/issue.py
arkid15r, line 1: We probably should add `/v1/` for consistent versioning approach w/ API urls (`/api/v1`)

backend/pyproject.toml
```diff
@@ -119,6 +119,7 @@ target-version = "py313"
 extend-select = ["I"]
 ignore = [
     "ANN",  # TODO(arkid15r): remove when all annotations are added.
+    "ARG001",
```
arkid15r: This is too wide, narrow it down to your specific cases/paths only.

backend/apps/github/api/user.py
```diff
@@ -1,60 +1,49 @@
...
+    name: str
+    public_repositories_count: int
+    title: str
+    twitter_username: str
+    updated_at: datetime
+    url: str
+
+
+@router.get("/", response=list[UserSchema])
+def get_user(request: HttpRequest) -> list[UserSchema]:
```
arkid15r:
```suggestion
def list_users(request: HttpRequest) -> list[UserSchema]:
```

backend/apps/github/api/user.py
```diff
@@ -1,60 +1,49 @@
...
-            return Response(data)
-        except User.DoesNotExist:
-            return Response({"detail": "User not found."}, status=404)
+@router.get("/login/{login}", response=UserSchema)
+def get_user_by_login(request: HttpRequest, login: str) -> UserSchema | None:
+    """Get user by login."""
+    try:
+        return User.objects.get(login=login)
+    except User.DoesNotExist:
+        return Response({"detail": "User not found."}, status=200)
```
arkid15r: This is generally 404, something like
```python
from ninja.errors import HttpError

@router.get("/{login}", response={200: UserSchema, 404: dict})
def get_user(request: HttpRequest, login: str):
    user = User.objects.filter(login=login).first()
    if not user:
        raise HttpError(404, "User not found")
    return user
```

aramattamara (commented):
LGTM! But I leave to @arkid15r to approve the changes.

backend/tests/apps/owasp/api/v1/chapter_test.py
```diff
@@ -29,7 +31,9 @@
     ],
 )
 def test_chapter_serializer_validation(data, expected):
```
arkid15r: `expected` is unused here and below

arkid15r (changes requested):
A couple more questions:
- can pagination be done in a unified fashion via `default_router` param (see the very bottom of https://django-ninja.dev/guides/response/pagination/) -- `api = NinjaAPI(default_router=RouterPaginated())`
- Let's make default and max limit for the API request constant `100`
- Let's add response codes to the schema
- We should think about adding a default order for API response data set to make sure it always returns the same result for same requests

abhayymishraa:
> can pagination be done in a unified fashion via default_router param (see the very bottom of https://django-ninja.dev/guides/response/pagination/) -- api = NinjaAPI(default_router=RouterPaginated())

While the [Django Ninja docs](https://django-ninja.dev/guides/response/pagination/#setting-default-pagination) suggest using default_router=RouterPaginated() for unified pagination, in practice, this may not yield paginated responses unless you explicitly use @paginate on individual routes.

In our case, simply initializing the API with Router() and relying on RouterPaginated globally did not produce paginated responses. Instead, applying the @paginate(PageNumberPagination, page_size=100) decorator directly to the route ensures the correct paginated output.

So, while default_router=RouterPaginated() seems intended for unified pagination setup, it doesn’t behave as expected unless paired with the @paginate decorator per endpoint — at least in the current version or your specific configuration.

arkid15r (approved):

PR #1650, opened 2025-06-21
Fix inactive label alignment

kasya:
Fixed inactive label alignment on the DetailsCard.
Before and after:

Screenshot 2025-06-21 at 2 41 55 PM
Screenshot 2025-06-21 at 2 49 56 PM


arkid15r (approved):
LGTM

PR #1651, opened 2025-06-22
Chunk model and embeddings of message

### LINKED ISSUE #1645: Create chunk model and embeddings
Create chunk model and store the chunks and embeddings of the messages

- [x] Yes
- [ ] No

**Additional context**

Dishant1804:
Resolves #1645 

- [x] created chunk model


backend/apps/ai/models/chunk.py
```diff
@@ -0,0 +1,70 @@
...
+    message = models.ForeignKey(Message, on_delete=models.CASCADE, related_name="chunks")
+    chunk_text = models.TextField(verbose_name="Chunk Text")
+    embedding = VectorField(verbose_name="Chunk Embedding", dimensions=1536)
+
+    def __str__(self):
+        """Human readable representation."""
+        text_preview = truncate(self.chunk_text, 50)
+        return f"Chunk {self.id} for Message {self.message.slack_message_id}: {text_preview}"
+
+    def from_chunk(self, chunk_text: str, message: Message, embedding=None) -> None:
```
arkid15r: This is too simple for an extra method and also `chunk.from_chunk` looks odd -- just pass all data within `Chunk()` creation

backend/apps/slack/management/commands/slack_sync_messages.py
```diff
@@ -64,7 +64,7 @@ def handle(self, *args, **options):
             conversations = (
                 Conversation.objects.filter(slack_channel_id=channel_id)
                 if channel_id
-                else Conversation.objects.filter(sync_messages=True, workspace=workspace)
+                else Conversation.objects.filter(workspace=workspace)
```
arkid15r: Please merge properly.

backend/apps/ai/models/chunk.py
```diff
@@ -0,0 +1,70 @@
...
+from apps.common.models import BulkSaveModel, TimestampedModel
+from apps.common.utils import truncate
+from apps.slack.models.message import Message
+
+
+class Chunk(TimestampedModel):
+    """Slack Chunk model."""
+
+    class Meta:
+        db_table = "slack_chunks"
```
arkid15r: Let's make it more universal and not couple w/ Slack app so closely if possible.
```suggestion
        db_table = "ai_chunks"
```

backend/apps/ai/models/chunk.py
```diff
@@ -0,0 +1,70 @@
...
+        verbose_name = "Chunks"
+        unique_together = ("message", "chunk_text")
+
+    message = models.ForeignKey(Message, on_delete=models.CASCADE, related_name="chunks")
+    chunk_text = models.TextField(verbose_name="Chunk Text")
+    embedding = VectorField(verbose_name="Chunk Embedding", dimensions=1536)
```
arkid15r: For this -- do we need all these fields/relations here -- message, text (obviously we need the embedding :) )

If we have message FK we can always get its text w/o saving it here. But I suspect I'm missing something important here.

backend/apps/ai/management/commands/slack_create_chunks.py
```diff
@@ -0,0 +1,117 @@
...
+
+            processed_count += len(batch_messages)
+
+        self.stdout.write(f"Completed processing all {total_messages} messages")
+
+    def create_chunks_from_message(
+        self, message: Message, cleaned_text: str
+    ) -> list[Chunk | None]:
+        """Create chunks from a message."""
+        if message.raw_data.get("subtype") in ["channel_join", "channel_leave"]:
```
arkid15r: Create `Message::subtype` property.

backend/apps/ai/management/commands/slack_create_chunks.py
```diff
@@ -0,0 +1,117 @@
...
+                f"No chunks created for message {message.slack_message_id} - text too short"
+            )
+            return []
+
+        try:
+            time_since_last_request = datetime.now(UTC) - getattr(
+                self, "last_request_time", datetime.now(UTC) - timedelta(seconds=2)
+            )
+
+            if time_since_last_request < timedelta(seconds=1.2):
```
arkid15r: Magic number.

backend/apps/ai/management/commands/slack_create_chunks.py
```diff
@@ -0,0 +1,117 @@
...
+        """Split message text into chunks."""
+        splitter = RecursiveCharacterTextSplitter(
+            chunk_size=300,
+            chunk_overlap=40,
+            length_function=len,
+            separators=["\n\n", "\n", " ", ""],
+        )
+        return splitter.split_text(message_text)
+
+    def clean_message_text(self, message_text: str) -> str:
```
arkid15r: Should it be part of `Chunk` model?

backend/apps/ai/management/commands/slack_create_chunks.py
```diff
@@ -0,0 +1,117 @@
...
+            if batch_chunks:
+                Chunk.bulk_save(batch_chunks)
+
+            processed_count += len(batch_messages)
+
+        self.stdout.write(f"Completed processing all {total_messages} messages")
+
+    def create_chunks_from_message(
+        self, message: Message, cleaned_text: str
+    ) -> list[Chunk | None]:
```
arkid15r: Do you expect `None` in the list?

backend/apps/ai/models/chunk.py
arkid15r: Maybe this should be named `ai_embeddings`?

backend/apps/slack/management/commands/slack_sync_messages.py
Dishant1804: messages were not syncing if it is true I tried it locally, I need to figure out why it is happening

backend/apps/ai/models/chunk.py
Dishant1804: `ai_chunks` is fine

backend/apps/ai/models/chunk.py
Dishant1804: embeddings will match and the relevant chunks will be retrieved, we also need the reference to the origin of the chunks (which is message), we will fetch the message and give it to model for good response generation

backend/apps/ai/management/commands/slack_create_chunks.py
Dishant1804: we dont expect but it was needed to match the types with `update_data` method in `chunk` model file
I have fixed it

backend/apps/ai/management/commands/slack_create_chunks.py
Dishant1804: since we were just storing all the messages in raw format in `message` model, I thought that we will clean it while chunking -- I think not an issue even if it is a part of `create_chunks` command as we are processing the message for chunks :)

backend/apps/ai/management/commands/ai_create_slack_message_chunks.py
```diff
@@ -0,0 +1,92 @@
...
+            )
+
+        self.stdout.write(f"Completed processing all {total_messages} messages")
+
+    def create_chunks(self, message: Message) -> list[Chunk]:
+        """Create chunks from a message."""
+        if message.subtype in {"channel_join", "channel_leave"}:
+            return []
+
+        if not (chunk_text := Chunk.split_text(message.cleaned_text)):
```
arkid15r, line 49: Should we skip short messages too? I saw embedding generated for simple `Yes` replies.

Dishant1804: If we try to skipp short messages then the chances are we might miss some important messages as well, for eg. if the reply to the question "please provide me slack channel name for cryptography" and some user replies "xyz" so we can get one word replies

Dishant1804:
> @Dishant1804 please check whether code works and my comment below.

Yes, the code works properly we can proceed with it

PR #1667, opened 2025-06-25
Add health score as a ranking factor.

### LINKED ISSUE #1612: Use project health score as ranking factor
When we're done w/ project health score implementation and tuning we want to use this value as a ranking factor for project index results. Probably, it'll be the next one after project level.

- [ ] Yes
- [x] No

ahmedxgouda:
Resolves #1612

I had to move the `ProjectLevel` and `ProjectType` outside the `Project` model to avoid circular imports when importing `ProjectHealthMetrics` in the `Project` model. This is because the `Project` model already imported in the `ProjectHealthRequirements` model which is imported in the `ProjectHealthMetrics` model


backend/apps/owasp/models/project.py
```diff
@@ -183,6 +165,17 @@ def nest_url(self) -> str:
...
         return get_absolute_url(f"projects/{self.nest_key}")
 
+    @property
+    def last_health_metrics(self) -> ProjectHealthMetrics | None:
+        """Return last health metrics for the project."""
+        return (
+            ProjectHealthMetrics.objects.filter(project=self)
+            .order_by(
+                "-nest_created_at",
+            )
```
aramattamara: Please, make this one line.

aramattamara (approved):
LGTM! With one minor comment.

backend/apps/owasp/models/project.py
ahmedxgouda: Thanks! I applied this change. Also, I didn't dismiss your review, GitHub bot did that after pushing my change @aramattamara.

aramattamara: Thank you! That's ok, no worries.

arkid15r (approved):

PR #1668, opened 2025-06-26
Add a feature flag for showing Project Health Metrics

kasya:
Added a flag to toggle Project Health metrics visibility on a project's page. 
This will remain in place while we finalize how to handle score and metric weighting per project level.


arkid15r (approved):
LGTM

PR #1686, opened 2025-07-01
Add auto-scroll to top on pathname change

kasya:
Added an `AutoScrollToTop` component to `RootLayout` to have pages scroll to top on route change. 

> This is a known behavior in Next.js (React) apps due to the way client-side routing works. When you navigate between pages using next/link or useRouter().push(), Next.js performs a soft navigation—it loads the new page content without refreshing the browser or resetting the scroll position unless explicitly told to do so.


arkid15r (approved):
LGTM

PR #1614, opened 2025-06-15
Enhance System Observability: Introduce Backend Health & Version Status Endpoint

rishyym0927:
#### Context

Fixes #1517 

This PR implements a dedicated **health and version status endpoint** at `/apps/core/api/status.py` to improve the backend’s observability and integration capabilities. The new `status_view` function returns a lightweight JSON response indicating the backend’s operational status (`"ok"`) and includes the current application release version, dynamically sourced from Django's `RELEASE_VERSION` setting.

This enhancement serves multiple critical use cases:

- ✅ **Health Monitoring**: Enables external systems (e.g., uptime monitoring tools, CI/CD pipelines) to programmatically verify backend availability.
- 🔄 **Deployment Validation**: Confirms version consistency across services during and after deployments.
- 🌐 **Frontend-Backend Synchronization**: Ensures the frontend always displays the accurate release version in the footer by rendering `NEXT_PUBLIC_RELEASE_VERSION` at runtime, eliminating discrepancies.

The endpoint follows standard RESTful practices using Django’s `JsonResponse`, ensuring minimal overhead and broad compatibility with status check tools.

image

PR #1713, opened 2025-07-04
Add `/status` endpoint to check backend version, show version in frontend footer

rudransh-shrivastava:
Resolves #1614

added `/status` endpoint that returns  `{ "version": "1.x.x }`.
added version in the footer.
 
I tried writing tests for the frontend, but I was unable to inject `NEXT_PUBLIC_RELEASE_VERSION` environment variable in a test.


rudransh-shrivastava:
> **Actionable comments posted: 1**
> ♻️ Duplicate comments (2)
> 
> > backend/tests/apps/core/api/status_test.py (2)> `14-14`: **Past review comment about missing class docstring is outdated.**
> > > The class docstring is present on line 15, so the previous review comment suggesting to add one is no longer applicable.
> > > 
> > > `32-42`: **Simplify the test approach and fix quote consistency.**
> > > The current approach of patching the entire settings module and deleting attributes is unnecessarily complex. The `getattr` function with a default value will naturally return "unknown" when the attribute doesn't exist.
> > > Apply this diff to simplify the test and fix the quote consistency:
> > > ```diff
> > >  def test_get_status_returns_unknown_when_no_version(self):
> > >      """Test get_status function returns 'unknown' when RELEASE_VERSION is not set."""
> > >      request = self.factory.get("/")
> > > 
> > > -    with patch('apps.core.api.status.settings') as mock_settings:
> > > -        del mock_settings.RELEASE_VERSION
> > > +    with patch.object(settings, "RELEASE_VERSION", side_effect=AttributeError):
> > >          response = get_status(request)
> > > 
> > >      assert response.status_code == HTTPStatus.OK
> > >      data = json.loads(response.content)
> > >      assert data["version"] == "unknown"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This approach:
> > > 
> > > 1. Uses consistent double quotes (addresses static analysis hint)
> > > 2. More accurately simulates the `AttributeError` that would occur when `RELEASE_VERSION` doesn't exist
> > > 3. Maintains consistency with the first test's patching approach
> > > 4. Is simpler and more readable
> 
> 📜 Review details

this coderabbit comment about adding a `AttributeError` `side_effect` does not seem to work as `MagicMock` is not JSON Serializable. Also, I don't think the `getattr` function throws an error as coderabbit suggests, it just defaults to `unknown`. There are other complicated ways to test this particular case but I found my current approach to be the simplest. Please let me know if there is a better way I would be happy to implement that.

rudransh-shrivastava:
I also had a question, should the `/status` endpoint also return an `ok` field?

PR #1718, opened 2025-07-07
Implement the presentation of dashboard metrics list with filters

### LINKED ISSUE #1701: Implement the presentation of the project health metrics dashboard list
Implement the presentation of the last health metrics for the projects with a summary for each project. This should be an overview only. The detailed metrics will be implemented in another PR.

- [x] Yes
- [ ] No

ahmedxgouda:
Resolves #1701 

https://github.com/user-attachments/assets/35ef0134-798f-4503-9c4c-52abb201605e


frontend/src/app/projects/dashboard/metrics/page.tsx
```diff
@@ -0,0 +1,144 @@
...
+              <DropdownItem key="incubator">Incubator</DropdownItem>
+              <DropdownItem key="lab">Lab</DropdownItem>
+              <DropdownItem key="production">Production</DropdownItem>
+              <DropdownItem key="flagship">Flagship</DropdownItem>
+            </DropdownMenu>
+          </Dropdown>
+          <Select
+            label="Sort By"
+            placeholder="Select sorting option"
+            selectionMode="multiple"
```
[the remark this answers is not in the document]
ahmedxgouda: It is not a bug. This was made intentionally. User can select multiple options

backend/apps/owasp/graphql/queries/project_health_metrics.py
```diff
@@ -6,17 +6,38 @@
...
+        Args:
+            filters (ProjectHealthMetricsFilter): Filters to apply on the metrics.
+            pagination (strawberry_django.pagination.OffsetPaginationInput): Pagination parameters.
+            ordering (list[ProjectHealthMetricsOrder], optional): Ordering parameters.
+
+        Returns:
+            list[ProjectHealthMetricsNode]: List of project health metrics.
+
+        """
+        return ProjectHealthMetrics.get_latest_health_metrics()
```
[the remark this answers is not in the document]
ahmedxgouda: This is not true. It is applied already with the current logic

frontend/src/components/MetricsCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <p className="text-md">{metric.starsCount}</p>
+        </div>
+        <div className="truncate">
+          <p className="text-md">{metric.forksCount}</p>
+        </div>
+        <div className="truncate">
+          <p className="text-md">{metric.contributorsCount}</p>
+        </div>
+        <div className="truncate">
+          <p className="text-md text-center">{new Date(metric.createdAt).toLocaleDateString()}</p>
```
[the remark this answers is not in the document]
ahmedxgouda: The createdAt is auto created in the backend, so it will not be null at all

frontend/src/app/projects/dashboard/metrics/page.tsx
```diff
@@ -0,0 +1,283 @@
...
+            }}
+          />
+        </div>
+      </div>
+      <div className="grid grid-cols-[4fr_1fr_1fr_1fr_1.5fr_1fr] p-4">
+        <div className="truncate font-semibold">Project Name</div>
+        <div className="truncate font-semibold">Stars</div>
+        <div className="truncate font-semibold">Forks</div>
+        <div className="truncate font-semibold">Contributors</div>
+        <div className="truncate text-center font-semibold">Created At</div>
```
kasya: This `Created at` column currently shows created_at for the metric itself, but the rest of the numbers we show are about the project, not metric. This makes it a bit confusing....  we should either show project creation with `Created at` label, or change the wording to specify that this is the date when the metric has been calculated.

ahmedxgouda:
Hey @kasya, I've uploaded a demo video.

frontend/src/app/projects/dashboard/metrics/page.tsx
```diff
@@ -0,0 +1,283 @@
...
+                incubator: '?level=incubator',
+                lab: '?level=lab',
+                production: '?level=production',
+                flagship: '?level=flagship',
+              }
+              router.push(`/projects/dashboard/metrics${paramsMapping[key] || ''}`)
+            }}
+          />
+
+          <ProjectsDashboardDropDown
```
kasya, line 194: We have a `SortBy` component already that is pretty customizable. Do you think we could use it here for style consistency? 🤔 
Screenshot 2025-07-13 at 4 13 31 PM

frontend/src/app/projects/dashboard/metrics/page.tsx
kasya, line 1: Can we also update color for nav items to be the same color as the breadcrumbs for style consistency?
Screenshot 2025-07-13 at 4 17 41 PM

kasya, line 1: Another idea about the dropdown:
what is we split it into several dropdowns? With current approach you can never filter by level + health status (e.g. flagship + unhealthy).

We could do 3 dropdowns, spread them on one line at the top of the table (maybe keeping two closer to the left side and keep sorting to the right), and they would be: Level, health status and sorting. 
Each of those two to the left (level and health) should have an option that will cover all (e.g. no filter). This way we don't need separate `reset all filters` option/button and users will be able to filter more granularly.
It would still be good to somehow reuse`SortBy` component if possible. 👍🏼 

Also, as for sorting, I'm not sure if we need all those options. Maybe let's only keep score for now?

backend/apps/owasp/graphql/queries/project_health_metrics.py
```diff
@@ -27,3 +48,13 @@ def project_health_stats(self) -> ProjectHealthStatsNode:
...
+
+    @strawberry.field
+    def project_health_metrics_distinct_length(self) -> int:
+        """Get the distinct length of project health metrics.
+
+        Returns:
+            int: The count of distinct project health metrics.
+
+        """
+        return ProjectHealthMetrics.get_latest_health_metrics().count()
```
kasya: I've left a message in Slack regarding this, but I think you should be able to address your issue with pagination result here if you pass filter to this method too, same way you do for `project_health_metrics()`.
Of course, you'd also need to update the GraphQL query to pass it filter params.
This should then count only filtered metrics and match what we return on `refetch` with filters.

kasya (commented):
Nice update! Left a few requests:

frontend/src/app/projects/dashboard/metrics/page.tsx
ahmedxgouda: We could use a multiple selection in the dropdown. I chose to create a general DropDown that will be usable in other cases that uses heroui DropDown instead of using `SortBy` component, because it was specific to its case.

backend/apps/owasp/graphql/queries/project_health_metrics.py
ahmedxgouda: I tried that, and the filters didn't work in the backend, just returning the length of all metrics

frontend/src/app/projects/dashboard/metrics/page.tsx
kasya: @ahmedxgouda oh, I didn't realize that when testing 🤔 I can see now that we can select multiple options. Can we at least make the design somewhat similar to that `sortBy` component? Like, darker shade. Ideally, I'd like it to show currently selected items, because it's not obvious any filter is applied unless you click into the dropdown 👀

frontend/src/app/projects/dashboard/metrics/page.tsx
```diff
@@ -178,30 +183,35 @@ const MetricsPage: FC = () => {
...
             sections={filteringSections}
-            selectionMode="single"
-            selectedKeys={[activeFilterKey]}
-            onAction={async (key: string) => {
+            selectionMode="multiple"
+            selectedKeys={activeFilters}
+            onAction={(key: string) => {
               // Because how apollo caches pagination, we need to reset the pagination.
               const newPagination = { offset: 0, limit: PAGINATION_LIMIT }
               setPagination(newPagination)
```
kasya: We can just do it inline - no need to a new variable here:
```suggestion
              setPagination({ offset: 0, limit: PAGINATION_LIMIT })
```

frontend/src/app/projects/dashboard/metrics/page.tsx
```diff
@@ -0,0 +1,296 @@
...
+        { label: 'Lab', key: 'lab' },
+        { label: 'Production', key: 'production' },
+        { label: 'Flagship', key: 'flagship' },
+      ],
+    },
+    {
+      title: 'Project Health',
+      items: [
+        { label: 'Healthy Projects', key: 'healthy' },
+        { label: 'Projects Needing Attention', key: 'needsAttention' },
```
kasya: Can we show labels instead of keys for Active filters? And for that to look ok on the component we'll probably need to make these shorter. What if we also do something like `Healthy`, `Need attention` and `Unhealthy`? 

Screenshot 2025-07-15 at 5 45 42 PM

frontend/src/app/projects/dashboard/metrics/page.tsx
```diff
@@ -0,0 +1,296 @@
...
+                    .map(([, value]) => value)
+                )
+              )
+              router.replace(`/projects/dashboard/metrics?${newParams.toString()}`)
+            }}
+          />
+
+          <ProjectsDashboardDropDown
+            buttonDisplayName="Sort By"
+            icon={faSort}
```
kasya: Can we also update this to just say `score` for active order and change icon based on the state (asc/desc)? 
I was thinking we could use [this one for descending](https://fontawesome.com/icons/arrow-down-wide-short?f=classic&s=solid) and [this one for ascending](https://fontawesome.com/icons/arrow-up-wide-short?f=classic&s=solid)

frontend/src/app/projects/dashboard/metrics/page.tsx
```diff
@@ -0,0 +1,296 @@
...
+  const getCurrentPage = () => {
+    return Math.floor(pagination.offset / PAGINATION_LIMIT) + 1
+  }
+
+  return (
+    <>
+      <div className="mb-4 flex items-center justify-between">
+        <h1 className="text-2xl font-bold">Project Health Metrics</h1>
+        <div className="flex items-center space-x-2">
+          <ProjectsDashboardDropDown
```
kasya, line 158: Also, just an idea and I know we've been through this already, but would like to reiterate still.
Whaat do you think of this - what if we split these into 2? I can see how this could be cleaner look if we have 2 separate dropdowns for Level and Health. I hardcoded some things to show how that might look like:
Screenshot 2025-07-15 at 6 03 56 PM
Would that be too much trouble to implement it this way? 🤔

kasya (changes requested):
Looks great! Left a couple more comments, but we are close! 🔥

frontend/src/app/projects/dashboard/metrics/page.tsx
kasya: See my other comment about these filters ⬇️

frontend/src/app/projects/dashboard/metrics/page.tsx
ahmedxgouda: It can be implemented, but we will need more work to change the logic of `onAction`. Although I think having two buttons for filtering is too much, and not scalable. The filtering could be updated with new filters and specific ones like `is_funding_compliant` and `is_leader_requirement_compliant`. So having a drop down for each will be too overkill.

ahmedxgouda:
Ok the failing test is from the main branch.

frontend/src/app/projects/dashboard/metrics/page.tsx
kasya: @ahmedxgouda Good point 👌🏼
That said, I do see one potential issue. If we add more filters later on, and all of them support multiple selections, this dropdown could end up getting quite wide to fit all the selected options.
Screenshot 2025-07-16 at 7 24 56 PM

For now, I think it’s fine to leave it as is — though it’s already stretching a bit when the longest options are selected. But if we introduce more filters in the future, we’ll probably want to rethink how we display the selected items, especially if we add something like `is_funding_compliant` and `is_leader_requirement_compliant`

frontend/src/types/healthMetrics.ts
```diff
@@ -51,18 +51,3 @@ export type HealthMetricsFilter = {
   }
   level?: string
 }
-
```
kasya: Was about to approve, but noticed that pagination doesn't work with sorting after these most recent changes :( 
Here's a screenshot of the error:
Screenshot 2025-07-16 at 7 45 43 PM

ahmedxgouda: Yes this is related to the refactor. I faced this bug before.

ahmedxgouda: I think it should work now

kasya (approved):
Looks good! 👍🏼

ahmedxgouda:
I was resolving conflicts, of course I would not dismiss your approval :) @kasya

arkid15r:
> I was resolving conflicts, of course I would not dismiss your approval :) @kasya

It seems you overthink it a bit, the workflow is fine and all understand whats and whys, don't sweat it @ahmedxgouda it's all good 👍

PR #1729, opened 2025-07-10
Improve chapters/project leaders presentation

### LINKED ISSUE #1431: Impove chapter/project leaders presentation
We show OWASP Nest leaders at the /about page in a pretty nice way.

Extend this look to leaders on all chapters/projects pages. The idea is to extract [this code](https://github.com/OWASP/Nest/blob/bacb8084b0d82e99f7187ae9c2665bf4fb6330e0/frontend/src/app/about/page.tsx#L193) into a separate components and reuse it on the about page (keep the layout w/o changes) and add `Leaders` block on chapters/projects details pages right above `Top Contributors` section.

- [ ] Yes
- [x] No

**Additional context**

Image

Image

rudransh-shrivastava:
Resolves #1431 

Hello, I need help @arkid15r ,

here's what I did in this PR: 
- exposed leaders (m2m relationship with github.User) as `leaders_temp` field
- extracted the `Leaders` Block in a component
- used it in about page and projects page (chapters left out for now)

questions/blockers:
1. I had to use `leaders_temp` as i could not come up with a different name. There's already a `leaders` field exposed but it's misleading as its actually `leaders_raw` in the database which is just `string[]`. I'm not sure what to name it. I think renaming the existing `leaders` to `leaders_raw` or `leaders_names` would be a good approach.
2. The original `Leaders` block requires data in the following format:
```ts
const leaders = {
  arkid15r: 'CCSP, CISSP, CSSLP',
  kasya: 'CC',
  mamicidal: 'CISSP',
}
```
however, I'm not sure how to get the `CCSP, CISSP, CSSLP` part from the database. 
3. since `leaders_temp` is of type `User[]` we need to convert it in the above format (type `LeadersListBlockProps`). in this implementation, I have just reduced it using code. But I'm not sure if this is a good approach, please let me know.

Please let me know what you think, I will update the code and the related tests accordingly.

### DIFF
--- a/backend/apps/owasp/graphql/nodes/common.py
+++ b/backend/apps/owasp/graphql/nodes/common.py
@@ -3,11 +3,17 @@
 import strawberry
 
 from apps.github.graphql.nodes.repository_contributor import RepositoryContributorNode
+from apps.github.graphql.nodes.user import UserNode
 
 
 @strawberry.type
 class GenericEntityNode:
     """Base node class for OWASP entities with common fields and resolvers."""
+
+    @strawberry.field
+    def leaders_temp(self) -> list[UserNode]:
+        """Resolve leaders logins."""
+        return self.leaders.all()
 
     @strawberry.field
     def leaders(self) -> list[str]:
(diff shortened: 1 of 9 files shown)


rudransh-shrivastava:
will resolve cr comments.

PR #1732, opened 2025-07-11
Introduce DisableIndexing context manager

arkid15r:
Refactor code to use `index.disable_indexing()` context manager.


kasya (approved):
LGTM!

PR #1746, opened 2025-07-16
Api features filtering, caching and ordering 

### LINKED ISSUE #1619: Add API response cache mechanism
For API responses that look same for all we want to have them cached instead of fetched from DB for each request.
Research options available for integrating existing solutions w/ Django Ninja or implement it from scratch if nothing fits our needs.

Aditionally,
- Add filtering, pagingation

- [ ] Yes
- [x] No

abhayymishraa:
Resolves #1619 

This is the GSoC API Developement Djanog ninja part.
In this Pull request i have covered:
- [x] Filtering 
- [X] caching
- [X] ordering

yet to be covered (After discussion)
text_search

Add the PR description here.


backend/apps/github/api/v1/repository.py
```diff
@@ -22,10 +31,13 @@ class RepositorySchema(Schema):
...
+@paginate(PageNumberPagination, page_size=PAGE_SIZE)
+def list_repository(
+    request: HttpRequest, filters: RepositoryFilterSchema = Query(...)
+) -> list[RepositorySchema] | dict:
     """Get all repositories."""
-    repositories = Repository.objects.all()
+    repositories = filters.filter(Repository.objects.all())
     if not repositories.exists():
         raise HttpError(404, "Repositories not found")
     return repositories
```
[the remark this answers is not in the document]
arkid15r: @abhayymishraa that's something that I suggest too -- return empty list instead of 404 for `list_*` methods.

abhayymishraa:
@coderabbitai can you do a full review of this pr look any enhancement we can do or is there any vulnerability ?

abhayymishraa:
I think all the improvement from the coderabbit feels irrelevant to me.
@arkid15r  any of them seems to you for worth giving a try.

backend/apps/common/constants.py
```diff
@@ -6,3 +6,5 @@
 OWASP_NEWS_URL = "https://owasp.org/news"
 OWASP_WEBSITE_URL = "https://owasp.org"
 TAB = "\t"
+PAGE_SIZE = 100
```
arkid15r: Is it the best place for the new items in this file?

backend/apps/github/api/v1/issue.py
```diff
@@ -23,11 +31,21 @@ class IssueSchema(Schema):
     url: str
 
 
-@router.get("/", response={200: list[IssueSchema], 404: dict})
-@paginate(PageNumberPagination, page_size=100)
-def list_issues(request: HttpRequest) -> list[IssueSchema] | dict:
+VALID_ISSUE_ORDERING_FIELDS = ["created_at", "updated_at"]
```
arkid15r: Make it a set

backend/pyproject.toml
```diff
@@ -147,7 +147,7 @@ select = ["ALL"]
 [tool.ruff.lint.per-file-ignores]
 "**/__init__.py" = ["D104", "F401"]
 "**/admin.py" = ["D100", "D101", "D104"]
-"**/api/*.py" = ["ARG001", "D106"]
+"**/api/*.py" = ["ARG001", "D106", "B008"]
```
arkid15r: Why do we need this?

abhayymishraa: Actually we are getting linting error about 
`B008 Do not perform function call `Query` in argument defaults; instead, perform the call within the function, or read the default from a module-level singleton variable`
But as the mentioned in the document they tell us to use this Query exactly how i used it. As i tried to find another method but seems like we need to suppress it for api/**
https://django-ninja.dev/guides/input/filtering/

arkid15r: I'd start from the question whether it makes sense to follow that recommendation. Why do we need the default value as `Query(...)` or similar?

The `ruff` recommendation is the way to go `Do not perform function call `Query` in argument defaults; instead, perform the call within the function` and I already suggested previously do not add new suppression instead of fixing it. 

I believe both filters/ordering are nullable and can have `None` as default?

abhayymishraa: hey @arkid15r  basically i used the documentation of the Django ninja, they do it like this so i followed the same. i will give it a try and update you soon.

arkid15r (approved):

PR #1748, opened 2025-07-17
Implement the detailed metrics dashboard page

### LINKED ISSUE #1702: Implement the presentation of the project health metrics details
After implementing the health stats and the summary for each project, we should implement a detailed page to represent all needed metrics. We could include charts in this one.

- [x] Yes
- [ ] No

ahmedxgouda:
Resolves #1702

https://github.com/user-attachments/assets/69a5ed1c-3878-4696-badc-4d24c9f34465


frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
```diff
@@ -0,0 +1,153 @@
...
+          </div>
+          <BarChart
+            title="Days Metrics"
+            icon={faChartArea}
+            labels={[
+              'Project Age',
+              'Days Since Last Commit',
+              'Days Since Last Release',
+              'Days Since Last Pull Request',
+              'Days Since OWASP Page Last Update',
```
Comment (kasya) [inline frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx:190]: Can we add a tooltip to this to show the full text when it gets truncated? 
Screenshot 2025-07-18 at 5 16 41 PM

frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
```diff
@@ -0,0 +1,153 @@
...
+              stats={millify(metrics.unassignedIssuesCount)}
+            />
+            <DashboardCard
+              title="Unanswered Issues"
+              icon={faQuestionCircle}
+              stats={millify(metrics.unansweredIssuesCount)}
+            />
+          </div>
+          <div className="grid grid-cols-1 gap-4 md:grid-cols-3">
+            <DashboardCard
```
Comment (kasya) [inline frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx]: Can we show these as charts? Basically this will look very much as a Health metrics part on the project page, only this will be an internal view. 
But I feel like we need to see the pattern/change for each, not just the current number. 🤔

frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
```diff
@@ -0,0 +1,153 @@
...
+  return (
+    <div className="flex flex-col gap-4">
+      {metrics && (
+        <>
+          <div className="flex items-center justify-between">
+            <h1 className="text-2xl font-bold">{metrics.projectName}</h1>
+            <MetricsScoreCircle score={metrics.score} />
+          </div>
+          <div className="grid grid-cols-1 gap-4 md:grid-cols-2">
+            <GeneralCompliantComponent
```
Comment (kasya) [inline frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx]: What if we show these as "badges" instead? 
Something like this FA icon https://fontawesome.com/icons/certificate?f=classic&s=solid and then the icons you chose on top of it. Then we'd show the same tooltip you've added. We'll change their color based on True/False state of course.

I was thinking maybe we could show them at the top right side, next to health score circle 🤔 
Screenshot 2025-07-18 at 5 27 24 PM

what do you think?

kasya (commented):
Some initial thoughts 👍🏼 ⬇️

frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
Comment (ahmedxgouda) [reply frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx:190]: This text is a part of the apexcharts Chart component. So, I don't have full control on it. Maybe it can be done, I will take a look.

frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
Comment (ahmedxgouda) [reply frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx]: Since they are just counters, the charts we used so far will not be appropriate. I don't know, but maybe apexcharts provide a chart for our case here. I chose to display the latest health metrics, so the page become consistent with list page as it contains the latest health metrics of all projects.

frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
Comment (ahmedxgouda) [reply frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx]: It seems to be a good idea. It will be better I think.

Comment (ahmedxgouda) [reply frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx]: Since we have this button in the new PR. Is it still a good idea to make those next to the health score circle?
image

Comment (kasya) [reply frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx]: @ahmedxgouda good point, I haven't checked that PR yet, sorry!

What if we have not a button but just an Icon next to the name (something like [this](https://fontawesome.com/icons/file-arrow-down?f=classic&s=solid) )? I'm not sure I like the button as it is right now.. Right now it takes the most prominent spot, but I don't feel like it should. It's an additional feature, but not as important. 
So I'm thinking maybe the icon next to the title (grey colored with tooltip and hover) will be better... Than we can do those requirement icons to the right as we wanted. 
Would love to hear your thoughts!

frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
Comment (kasya) [reply frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx]: @ahmedxgouda oh, with data I have locally I didn't realize this was a single metric view. I think that would make more sense to show the change.. like what we show on the project page right now.
I'm not 100% sure how this will be used internally in OWASP, but I'd assume seeing the progress for a project is important. For example, the project was falling behind on some requirements, but improving the numbers. Showing those as a single number won't tell us if there's a pattern 🤔

frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
Comment (kasya) [reply frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx:190]: @ahmedxgouda Gotcha! I'll check that too! Thank you!

frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
Comment (ahmedxgouda) [reply frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx]: No worries. I liked this idea. Thank you!

frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx
Comment (ahmedxgouda) [reply frontend/src/app/projects/dashboard/metrics/[projectKey]/page.tsx:190]: @kasya Didn't find a way, but a tooltip will show when the user hover on the bar itself.

kasya (approved):
@ahmedxgouda I’ve pushed some updates to the Compliance badges styling. Hope that's ok with you.

This looks good for now as an MVP! 
We’ll likely revisit this page later once I figure out a clearer way to highlight any missing requirements. The goal is to make it obvious to project admins which requirements are still unmet and how they impact the score, so they know what needs their attention.

Right now, I feel like only our Age BarChart communicates that somewhat clearly. I’ll need to spend a bit more time thinking about the best design approach for this.

If you have any ideas in the meantime, feel free to share too!

PR #1758, opened 2025-07-19
Refactor Slack application code

arkid15r:

Slack application code refactoring.


backend/apps/slack/commands/events.py
```diff
@@ -1,29 +1,42 @@
...
-from apps.common.constants import OWASP_WEBSITE_URL
+from apps.common.constants import OWASP_URL
 from apps.slack.commands.command import CommandBase
-from apps.slack.utils import get_events_data
+
+
+def get_events_data():
+    """Get events data for the template."""
+    from apps.owasp.models.event import Event
+
```
kasya:
```suggestion
from apps.owasp.models.event import Event

def get_events_data():
    """Get events data for the template."""

```

backend/apps/slack/events/event.py
```diff
@@ -35,25 +33,113 @@ def configure_events():
...
+
+        """
+        return (
+            env.get_template(str(self.ephemeral_message_template_path))
+            if self.ephemeral_message_template_path
+            else None
+        )
+
+    @property
+    def ephemeral_message_template_path(self) -> Path | None:
```
kasya, line 83: This always returns `None`. Is that expected behavior? 🤔

backend/apps/slack/commands/events.py
arkid15r: Added a comment.

backend/apps/slack/events/event.py
arkid15r: Added a comment.

kasya (approved):

PR #1823, opened 2025-07-23
Implement Dashboard Auth logic

### LINKED ISSUE #1791: Implement the internal dashboard authentication logic
We should add `is_owasp_staff` field to the GitHub user model and guard endpoints based on that.

- [x] Yes
- [ ] No

ahmedxgouda:
## Proposed change

Resolves #1791

https://github.com/user-attachments/assets/f79f1b62-c208-4dd0-b66f-1c886ebeb826

## Checklist


ahmedxgouda:
@coderabbitai Please look for any security vulnerabilities or improvements in these changes

ahmedxgouda:
@coderabbitai Since we add a new GraphQL query that returns the user `is_owasp_staff` field, and based on this query we render the dashboard pages if this field is true and direct the user to 404 page if the field is false, suggest approaches to handle the new logic in dashboard pages e2e tests to make sure that the pages are rendered and do not navigate to 404 page.

ahmedxgouda:
@coderabbitai the django user auth model of the view request object in backend/apps/owasp/api/internal/views/project_health_metrics.py return anonymous user for authenticated users. We use django sessions, and in this PR we have strawberry info.context.user which returns the authenticated user. Could you suggest a way to access the user that strawberry uses in its context in django views?

ahmedxgouda:
@coderabbitai The solution you provided didn't work. The request.session object does not have any keys or values in this view.

frontend/__tests__/e2e/pages/Login.spec.ts
```diff
@@ -22,10 +22,14 @@ test.describe('Login Page', () => {
...
 
-    const disabledMessage = page.getByText(/with GitHub is not enabled/i)
-    await expect(disabledMessage).toBeVisible()
+    const welcomeMessage = page.getByText('Welcome back')
+    const gitHubLoginButton = page.getByRole('button', { name: 'Sign In with GitHub' })
+    const gitHubMessage = page.getByText('Sign in with your GitHub account to continue')
+    await expect(welcomeMessage).toBeVisible()
+    await expect(gitHubLoginButton).toBeVisible()
+    await expect(gitHubMessage).toBeVisible()
   })
```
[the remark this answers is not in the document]
ahmedxgouda: There will be no difference in the page whether the user is OWASP Staff or not

backend/apps/owasp/api/internal/permissions/project_health_metrics.py
```diff
@@ -0,0 +1,19 @@
...
+    """Permission class to check if the user is an OWASP staff member."""
+
+    message = "You must be an OWASP staff member to access this resource."
+
+    def has_permission(self, source, info, **kwargs) -> bool:
+        """Check if the user is an OWASP staff member."""
+        user = info.context.request.user
+        return (
+            user
+            and hasattr(user, "github_user")
```
arkid15r: I'm not sure -- do we really need the `hasattr()` check here?
For anonymous ones?

backend/apps/owasp/api/internal/permissions/project_health_metrics.py
```diff
@@ -0,0 +1,19 @@
+"""Strawberry Permission Classes for Project Health Metrics."""
+
+from strawberry.permission import BasePermission
+
+
+class IsOWASPStaff(BasePermission):
```
arkid15r:
```suggestion
class IsOwaspStaff(BasePermission):
```

backend/apps/owasp/api/internal/views/permissions.py
```diff
@@ -0,0 +1,41 @@
...
+from django.contrib.auth import get_user
+from django.http import HttpResponseForbidden
+
+
+def has_owasp_staff_permission(request):
+    """Check if user is an OWASP staff member."""
+    # Returns Anonymous user even if authenticated
+    # Strawberry returns the authenticated user
+    user = get_user(request)
+    return bool(
```
arkid15r: Do you need `bool()` here?

backend/apps/owasp/api/internal/views/permissions.py
```diff
@@ -0,0 +1,41 @@
...
+        user
+        and hasattr(user, "github_user")
+        and user.github_user
+        and user.github_user.is_owasp_staff
+    )
+
+
+def owasp_staff_required(view_func):
```
arkid15r: This naming approach limits us to the existing logic. When we are ready to extend it for a broader audience, e.g. project leaders, this name becomes obsolete.

backend/apps/owasp/api/internal/permissions/project_health_metrics.py
arkid15r: We decided to start from staff users. However, in future we highly likely to extend it for users who need the access but do not actually work for OWASP. The implementation should be flexible enough to accommodate those changes when time comes.

frontend/src/hooks/useDjangoSession.ts
```diff
@@ -37,6 +39,7 @@ export const useDjangoSession = () => {
           const githubAuth = response?.data?.githubAuth
           if (githubAuth?.ok) {
             sessionStorage.setItem(SYNC_STATUS_KEY, 'true')
+            sessionStorage.setItem(IS_OWASP_STAFF_KEY, String(githubAuth.user?.isOwaspStaff))
```
arkid15r: I believe we should use Next.js mechanism for setting user related attributes and access them later instead of `sessionStorage`.

frontend/src/server/queries/authQueries.ts
```diff
@@ -15,6 +15,9 @@ export const SYNC_DJANGO_SESSION_MUTATION = gql`
     githubAuth(accessToken: $accessToken) {
       message
       ok
+      user {
+        isOwaspStaff
```
arkid15r, line 19: We'll probably want all attrs the BE sets for user.

backend/apps/owasp/api/internal/permissions/project_health_metrics.py
ahmedxgouda: Github user is a one to one field. If a user has not a github user associated with it, it will not have the github_user attribute. This is documented in Django page here:
image

https://docs.djangoproject.com/en/5.2/topics/db/examples/one_to_one/

backend/apps/owasp/api/internal/views/permissions.py
ahmedxgouda: Not really in this case, but it is useful in some cases when we need to make sure that the expression will evaluate a boolean value.

frontend/src/server/queries/authQueries.ts
ahmedxgouda: We don't want it for now right? I mean why would we fetch attrs we don't need? Also, other attrs like login, image are fetched by GitHub already and exist in NextAuth session object.

arkid15r:
@abhayymishraa @Rajgupta36 could you review the NextAuth related part?

Thank you!

Rajgupta36 (dismissed):

backend/apps/owasp/api/internal/permissions/project_health_metrics.py
```diff
@@ -0,0 +1,19 @@
...
+    message = "You must have dashboard access to access this resource."
+
+    def has_permission(self, source, info, **kwargs) -> bool:
+        """Check if the user has dashboard access."""
+        user = info.context.request.user
+        return (
+            user
+            and hasattr(user, "github_user")
+            and user.github_user
+            and user.github_user.is_owasp_staff
```
arkid15r: Can we simplify this? I suggest adding user.is_authenticated here and rely on the fact that authenticated user will have `github_user` object.

backend/apps/owasp/api/internal/views/permissions.py
```diff
@@ -0,0 +1,37 @@
...
+
+from functools import wraps
+
+from django.contrib.auth import get_user
+from django.http import HttpResponseForbidden
+
+
+def has_dashboard_permission(request):
+    """Check if user has dashboard access."""
+    user = get_user(request)
```
arkid15r: `request.user`?

frontend/src/components/ProjectsWrapper.tsx
arkid15r, line 1: Let's replace this w/ a user menu item `Projects Health` visible to OWASP staff only.

arkid15r: Right, on the BE you changed it to more general name but here we still have OWASP staff terminology instead of something like `hasDashboardAccess`

frontend/src/components/UserMenu.tsx
```diff
@@ -84,13 +88,17 @@ export default function UserMenu({
...
+          <div className="flex flex-col p-2">
+            <UserMenuItem onClick={handleLogout} disabled={isLoggingOut}>
+              {isLoggingOut ? 'Signing out...' : 'Sign out'}
+            </UserMenuItem>
+
+            {isOwaspStaff && (
+              <Link href="/projects/dashboard">
+                <UserMenuItem onClick={() => setIsOpen(false)}>Projects Health</UserMenuItem>
+              </Link>
+            )}
```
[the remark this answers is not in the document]
ahmedxgouda: Is this we should look into? @kasya @arkid15r

arkid15r (approved):
Congrats on completing this @ahmedxgouda ! The auth part is still looks somewhat fuzzy to me and no one helped to review it... It passes my manual smoke tests though :)

I'm merging this.

PR #1824, opened 2025-07-23
Fixed issue #1807

### LINKED ISSUE #1807: Add tests for <DashboardCard> component
Write unit tests for the `<DashboardCard>` React component to ensure expected behavior, edge case handling, and rendering logic.

## Essential Test Coverage Checklist

- [ ] **Renders successfully with minimal required props**  
- [ ] **Conditional rendering logic**  
- [ ] **Prop-based behavior** – different props affect output  
- [ ] **Event handling** – simulate user actions and verify callbacks  
- [ ] **State changes / internal logic**  
- [ ] **Default values and fallbacks**  
- [ ] **Text and content rendering**  
- [ ] **Handles edge cases and invalid inputs**  
- [ ] **Accessibility roles and labels**  
- [ ] **DOM structure / classNames / styles**

## Test Reference  
You can refer to the `AutoScrollToTop.test.tsx` file for an example of structure and best practices.  
To explore more examples, see the full component tests folder.

VrajVed:
## Proposed change

Resolves #1807 

Added fallback checks,
Tested edge cases with missing values,
Tested if props pass and components render

## Checklist

### DIFF
--- a/frontend/__tests__/unit/components/ActionButton.test.tsx
+++ b/frontend/__tests__/unit/components/ActionButton.test.tsx
@@ -1,0 +1,119 @@
+import { cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react'
+import React from 'react'
+import ActionButton from 'components/ActionButton'
+
+jest.mock('@heroui/tooltip', () => ({
+  Tooltip: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+}))
+
+jest.mock('next/link', () => {
+  return ({
+    children,
+    href,
+    onClick,
+  }: {
+    children: React.ReactNode
+    href: string
+    onClick?: () => void
+  }) => (
+    <a
+      href={href}
+      onClick={(e) => {
+        e.preventDefault()
+        onClick?.()
+      }}
+    >
+      {children}
+    </a>
+  )
+})
+
+describe('ActionButton', () => {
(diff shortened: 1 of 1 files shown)


frontend/src/components/DashboardCard.tsx
arkid15r, line 1: Why did you change the component's code?

VrajVed: I did that to add fallbacks to defaults in case the values passed are missing/null, is there a different way to do it ??

arkid15r: Well, I just didn't expect any changes to the source code unless it's a bug fix.

As for the changes -- both title/icon are mandatory params but you've changed icon to optional for some reason. I suggest reverting the component code changes and keeping the tests for review. Also we don't use `testids` I guess.

arkid15r (changes requested):

VrajVed:
@arkid15r 

image


This is what happens when I keep icon as compulsory because the icon value doesn't take `null` as an argument. Also to add the defaults and fall backs, the component needs to be changed. I needed to add the `title` attribute to test for the icon 

```jsx
expect(screen.getByTitle('user-icon')).toBeInTheDocument() /*Test Passed*/
```
```jsx
<FontAwesomeIcon icon={icon} title='user-icon'/>
```

PR #1891, opened 2025-07-28
NestBot AI Assistant Contexts

### LINKED ISSUE #1873: Extend AI Application to Support NestBot AI Assistant Contexts

We want to extend the OWASP Nest AI application to introduce **NestBot AI Assistant functionality**. This will require a new model for storing **contexts**, which are generated text entries that can optionally be linked to specific OWASP entities (e.g., Projects, Chapters, Committees).

Currently, this relationship is handled directly in the `Chunk` model using a generic foreign key:

```python
content_object = GenericForeignKey("content_type", "object_id")
content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, blank=True, null=True)
```

### Proposed Changes
Create a new `Context` model to store:

* Generated text
* Optional relation to OWASP entities via `GenericForeignKey`
* Metadata (timestamps, source, etc.)
* Update the `Chunk` model to link to the new `Context` model instead of handling entity relations directly.
* Migrate existing data from `Chunk` to `Context` to ensure backward compatibility.

### Goals

* Centralize the management of AI-generated contexts.
* Provide a scalable foundation for **NestBot AI Assistant** to handle context-aware responses.
* Decouple content chunks from OWASP entities to improve maintainability and future extension.

### Tasks

1. Define `Context` model with required fields and generic relation to OWASP entities.
2. Update `Chunk` model to reference `Context`.
3. Write migration to move existing `content_type` and `object_id` from `Chunk` to `Context`.
4. Update any code paths that rely on `Chunk.content_object` to use `Chunk.context`.
5. Add tests to ensure data integrity and backward compatibility.

Dishant1804:
## Proposed change

Resolves #1873 

- [x] Define Context model with required fields and generic relation to OWASP entities.
- [x] Update Chunk model to reference Context.
- [x] Write migration to move existing content_type and object_id from Chunk to Context.
- [x] Update any code paths that rely on Chunk.content_object to use Chunk.context.
- [x] Add tests to ensure data integrity and backward compatibility.

## Checklist


backend/apps/ai/models/chunk.py
```diff
@@ -16,25 +15,18 @@ class Chunk(TimestampedModel):
...
+        unique_together = ("context", "text")
 
-    content_object = GenericForeignKey("content_type", "object_id")
-    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, blank=True, null=True)
+    context = models.ForeignKey(
+        Context, on_delete=models.CASCADE, related_name="chunks", null=True, blank=True
+    )
     embedding = VectorField(verbose_name="Embedding", dimensions=1536)
-    object_id = models.PositiveIntegerField(default=0)
     text = models.TextField(verbose_name="Text")
```
arkid15r, line 22: Do we still need `text` after adding `context`?

backend/apps/ai/models/chunk.py
```diff
@@ -16,25 +15,18 @@ class Chunk(TimestampedModel):
     class Meta:
         db_table = "ai_chunks"
         verbose_name = "Chunk"
-        unique_together = ("content_type", "object_id", "text")
+        unique_together = ("context", "text")
 
-    content_object = GenericForeignKey("content_type", "object_id")
-    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, blank=True, null=True)
+    context = models.ForeignKey(
```
arkid15r: Do you expect a context to have multiple related chunks?

arkid15r: Also do we want it to be nullable?

backend/apps/ai/models/context.py
```diff
@@ -0,0 +1,33 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+
+from apps.common.models import TimestampedModel
+
+
+class Context(TimestampedModel):
+    """Context model for storing generated text and optional relation to OWASP entities."""
+
+    generated_text = models.TextField(verbose_name="Generated Text")
```
arkid15r: Maybe `content` instead?

backend/apps/ai/models/context.py
```diff
@@ -0,0 +1,33 @@
...
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, blank=True, null=True)
+    object_id = models.PositiveIntegerField(default=0)
+    content_object = GenericForeignKey("content_type", "object_id")
+    source = models.CharField(max_length=100, blank=True, default="")
+
+    class Meta:
```
arkid15r, line 24: Only a single context instance should exist for each entity.

backend/apps/ai/models/chunk.py
Dishant1804: yes as this is the chunked text

backend/apps/ai/models/chunk.py
Dishant1804: multiple chunks are possible

backend/apps/ai/common/utils.py
```diff
@@ -53,7 +60,7 @@ def create_chunks_and_embeddings(
             if (
                 chunk := Chunk.update_data(
                     text=text,
-                    content_object=content_object,
+                    context=context,
```
arkid15r: We need to de-couple chunk/context handling. I'd like to have independent jobs for context updates and chunks generation.

arkid15r (commented):
The `context` introduction requires more structural changes to how we run the workflows. We need to split context/chunk generation processes. Now it's context first perspective. This can be done either in this or the next PR -- up to you.

Dishant1804:
> The `context` introduction requires more structural changes to how we run the workflows. We need to split context/chunk generation processes. Now it's context first perspective. This can be done either in this or the next PR -- up to you.

let's de-couple the logic in this PR itself

backend/apps/ai/common/utils.py
arkid15r: Has this suggestion been addressed?

backend/apps/ai/common/utils.py
```diff
@@ -4,29 +4,54 @@
...
+        content_object: Optional related object
+        source (str): Source identifier
+
+    Returns:
+        Context: Created Context instance
+
+    """
+    context = Context(content=content, content_object=content_object, source=source)
+    context.save()
+    return context
```
arkid15r: I couldn't find a logic related to context updates -- only creation. It seems this approach create multiple contexts to the same entity while the idea is have a single source of truth for each entity context.

This should be part of `Context` model -- you already had similar issue to fix earlier. Please follow our regular way of handling model updates.

backend/apps/ai/common/utils.py
Dishant1804: yes I have handled the de-coupling with the help of flags in `ai_create_*_chunks.py` files

arkid15r: >  yes I have handled the de-coupling with the help of flags in ai_create_*_chunks.py files

So in this decoupled solution ai_create_*_chunks.py jobs also handle the related contexts?

How can I update a context w/o re-generating chunks for it?

backend/apps/ai/common/extractors.py
```diff
@@ -0,0 +1,269 @@
...
+    prose_parts = []
+    metadata_parts = []
+
+    if committee.description:
+        prose_parts.append(f"Description: {committee.description}")
+
+    if committee.summary:
+        prose_parts.append(f"Summary: {committee.summary}")
+
+    if hasattr(committee, "owasp_repository") and committee.owasp_repository:
```
arkid15r: Do you need this `hasattr(committee, "owasp_repository")` check?

backend/apps/ai/common/extractors.py
```diff
@@ -0,0 +1,269 @@
...
+        metadata_parts.append(f"Licenses: {', '.join(project.licenses)}")
+
+    if project.tags:
+        metadata_parts.append(f"Tags: {', '.join(project.tags)}")
+
+    if project.custom_tags:
+        metadata_parts.append(f"Custom Tags: {', '.join(project.custom_tags)}")
+
+    stats_parts = []
+    if (project.stars_count or 0) > 0:
```
arkid15r: These are integers which are 0 by default. Would `if project.stars_count:` work here?

backend/apps/ai/common/extractors.py
```diff
@@ -0,0 +1,269 @@
...
+
+    if project.released_at:
+        metadata_parts.append(f"Last Release: {project.released_at.strftime('%Y-%m-%d')}")
+
+    if project.health_score is not None:
+        metadata_parts.append(f"Health Score: {project.health_score:.2f}")
+
+    metadata_parts.append(f"Active Project: {'Yes' if project.is_active else 'No'}")
+
+    metadata_parts.append(f"Issue Tracking: {'Enabled' if project.track_issues else 'Disabled'}")
```
arkid15r: This is an internal field, remove it from the context.

backend/apps/ai/common/extractors.py
arkid15r, line 1: Split this file into extractors/\<entity\>.py

backend/apps/ai/common/extractors.py
```diff
@@ -0,0 +1,269 @@
...
+        metadata_parts.append(f"Chapter Leaders: {', '.join(chapter.leaders_raw)}")
+
+    if chapter.related_urls:
+        invalid_urls = getattr(chapter, "invalid_urls", []) or []
+        valid_urls = [url for url in chapter.related_urls if url and url not in invalid_urls]
+
+        if valid_urls:
+            metadata_parts.append(f"Related URLs: {', '.join(valid_urls)}")
+
+    metadata_parts.append(f"Active Chapter: {'Yes' if chapter.is_active else 'No'}")
```
arkid15r: Let's not include inactive entities -- we shouldn't mention them at all.

backend/apps/ai/common/utils.py
```diff
@@ -4,31 +4,72 @@
...
+    """
+    context = Context.update_data(content=content, content_object=content_object, source=source)
+    if context is None:
+        if content_object:
+            content_type = ContentType.objects.get_for_model(content_object)
+            context = Context.objects.get(
+                content_type=content_type, object_id=content_object.pk, content=content
+            )
+        else:
+            context = Context.objects.get(content=content, content_object__isnull=True)
```
arkid15r: Why `update_data` can't handle this?

backend/apps/ai/management/commands/ai_create_event_chunks.py
```diff
@@ -55,79 +57,61 @@ def handle(self, *args, **options):
...
+        batch_chunks = []
+
+        event_content_type = ContentType.objects.get_for_model(Event)
+        event_ids = [e.id for e in events]
+        contexts_by_id = {
+            c.object_id: c
+            for c in Context.objects.filter(
+                content_type=event_content_type, object_id__in=event_ids
+            )
+        }
```
arkid15r: What's your idea behind generating this dict? The code doesn't look clear to me.

backend/apps/ai/management/commands/ai_create_project_context.py
arkid15r, line 1: Can we apply some good OOP approach here and consolidate the common logic in a base class and have only custom parts in project/event/chapter/... related commands?

backend/apps/ai/models/context.py
```diff
@@ -0,0 +1,71 @@
...
+                return None
+        elif Context.objects.filter(content=content, content_object__isnull=True).exists():
+            return None
+
+        context = Context(content=content, content_object=content_object, source=source)
+
+        if save:
+            context.save()
+
+        return context
```
arkid15r: This code combined with the https://github.com/OWASP/Nest/pull/1891/files#diff-6f55d3dd696736b0b7a32b136a57d3fac31731e6c6b1430ee62da032020a71f7R32 looks too cumbersome. I'm not sure maybe we should make `content_type` and `object_id` required fields to simplify this?

arkid15r (changes requested):
Some structure and complexity related suggestions:

backend/apps/ai/common/extractors.py
Dishant1804: no we already have the owasp_repository, removing the extra check

backend/apps/ai/common/extractors.py
Dishant1804: this was suggested by code rabbit, I'll remove it

backend/apps/ai/management/commands/ai_create_event_chunks.py
Dishant1804: it was again suggested by code rabbit, I was a bit hesitant to implement it, I'll remove it because it is not readable and our previous implementation works properly

backend/apps/ai/common/extractors/chapter.py
```diff
@@ -0,0 +1,80 @@
...
+        prose_parts.append(f"Description: {chapter.description}")
+
+    if chapter.summary:
+        prose_parts.append(f"Summary: {chapter.summary}")
+
+    if chapter.owasp_repository:
+        repo = chapter.owasp_repository
+        if repo.description:
+            prose_parts.append(f"Repository Description: {repo.description}")
+        if repo.topics and hasattr(repo.topics, "__iter__") and not isinstance(repo.topics, str):
```
arkid15r: Why is this needed? Do we somehow store different type content in chapter topics?

backend/apps/ai/management/commands/ai_create_chapter_chunks.py
```diff
@@ -1,161 +1,29 @@
...
-        if chapter.meetup_group:
-            metadata_parts.append(f"Meetup Group: {chapter.meetup_group}")
-
-        if chapter.tags:
-            metadata_parts.append(f"Tags: {', '.join(chapter.tags)}")
-
-        if chapter.topics:
-            metadata_parts.append(f"Topics: {', '.join(chapter.topics)}")
+class Command(BaseChunkCommand):
+    @property
```
arkid15r: Can we use class level attributes instead of properties here?

backend/apps/ai/common/base.py
```diff
@@ -0,0 +1,257 @@
...
+
+class BaseAICommand(BaseCommand, ABC):
+    """Base class for AI management commands with common functionality."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize the AI command with OpenAI client placeholder."""
+        super().__init__(*args, **kwargs)
+        self.openai_client: openai.OpenAI | None = None
+
+    @property
```
arkid15r: Why do we need to define them as properties vs class level attributes?

backend/apps/ai/common/base.py
```diff
@@ -0,0 +1,257 @@
...
+            processed_count += process_batch_func(list(batch_items))
+
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"Completed processing {processed_count}/{total_count} {self.entity_name_plural}"
+            )
+        )
+
+
+class BaseContextCommand(BaseAICommand):
```
arkid15r: Please split this into 3 files -- a class per file.

backend/apps/ai/common/utils.py
```diff
@@ -4,31 +4,61 @@
...
     DEFAULT_LAST_REQUEST_OFFSET_SECONDS,
     MIN_REQUEST_INTERVAL_SECONDS,
 )
 from apps.ai.models.chunk import Chunk
+from apps.ai.models.context import Context
 
 logger: logging.Logger = logging.getLogger(__name__)
 
 
+def create_context(content: str, content_object=None, source: str = "") -> Context:
```
arkid15r: This one looks redundant.

backend/apps/ai/models/chunk.py
```diff
@@ -54,35 +44,27 @@ def split_text(text: str) -> list[str]:
...
-        ).exists():
-            return None
-
-        chunk = Chunk(
-            content_type=content_type, object_id=content_object.id, text=text, embedding=embedding
-        )
+        chunk = Chunk(text=text, embedding=embedding)
 
         if save:
+            if chunk.context_id is None:
```
arkid15r: Why is this a case?

backend/apps/ai/common/utils.py
```diff
@@ -39,26 +69,24 @@ def create_chunks_and_embeddings(
...
+            model=model,
         )
+        embeddings = [d.embedding for d in response.data]
+
+        chunks = []
+        for text, embedding in zip(chunk_texts, embeddings, strict=True):
+            chunk = Chunk.update_data(text=text, embedding=embedding, save=False)
+            chunk.context = context
+            if save:
+                chunk.save()
```
arkid15r: Any reason why this is out of `update_data` method?

backend/apps/ai/admin.py
arkid15r, line 1: Can you order the changes you add according to existing ordering convention?

arkid15r (changes requested):
It looks cleaner, however lacks tests:

```
apps/ai/agent/tools/rag/generator.py                                              36     24      6      0  28.57%   45-51, 63-74, 87-120
apps/ai/agent/tools/rag/rag_tool.py                                               12      4      0      0  66.67%   31-32, 53-55
apps/ai/agent/tools/rag/retriever.py                                              78     60     30      0  16.67%   42-48, 60-71, 75-79, 92-187, 208-252, 264-275
apps/ai/common/base.py                                                           129      0     30      1  99.37%   230->201
apps/ai/common/constants.py                                                        5      0      0      0 100.00%
apps/ai/common/extractors/chapter.py                                              48     46     38      0   2.33%   16-77
apps/ai/common/extractors/committee.py                                            29     27     22      0   3.92%   16-52
apps/ai/common/extractors/event.py                                                23     21     18      0   4.88%   16-46
apps/ai/common/extractors/project.py                                              60     58     52      0   1.79%   16-94
apps/ai/common/utils.py                                                           35      7      8      3  76.74%   31, 59-60, 69, 81->83, 88-90
```

Some comments from another review round:

backend/apps/ai/models/chunk.py
Dishant1804: just an extra check I'll remove it

backend/apps/ai/common/base.py
Dishant1804: I looked at the other code and i found that it is not necessary to define the abstarctmethods as property that is not how we write it

backend/apps/ai/common/utils.py
Dishant1804: the field is a required foreign key If we let update_data save the chunk with save=True, it would try to save a chunk without a context, which would fail because the context is required field in the chunks model

backend/apps/ai/common/utils.py
Dishant1804: can you elaborate this one I am unable to understand it

backend/apps/ai/admin.py
Dishant1804: can you elaborate this one I am unable to understand it

arkid15r: Your ContextAdmin class goes before teh ChunkAdmin and the same for register(). Compare them to the imports order for example.

backend/apps/ai/common/utils.py
arkid15r: This function only has 1 statement -- Context.update_data.
My question -- why do you need this function at all?

backend/apps/ai/common/utils.py
arkid15r: I don't understand your logic, sorry.
The `context` is required but you don't pass it into update_data but set later and save the model using the duplicated code? This all can be done within update_data method. Am I missing something?

arkid15r: Please also don't mark reviewers' comments as resolved -- it's up to them to decide.

backend/apps/ai/common/base.py
arkid15r: Not sure if that means you're migrating to the suggested approach?

Dishant1804: I have removed property

backend/apps/ai/models/chunk.py
```diff
@@ -54,33 +44,24 @@ def split_text(text: str) -> list[str]:
...
-
-        if Chunk.objects.filter(
-            content_type=content_type, object_id=content_object.id, text=text
-        ).exists():
-            return None
-
-        chunk = Chunk(
-            content_type=content_type, object_id=content_object.id, text=text, embedding=embedding
-        )
+        chunk = Chunk(text=text, embedding=embedding, context=context)
```
arkid15r, line 71: By this design what happens to existing/old chunks if you only create new ones in update_data method? How do we make sure the context and chunks are still in sync?

backend/apps/ai/admin.py
arkid15r: This is still not addressed for some reason 🤷‍♂️

backend/apps/ai/common/base.py
arkid15r: Your classes still are full of excessive abstract methods and now functions that just return values (e.g. `entity_name, `model_class`, etc). I'm asking again to make them just class level attributes. Please.

backend/apps/ai/models/chunk.py
arkid15r: I still don't see answer to my question. Most of them are ignored for some reason (with or without code updates).

arkid15r: Your current implementation addresses chunk duplication. What are we going to do with stale chunks that are no longer in sync with context?

backend/apps/ai/models/context.py
```diff
@@ -0,0 +1,55 @@
...
+from apps.common.models import TimestampedModel
+
+
+class Context(TimestampedModel):
+    """Context model for storing generated text related to OWASP entities."""
+
+    content = models.TextField(verbose_name="Generated Text")
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    content_object = GenericForeignKey("content_type", "object_id")
```
arkid15r: I suggest renaming these to `entity_type`, `entity_id`, `entity` to avoid confusion with the `content` field.

backend/apps/ai/models/context.py
```diff
@@ -0,0 +1,55 @@
...
+    def update_data(
+        content: str,
+        content_object,
+        source: str = "",
+        *,
+        save: bool = True,
+    ) -> "Context":
+        """Retrieve existing or create new context."""
+        content_type = ContentType.objects.get_for_model(content_object)
+        object_id = content_object.pk
```
arkid15r: This variable is used just once.

backend/apps/ai/models/context.py
```diff
@@ -0,0 +1,55 @@
...
+        existing_context = Context.objects.filter(
+            content_type=content_type, object_id=object_id, content=content
+        ).first()
+        if existing_context:
+            return existing_context
+
+        context = Context(content=content, content_object=content_object, source=source)
```
arkid15r: Here I have the same question -- is my understanding correct that we are going to have multiple context instances for a single object?

backend/apps/ai/management/commands/ai_create_chapter_chunks.py
arkid15r, line 1: These files (both chunk and context) should be named `ai_update_*_chunks` with actual update logic instead of multiplying objects.

arkid15r (changes requested):

backend/apps/ai/admin.py
Dishant1804: I have made the changes now

backend/apps/ai/models/context.py
Dishant1804: I'll refactor it

backend/apps/ai/models/context.py
Dishant1804: the intended behaviour is to have one context per entity so I have updated to  `unique_together = ("entity_type", "entity_id")`

backend/apps/ai/models/context.py
Dishant1804: made these changes

backend/apps/ai/management/commands/ai_create_chapter_chunks.py
Dishant1804: I have updated the names of the context files as I have written a logic that regenerates the old stale chunks and creates new one for a particular context that is updated, in this way we can maintain a sync between context and chunks

backend/apps/ai/models/context.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+        if not created and (context.content != content or context.source != source):
+            context.content = content
+            context.source = source
+            if save:
+                context.save(update_fields=["content", "source"])
+            regenerate_chunks_for_context(context=context)
+
+        return context
```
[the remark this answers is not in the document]
Dishant1804: the context will never be None, It has to exist

backend/apps/ai/models/context.py
```diff
@@ -0,0 +1,66 @@
...
+            entity_id=content_object.pk,
+            defaults={"content": content, "source": source},
+        )
+
+        if not created and (context.content != content or context.source != source):
+            context.content = content
+            context.source = source
+            if save:
+                context.save(update_fields=["content", "source"])
+            regenerate_chunks_for_context(context=context)
```
arkid15r: This introduces a tight coupling between context/chunks again.
If you check **any** `model::update_data` method you'll see that it never creates a new instance unless it has`save=True` (this allows to use bulk save). It also never deals with unrelated entities.

The context data update via `update_data` has nothing to do with chunks regeneration directly.

backend/apps/ai/management/commands/ai_create_chapter_chunks.py
```diff
@@ -1,161 +1,16 @@
...
-                for url in chapter.related_urls
-                if url and url not in (chapter.invalid_urls or [])
-            ]
-            if valid_urls:
-                metadata_parts.append(f"Related URLs: {', '.join(valid_urls)}")
-
-        metadata_parts.append(f"Active Chapter: {'Yes' if chapter.is_active else 'No'}")
+class Command(BaseChunkCommand):
+    entity_name = "chapter"
+    entity_name_plural = "chapters"
```
arkid15r: It looks like all plural ones are just name + `s`. You can get rid of this by replacing with a property that return entity_name with `s` -- can be overridden where needed.

backend/apps/ai/management/commands/ai_create_chapter_chunks.py
```diff
@@ -1,161 +1,16 @@
...
+    key_field_name = "key"
+    model_class = Chapter
```
arkid15r: I believe model_class is the only thing you need to have (entity_name and entity_name_plural can be computed based on it).

backend/apps/ai/models/context.py
arkid15r: Moreover, why do we need `ai_create*_chunks` commands for?

arkid15r (changes requested):
I can't merge this PR until the code is clean and maintains context/chunk de-coupled approach I suggested earlier.

backend/apps/ai/models/context.py
Dishant1804: I am planning to add a filed in `context.py` model which will be a boolean flag `are_chunks_stale`, this will tell us whether the chunks associated with the particular context are stale, so that we can regenerate them

does this sound good?

arkid15r: This sounds like a redundant flag. Can it be determined based on context last update date and chunk creation date?

Dishant1804: yes we can do that, i'll implement it

backend/apps/ai/models/context.py
```diff
@@ -0,0 +1,67 @@
...
+from apps.common.models import TimestampedModel
+
+logger = logging.getLogger(__name__)
+
+
+def regenerate_chunks_for_context(context):
+    """Import regenerate_chunks_for_context to avoid circular import."""
+    from apps.ai.common.utils import regenerate_chunks_for_context as _regenerate_chunks
+
+    return _regenerate_chunks(context)
```
arkid15r: This is here for what purpose?

backend/apps/ai/common/extractors/chapter.py
```diff
@@ -0,0 +1,76 @@
+"""Content extractor for Chapter."""
+
+from apps.ai.common.constants import DELIMITER
+
+
+def extract_chapter_content(chapter) -> tuple[str, str]:
```
arkid15r, line 6: I have 2 suggestions for the extractors:
 - why do you need separation for prose/metadata if you join them later anyway?
  - this needs to be migrated to Jinja templates like ones we have for Slack responses
  
Please create an issue to track this refactoring.

backend/apps/ai/common/utils.py
```diff
@@ -39,26 +52,50 @@ def create_chunks_and_embeddings(
...
+def regenerate_chunks_for_context(context: Context):
+    """Regenerates all chunks for a single, specific context instance.
+
+    Args:
+      context (Context): The specific context instance to be updated.
+
+    """
+    from apps.ai.models.chunk import Chunk
+
+    old_chunk_count = context.chunks.count()
```
arkid15r: Is it possible to delete them w/o pre-checks?

backend/apps/ai/agent/tools/rag/retriever.py
```diff
@@ -201,51 +196,45 @@ def retrieve(
...
-            if not chunk.content_object:
+            if not chunk.context or not chunk.context.entity:
                 logger.warning("Content object is None for chunk %s. Skipping.", chunk.id)
                 continue
 
-            source_name = self.get_source_name(chunk.content_object)
+            source_name = self.get_source_name(chunk.context.entity)
             additional_context = self.get_additional_context(
-                chunk.content_object, chunk.content_type.model
+                chunk.context.entity, chunk.context.entity_type.model
```
arkid15r: It seems you don't need the second param as you can figure that our from the model instance inside of `get_additional_context`

backend/apps/ai/models/context.py
Dishant1804: I was trying to resolve the circular import I cleaned out the code but this missed out, removed it

backend/apps/ai/common/utils.py
Dishant1804: yes -- removed the checks

backend/tests/apps/common/open_ai_test.py
```diff
@@ -60,6 +60,3 @@ def test_complete_general_exception(self, mock_openai, mock_logger, openai_insta
         response = openai_instance.complete()
 
         assert response is None
-        mock_logger.exception.assert_called_once_with(
```
arkid15r, line 63: Why this change?

backend/tests/apps/ai/models/context_test.py
```diff
@@ -0,0 +1,376 @@
...
+                type(context),
+                "entity_type",
+                new_callable=PropertyMock,
+                return_value=entity_type,
+            ),
+        ):
+            result = str(context)
+            assert result == (
+                "test_model Test Object: This is test content that is longer than 50 charac"
+                # cspell:ignore charac
```
arkid15r: This is not something we normally do. Use `truncate` for Context::__str__ instead.

arkid15r (changes requested):

backend/tests/apps/common/open_ai_test.py
Dishant1804: I acidentally removed it while I was testing other assertions -- I have added it back

backend/tests/apps/ai/models/context_test.py
Dishant1804: made the changes

arkid15r (approved):

PR #1967, opened 2025-08-04
Improve dashboard metrics list

### LINKED ISSUE #1958: Improve project health dashboard

 - Do not show inactive projects in the `Project Health Metrics` list
 - Remove duplicated projects from the `Project Health Metrics`

- [ ] Yes
- [x] No

**Additional context**

Project iGNITA is inactive and appears on both pages 1,2.

Image
Image

ahmedxgouda:
## Proposed change

Resolves #1958

Now we should not have neither inactive projects nor duplicates

## Checklist


backend/apps/owasp/api/internal/filters/project_health_metrics.py
```diff
@@ -20,3 +20,8 @@ class ProjectHealthMetricsFilter:
     def level(self, value: str, prefix: str):
         """Filter by project level."""
         return Q(project__level=ProjectLevel(value)) if value else Q()
+
+    @strawberry_django.filter_field
+    def is_active(self, *, value: bool, prefix: str):
+        """Filter by active projects."""
+        return Q(project__is_active=value) if value else Q()
```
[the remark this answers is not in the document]
ahmedxgouda: Actually, we don't want to filter inactive projects at all. It is a workaround to eliminate inactive projects from the query.

arkid15r (changes requested):
It seems the filtering for active/inactive projects is available for the end-user. We don't want to expose inactive projects at all -- they just should not exist for a user is they are inactive.

backend/apps/owasp/api/internal/ordering/project_health_metrics.py
```diff
@@ -10,4 +10,9 @@
 class ProjectHealthMetricsOrder:
     """Ordering for Project Health Metrics."""
 
-    score: auto
+    score: strawberry.auto
+    # We need to order by another field in case of equal scores
+    # to ensure unique metrics in pagination.
+    # because SQL returns random order if no order is specified.
+    # We didn't do this ordering in the model since we order already in the query.
+    project__name: strawberry.auto
```
arkid15r, line 19: Any reason for having `__` in the name?

frontend/src/app/projects/dashboard/metrics/page.tsx
```diff
@@ -99,7 +99,12 @@ const MetricsPage: FC = () => {
     variables: {
       filters,
       pagination: { offset: 0, limit: PAGINATION_LIMIT },
-      ordering,
+      ordering: [
+        ordering,
+        {
+          ['project_Name']: 'ASC',
```
arkid15r, line 105: What naming convention was used here?

backend/apps/owasp/api/internal/ordering/project_health_metrics.py
ahmedxgouda: strawberry use these attributes in the `order_by`  method. project_name is not an attribute of the metrics but it is an attribute of the project. And like we access nested attributes in django methods, strawberry does the same.

frontend/src/app/projects/dashboard/metrics/page.tsx
ahmedxgouda: It is automatically converted by strawberry. Since the attribute is `project__name`, it is handled like other attributes. Just first underscore removed and capitalize the second name.

arkid15r (approved):

frontend/src/app/projects/dashboard/metrics/page.tsx
arkid15r: I'm not sure how it works though. Let's merge it as is and improve sorting capabilities later.

PR #2024, opened 2025-08-10
Sync OWASP Awards data and integrate with user profiles

### LINKED ISSUE #1766: Sync OWASP Awards data and integrate with user profiles

We want to import and manage OWASP Awards data within Nest based on the canonical source at https://github.com/OWASP/owasp.github.io/blob/main/_data/awards.yml

1. Create a new model in the owasp Django app that aligns with the structure of the awards YAML file.

The model should include at least the following fields:

    category: CharField (e.g. "WASPY", "Lifetime Achievement")
    name: CharField (unique)
    description: TextField
    year: IntegerField
    user: ForeignKey to User (optional, award recipient)
    created_at, updated_at: standard timestamp fields

2. Implement award sync command

Create a new management command similar to [owasp_update_events.py](https://github.com/OWASP/Nest/blob/main/backend/apps/owasp/management/commands/owasp_update_events.py) that:

    - Downloads and parses awards.yml from the owasp.github.io repository
    - Creates or updates award records in the database
    - Attempts to match recipients with existing users by name or GitHub username
    - Log warnings for unmatched users, and make the command idempotent.

3. Integrate with badges job

Once the sync is in place, the daily badge job will need to be updated to:

    - Assign a "WASPY Award Winner" badge to users who have received any WASPY award (based on the Award model)
    - Ensure the badge is removed if the award is no longer associated with the user

- [ ] Yes
- [x] No

**Additional context**

Depends on Badge implementation completion.

trucodd:
## Proposed change

Resolves #1766

Awards are synced from canonical source
Users are matched to awards automatically
Data is stored and accessible via Award model
Integrated into daily sync pipeline


backend/apps/nest/models/badge.py
```diff
@@ -0,0 +1,110 @@
+"""Nest app badge models."""
+
+from __future__ import annotations
+
+from django.db import models
+
+from apps.common.models import TimestampedModel
+
+
+class BadgeType(TimestampedModel):
```
arkid15r: The Badge model already exists at `owasp/models/badge.py`

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,246 @@
...
+    award_type = models.CharField(
+        verbose_name="Award Type",
+        max_length=20,
+        choices=[
+            ("category", "Category"),
+            ("award", "Award"),
+        ],
+        default="award",
+        help_text="Type of entry: category definition or individual award",
+    )
```
arkid15r: This doesn't look right, it seems we need `Distinguished Lifetime Memberships` and `WASPY` -- please double check.

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,246 @@
...
+        default="",
+        help_text="Path to winner's image",
+    )
+
+    # Award type from YAML (category or award)
+    award_type = models.CharField(
+        verbose_name="Award Type",
+        max_length=20,
+        choices=[
+            ("category", "Category"),
```
arkid15r: Please also make the categories a separate subclass (see an [example](https://github.com/OWASP/Nest/blob/9df6fae00d9c787c1b97c9b9ca74f4610facdd19/backend/apps/owasp/models/snapshot.py#L14))

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,246 @@
...
+        help_text="Award name/title (e.g., 'Event Person of the Year')",
+    )
+    description = models.TextField(
+        verbose_name="Description", blank=True, default="", help_text="Award description"
+    )
+    year = models.IntegerField(
+        verbose_name="Year",
+        null=True,
+        blank=True,
+        help_text="Year the award was given (null for category definitions)",
```
arkid15r: The category part needs to be hardcoded. I don't think we should create a separate table for them as they rarely change.

arkid15r: After removing that you can focus completely on award entities.

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,246 @@
...
+    def get_waspy_award_winners(cls):
+        """Get all users who have won WASPY awards.
+
+        Returns:
+            QuerySet of GitHub users who have won WASPY awards
+
+        """
+        from apps.github.models.user import User
+
+        return User.objects.filter(awards__category="WASPY", awards__award_type="award").distinct()
```
arkid15r: Reuse WASPY value after defining a separate choice field for categories.

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,246 @@
...
+                fields=["name", "year", "category"], name="unique_award_name_year_category"
+            )
+        ]
+
+    # Core fields based on YAML structure
+    category = models.CharField(
+        verbose_name="Category",
+        max_length=100,
+        help_text="Award category (e.g., 'WASPY', 'Lifetime Achievement')",
+    )
```
arkid15r:
```suggestion
```

arkid15r (changes requested):
This is a decent code 👍 
Moreover, I like some of your ideas around badges implementation.

Let's do the following: remove badge related code from the PR while focusing on syncing the awards and matching them to users.

We use badges separately just for displaying some user attributes to visitors

trucodd:
The badge model was not there when I started so I created I will refactor to use the existing model and make the other requested changes.

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,183 @@
...
+            "Distinguished Lifetime Memberships",
+            "Distinguished Lifetime Memberships",
+        )
+
+    class Meta:
+        db_table = "owasp_awards"
+        indexes = [
+            models.Index(fields=["category", "year"], name="owasp_award_category_year"),
+            models.Index(fields=["-year"], name="owasp_award_year_desc"),
+            models.Index(fields=["name"], name="owasp_award_name"),
```
arkid15r: I'd prefer to introduce indexes later.

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,183 @@
...
+        default="",
+        help_text="Name of the award recipient",
+    )
+    winner_info = models.TextField(
+        verbose_name="Winner Information",
+        blank=True,
+        default="",
+        help_text="Detailed information about the winner",
+    )
+    winner_image = models.CharField(
```
arkid15r: It probably could be `winner_image_url` with proper URL type?

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,183 @@
...
+    user = models.ForeignKey(
+        "github.User",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        related_name="awards",
+        verbose_name="User",
+        help_text="Associated GitHub user (if matched)",
+    )
+
```
arkid15r, line 86: I think the matching process needs to be verified by a human. Could you introduce `is_reviewed` flag for this model?

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,183 @@
...
+        Returns:
+            Award instance or list of Award instances
+
+        """
+        if award_data.get("type") == "award":
+            return Award._create_awards_from_winners(award_data, save=save)
+        return None
+
+    @staticmethod
+    def _create_awards_from_winners(award_data: dict, *, save: bool = True):
```
arkid15r: This code doesn't look consistent with the rest of models where `update_data` is responsible for a single instance creation/update logic.

backend/apps/owasp/models/award.py
```diff
@@ -0,0 +1,183 @@
...
+            user: GitHub User instance
+
+        Returns:
+            QuerySet of WASPY awards for the user
+
+        """
+        return cls.objects.filter(user=user, category=cls.Category.WASPY)
+
+    @staticmethod
+    def update_data(award_data: dict, *, save: bool = True):
```
arkid15r: Please see [event](https://github.com/OWASP/Nest/blob/main/backend/apps/owasp/models/event.py) or [post](https://github.com/OWASP/Nest/blob/main/backend/apps/owasp/models/post.py) handling as an example (note the bulk saving approach).

backend/apps/owasp/management/commands/owasp_update_badges.py
arkid15r, line 1: This one should wait until the initial badge processing job is implemented in another PR. Then we'll just update it with additional logic.

arkid15r (changes requested):
This code probably works and does the job. However, I'd like to suggest structural improvements and some performance related changes:

trucodd:
> @trucodd if you're still interested in this task it needs some more love with:
> 
> * code conflicts
> * migration conflicts
> * coderabbit suggestions

yes i am still interested

trucodd:
I’ve opened a new PR #2228

PR #2076, opened 2025-08-17
Fix: Bold text on Community dropdown

### LINKED ISSUE #2035: Fix bold text on Community dropdown select
**Describe the bug**
We have a Community dropdown in the header. When user selects it from any other page except for pages in the dropdown itself (e.g. from `Projects`)  - the dropdown looks as it should, with regular text.

When user toggles the dropdown while on one of the dropdown pages (e.g. on the `Organizations` page) - the whole dropdown text gets bold, same as the `Community` item.

We should make sure the dropdown looks the same no matter if user is on one of the Community pages or not. The only difference should be that the item for a page that the user is on - should be bolded/highlighted in the dropdown.

- [ ] Yes
- [x] No

**Screenshots**

Image
Image

Srijatalamarla:
## Proposed change

Resolves #2035 

Before fix:
Screenshot 2025-08-17 094740

After fix:
Screenshot 2025-08-17 094757

## Checklist


kasya (changes requested):
This looks good! 👍🏼 
Could you also add a couple of tests to check for styling in active/inactive state? Nothing crazy, just these two cases.

Srijatalamarla:
Sure! I'll add the tests soon. Thanks for the review!

PR #2114, opened 2025-08-21
Implemented Initial Query Parser for Search Interfaces

### LINKED ISSUE #2038: Implement Query Parser for Search Interfaces
**Description:**
We need to implement a query parser that will be used by both the **OWASP API** and **OWASP UI** search interfaces. The parser should support a GitHub-like syntax for filtering entities based on their attributes.

**Requirements:**

* **Basic string equality checks**

  * Example: `author:name`
* **Comparison operators for numeric fields**

  * Example: `stars:>100`, `stars:<50`, `issues:>=10`
* Ability to combine multiple conditions (logical AND; potential future support for OR).
* Output should be structured in a way that can be easily consumed by backend search logic (e.g., dict or list of conditions).

**Possible Features for Later Iterations:**

* Quoted strings for multi-word values: `project:"OWASP ZAP"`
* OR conditions: `(author:alice OR author:bob)`
* Negations: `-archived:true`

**Suggested Python Packages to Explore:**

* https://pypi.org/project/pyparsing/

**Benefits:**

* Unifies search syntax across OWASP projects.
* Enables advanced filtering and more precise search results.
* Provides a foundation for future improvements in search capabilities.

- [ ] Yes
- [x] No

**Additional context**

* GitHub search syntax: [https://docs.github.com/en/search-github/searching-on-github](https://docs.github.com/en/search-github/searching-on-github)

Adarsh0427:
## Proposed change
Resolves #2038

### Core Features
- **Field Types**: String, Number, Date, Boolean.
- **Comparison Operators**: `>`, `<`, `>=`, `<=`, `=` for numeric and date fields.
- **Free Text Search**: Unmatched tokens go to default field.
- **Case Sensitivity Control**: Optional case-sensitive field matching.
- **Strict Mode**: Toggle between lenient and strict field validation.
- **Error Handling**: Comprehensive error messages for debugging.
- **Field Validations**:
        -  `Quoted strings`: Support for multi-word values: `project:"OWASP ZAP"` (support all ASCII characters inside quotes).
        - `Unquoted string`: Support single word value: `language:python`.
        - `Nested quoted string`: Support nested quoted values: `project:"OWASP "ZAP""`.
        - `Boolean values`: `true`, `false`, `yes`, `no`, `1`, `0`, `on`, `off`.
        - `Date format`: YYYY-MM-DD/YYYYMMDD with comparison operators .
        - `Number`: Integer values within the range of [0,2\^32].
- **Robust Parsing**: Handles malformed queries gracefully

### Query Output Structure:
```python
[
  {
    "type": "string | number | date | boolean",
    "field": "field_name",
    "op": "comparison_operator",   // optional: used only for number/date fields
    "value": '[str | bool | int]'
  }
]
```

### Example 1: Multi-field Query
```
Query: author:alice stars:100 language:"python" archived:false

Output:
{'field': 'author', 'type': 'string', 'value': 'alice'}
{'field': 'stars', 'type': 'number', 'op': '=', 'value': 100}
{'field': 'language', 'type': 'string', 'value': 'python'}
{'field': 'archived', 'type': 'boolean', 'value': False}
```

### Example 2: Numeric Comparison
```
Query: stars:<50 author:bob

Output:
{'field': 'stars', 'type': 'number', 'op': '<', 'value': 50}
{'field': 'author', 'type': 'string', 'value': 'bob'}
```

### Example 3: Quoted String
```
Query: project:"OWASP ZAP"

Output:
{'field': 'project', 'type': 'string', 'value': 'OWASP ZAP'}
```

### Example 4: Date Range Query
```
Query: created:>=2023-01-01 stars:>50

Output:
{'field': 'created', 'type': 'date', 'op': '>=', 'value': '2023-01-01'}
{'field': 'stars', 'type': 'number', 'op': '>', 'value': 50}
```

### Example 5: Free Text Search (`default_field = "query"`)
```
Query: author:"aman" free_text "quoted text" 

Output:
{'type': 'string', 'field': 'author', 'value': 'aman'}
{'type': 'string', 'field': 'query', 'value': 'free_text'}
{'type': 'string', 'field': 'query', 'value': 'quoted text'}
```

## Checklist
- [x] Equality, comparisons, quoted strings
- [x] Implicit AND
- [x] **Unknown field validation with clear error messages**
- [x] **Define initial OWASP-focused field set**
- [x] Query validation and syntax error handling
- [x] **defined comprehensive test cases**


backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,469 @@
...
+
+        """
+        try:
+            result = QueryParser._UNQUOTED_STRING.parseString(value)
+            return result[0]
+        except ParseException as e:
+            raise QueryParserError(
+                message="Invalid string value", error_type="STRING_VALUE_ERROR", field=field
+            ) from e
+
```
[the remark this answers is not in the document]
Adarsh0427: I have handled this issue:
- for quoted string pass the value as it is with quotes for example `query = language:"C# C++"` then `parsed_query = {'type':'string', 'field':'language', 'string':'"C# C++"'}`
- for unquoted string check for `unquoted_value`. `query = language:"C# C++"` will not work.
``` python
_UNQUOTED_VALUE = Word(alphanums + "+-.<>=/_")
@@
def _parse_string_value(field: str, value: str) -> str:
        """Parse and validate string value(handle quotes and unquoted string).

        Args:
            field: Field name
            value: Raw string value

        Returns:
            Cleaned string value

        Raises:
            QueryParserError: If string parsing fails

        """
        if value.startswith('"') and value.endswith('"'):
            return value
        try:
            result = QueryParser._UNQUOTED_VALUE.parseString(value)
            return result[0]
        except ParseException as e:
            msg = f"Invalid string value for {field}:{value}"
            raise QueryParserError(msg) from e

File: backend/apps/common/search/query_parser.py [hunk 2]
```diff
@@ -0,0 +1,469 @@
...
+        token_regex = "|".join(reg)
+        parser = ZeroOrMore(Regex(token_regex))
+
+        try:
+            result = parser.parseString(query, parseAll=True)
+            return result.asList()
+        except ParseException as e:
+            raise QueryParserError(
+                message="Failed to tokenize query", error_type="TOKENIZATION_ERROR", query=query
+            ) from e
```
Adarsh0427, line 212: > **Actionable comments posted: 5**
> 
> 🧹 Nitpick comments (5)
> > backend/apps/common/search/query_parser.py (3)> `166-176`: **Field name validation relies on truthiness of ParseResults; wrap ParseException for consistent error types.**
> > >
> > > Also applies to: 31-37
> > > 
> > > `191-213`: **Tokenization via Regex doesn’t handle escaped quotes inside quoted strings.**
> > > Example: `project:"OWASP \"ZAP\" Guide"` will split incorrectly. Consider replacing `_split_tokens` with a pyparsing-based tokenizer that reuses `_QUOTED_VALUE` and `_TOKEN_PATTERN`, or, at minimum, extend the regex to account for escaped quotes.
> > > I can provide a consolidated grammar (`OneOrMore(Group(_FIELD_NAME + ':' + _FIELD_VALUE) | _QUOTED_VALUE | Word(…))`) to replace `_split_tokens` if you’d like.

what should be the expected output for this  query:  `project:"OWASP \"ZAP\" Guide"`

backend/tests/apps/common/search/query_parser_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+    @pytest.mark.parametrize("query", unknown_field_queries)
+    def test_unknown_field_strict_raises(self, query):
+        with pytest.raises(QueryParserError) as ei:
+            self.strict_parser.parse(query)
+        assert ei.value.error_type == "UNKNOWN_FIELD_ERROR"
+
+    def test_date_format_validation(self):
+        valid_dates = ["2023-12-31", '"2023-12-31"', '"2023-12-31']
+        for date_str in valid_dates:
```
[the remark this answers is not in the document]
Adarsh0427: According to my implementation incorrect quotes can be handled, ex, query= `created:"2024-12-01` can be parser giving output =`{'field': 'created', 'type': 'date', 'op': '=', 'date': '2024-12-01'}`.

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,525 @@
...
+
+    # Grammar definitions (class-level constants)
+    _FIELD_NAME = Word(alphas, alphanums + "_")
+    _QUOTED_VALUE = QuotedString(quoteChar='"', escChar="\\", unquoteResults=False)
+    _UNQUOTED_VALUE = Word(alphanums + '+-.<>=/_"')
+    _FIELD_VALUE = _QUOTED_VALUE | _UNQUOTED_VALUE
+    _COMPARISON_OPERATORS = {">", "<", ">=", "<=", "="}
+    _COMP_OPERATOR = Optional(oneOf(_COMPARISON_OPERATORS), default="=")
+    _COMP_PATTERN = Group(_COMP_OPERATOR + _UNQUOTED_VALUE)
+    _DATE_PATTERN = Regex(r"\d{4}-\d{2}-\d{2}")  # YYYY-MM-DD format
```
arkid15r: Let's support `YYYYMMDD` too

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,525 @@
...
+    - Only supports implicit AND logic between conditions
+    """
+
+    # Grammar definitions (class-level constants)
+    _FIELD_NAME = Word(alphas, alphanums + "_")
+    _QUOTED_VALUE = QuotedString(quoteChar='"', escChar="\\", unquoteResults=False)
+    _UNQUOTED_VALUE = Word(alphanums + '+-.<>=/_"')
+    _FIELD_VALUE = _QUOTED_VALUE | _UNQUOTED_VALUE
+    _COMPARISON_OPERATORS = {">", "<", ">=", "<=", "="}
+    _COMP_OPERATOR = Optional(oneOf(_COMPARISON_OPERATORS), default="=")
```
arkid15r: Use consistent naming -- COMPARISON_OPERATOR, COMPARISON_PATTERN

backend/tests/apps/common/search/query_parser_test.py
```diff
@@ -0,0 +1,181 @@
...
+        self.strict_parser = QueryParser(allowed_fields=self.allowed_fields, strict=True)
+        self.case_sensitive_strict_parser = QueryParser(
+            allowed_fields=self.allowed_fields, strict=True, case_sensitive=True
+        )
+        self.boolean_values = {"true", "1", "yes", "on", "false", "0", "no", "off"}
+        self.true_boolean_values = {"true", "1", "yes", "on"}
+        self.comparison_operators = ["<", "<=", ">", ">=", "="]
+
+    def test_invalid_parser_field_validation(self):
+        with pytest.raises(QueryParserError) as ei:
```
arkid15r: What's `ei`?

backend/tests/apps/common/search/query_parser_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class TestQueryParser:
+    def setup_method(self):
+        self.allowed_fields = {
+            "author": "string",
+            "stars": "number",
+            "issues": "number",
+            "project": "string",
+            "archived": "boolean",
```
arkid15r:
```suggestion
```

backend/tests/apps/common/search/query_parser_test.py
```diff
@@ -0,0 +1,181 @@
...
+            ("case_sensitive_strict_parser", '"John Doe"'),
+        ],
+    )
+    def test_basic_field_parsing_all_types(self, parser_type, expected_string):
+        parser = getattr(self, parser_type)
+        result = parser.parse('Author:"John Doe"')
+        assert len(result) == 1
+        assert result[0]["field"] == "author"
+        assert result[0]["type"] == "string"
+        assert result[0]["string"] == expected_string
```
arkid15r: Let's assert dicts are equal

backend/tests/apps/common/search/query_parser_test.py
```diff
@@ -0,0 +1,181 @@
...
+        assert result[0]["field"] == "author"
+
+    def test_comparison_operators(self):
+        for op in self.comparison_operators:
+            result = self.parser.parse(f"stars:{op}100")
+            assert len(result) == 1
+            assert result[0]["field"] == "stars"
+            assert result[0]["type"] == "number"
+            assert result[0]["op"] == op
+            assert result[0]["number"] == "100"
```
arkid15r: Yes, dict comparison would look cleaner. Please update all assertions.

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,525 @@
...
+
+        Raises:
+            QueryParserError: If parsing fails
+
+        """
+        condition_dict = {
+            "type": field_type.value,
+            "field": field,
+        }
+        if field_type == FieldType.BOOLEAN:
```
arkid15r: Maybe use `match` here?

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,525 @@
...
+            condition_dict["op"] = operator
+            condition_dict["number"] = str(numeric_value)
+        elif field_type == FieldType.DATE:
+            operator, date_value = self._parse_date_value(field, raw_value)
+            condition_dict["op"] = operator
+            condition_dict["date"] = date_value
+        else:  # STRING
+            condition_dict["string"] = self._parse_string_value(field, raw_value)
+
+        return condition_dict
```
arkid15r: I suggest avoiding `_dict` (also _list, _set) suffixes

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,525 @@
...
+
+    @staticmethod
+    def _is_field_name(field_name: str) -> bool:
+        """Check if a field name is valid."""
+        try:
+            QueryParser._FIELD_NAME.parseString(field_name, parseAll=True)
+        except ParseException:
+            return False
+        else:
+            return bool(field_name.islower())
```
arkid15r:
```suggestion

        return field_name.islower()
```

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,525 @@
...
+            raise QueryParserError(
+                message="Failed to tokenize query", error_type="TOKENIZATION_ERROR", query=query
+            ) from e
+
+    @staticmethod
+    def _parse_token(token: str) -> tuple[str | None, str]:
+        """Parse a single token into field name and value.
+
+        Args:
+            token: Individual token string (e.g., "field:value") raise exception if null
```
arkid15r: I didn't notice any raised exceptions

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,525 @@
...
+
+        Returns:
+            Tuple of (field_name, raw_value) or (None, token) if parsing fails
+
+        """
+        token = token.strip()
+        if token[0] == '"':
+            return (None, token)
```
arkid15r: Use tuples with no () -- it's cleaner and good enough in this case.

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,525 @@
...
+    @staticmethod
+    def _parse_boolean_value(field: str, value: str) -> str:
+        """Parse boolean value from string and return its string representation.
+
+        Args:
+            field: Field name
+            value: String value to parse
+
+        Returns:
+            String representation of boolean value ("True" or "False")
```
arkid15r: I see returning str() form of True/False is by design. Could you add more details about this, maybe a comment.

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,525 @@
...
+
+        Raises:
+            QueryParserError: If value is not a valid boolean
+
+        """
+        value = QueryParser._remove_quotes(value)
+        value = value.lower()
```
arkid15r:
```suggestion
        value = QueryParser._remove_quotes(value).lower()
```

arkid15r (changes requested):
Great work @Adarsh0427 

I have suggestions and some questions about the implementation:

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,542 @@
...
+        Raises:
+            QueryParserError: If parsing fails
+
+        """
+        conditions = []
+        query = query.strip()
+        if not self.case_sensitive:
+            query = query.lower()
+        tokens = self._split_tokens(query)
+        try:
```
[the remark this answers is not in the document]
Adarsh0427: field name must always be in lowercase even when `case_sensitive=True`.
For example : query: `Repo:Nest` and  `case_sensitive=True` then also it must pass and output will be `{'field': 'repo', 'type': 'string', 'string': 'Nest'}`

backend/tests/apps/common/search/query_parser_test.py
Adarsh0427: But in this way i will not be able to test `boolean` field types in the testcase.
```python
def test_boolean_variations(self):
  for bool_val in self.boolean_values:
    result = self.parser.parse(f"archived:{bool_val}")
    expected_bool = "True" if bool_val in self.true_boolean_values else "False"
    expected = [{"field": "archived", "type": "boolean", "boolean": expected_bool}]
    assert result == expected

```

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,541 @@
...
+
+        reg = [
+            reg_components["key_value"],
+            reg_components["quoted_string"],
+            reg_components["unquoted_word"],
+        ]
+
+        token_regex = "|".join(reg)
+        parser = ZeroOrMore(Regex(token_regex))
+
```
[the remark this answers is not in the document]
Adarsh0427: But how will it solve this problem can you explain with one more example with the expected output.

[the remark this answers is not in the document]
Adarsh0427: `'project:"OWASP \"ZAP\""'` will not work but `'author:"OWASP \\"ZAP\\""'` will work. Is this expected behaviour ?

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,540 @@
...
+        Raises:
+            QueryParserError: If value is not a valid number
+
+        """
+        try:
+            value = QueryParser._remove_quotes(value)
+            operator, clean_value = QueryParser._parse_comparison_pattern(value)
+            numeric_value = int(float(clean_value))
+            return operator, str(numeric_value)
+        except QueryParserError:
```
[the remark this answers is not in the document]
Adarsh0427: The expected behavior is that number should be converted to int even if floats/exponents are given in field value

backend/apps/common/search/query_parser.py
Adarsh0427: made this change

backend/apps/common/search/query_parser.py
Adarsh0427: changed to this `Returns: String ("True" or "False") based on the boolean input value.`.
Is this statement fine ?

backend/apps/common/search/query_parser.py
Adarsh0427: removed `(`, `)` from return statement

backend/apps/common/search/query_parser.py
Adarsh0427: Yes it will not raise any exception.
I have changed this line by removing `raise exception if null` as it will never get null token

backend/apps/common/search/query_parser.py
Adarsh0427: Changed field name regex expression to this `_FIELD_NAME = Regex(r"[a-z0-9_]+")`.
So it will parse only lower case letters and return `True` if parsed or `False` if failed.

backend/apps/common/search/query_parser.py
Adarsh0427: I have changed this logic, now I am using `match-case` statement instead of `if-else` conditional statements.

backend/apps/common/search/query_parser.py
Adarsh0427: Changed following variables and methods names
- `condition_dict` -> `condition` in create_condition method which will `create different dictionary structure based on field type`. 
- `to_dict` -> `create_condition` in QueryParser Class which will `create a structured condition dictionary from field and raw value`.
- `to_dict` -> `to_response` in QueryParserError Class which will `return error information as dictionary`.

backend/tests/apps/common/search/query_parser_test.py
Adarsh0427: I have made these changes everywhere in `query_parser_test.py` file to assert dicts are equal.

backend/tests/apps/common/search/query_parser_test.py
Adarsh0427: `ei` is an instance name of error which I will use to assert error_type of exception raised.
Now I have changed its name to `exception_info` which is more self explanatory.

backend/apps/common/search/query_parser.py
Adarsh0427: Changed to this :
```python
_COMPARISON_OPERATOR = Optional(oneOf(">= <= > < ="), default="=")
    _COMPARISON_PATTERN = Group(_COMPARISON_OPERATOR + _UNQUOTED_VALUE) = Optional(oneOf(">= <= > < ="), default="=")
_COMPARISON_PATTERN = Group(_COMPARISON_OPERATOR + _UNQUOTED_VALUE)
```
removed `_COMPARISON_OPERATORS` as it was not required and also made sure that naming is consistent everywhere.

backend/apps/common/search/query_parser.py
Adarsh0427: Added support for `YYYYMMDD` date format and also used a function to return the date in `YYYY-MM-DD` so that date format is consistent in every query.
```python 
def _normalize_date(date: str) -> str:
    """Normalize the date string to YYYY-MM-DD format.

    Args:
        date: Raw date string

    Returns:
        Normalized date string

    """
    if "-" not in date:
        return f"{date[:4]}-{date[4:6]}-{date[6:]}"
    return date
``` 
This function will be called once date format is checked, so only two formats can be passed in this function `YYYY-MM-DD` or `YYYYMMDD`. 
So now if query is `created:<20250829` then output will be `{'field': 'created', 'type': 'date', 'op': '<', 'date': '2025-08-29'}`

kasya:
hey @Adarsh0427 ! Could you resolve conflicts on this one? thanks!

Adarsh0427:
@kasya I `merged upstream/main` branch and then ran `make check`, which fixed `pnpm run formate` in some frontend files.
Should I revert back to previous commit?

backend/apps/common/search/query_parser.py
arkid15r: Well, I expected something that explains the decision to return strings instead of booleans. Basically, why do you return "True" instead of `True` here?

backend/apps/common/search/query_parser.py
arkid15r: `to_dict` was legit, the suggestion is valid for variable naming only.

backend/tests/apps/common/search/query_parser_test.py
arkid15r: I see, yes let's keep it.

backend/tests/apps/common/search/query_parser_test.py
arkid15r: `as e` normally works fine

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,541 @@
...
+    - Numeric and date comparisons (>, <, >=, <=, =)
+    - Quoted strings for multi-word values
+    - Boolean field support
+    - Date field parsing with comparisons
+    - Case sensitivity: normalize field names only, preserve original value casing
+    - Only supports implicit AND logic between conditions
+    """
+
+    # Grammar definitions (class-level constants)
+    _FIELD_NAME = Regex(r"[a-z0-9_]+")
```
arkid15r: I don't think the `_` prefix is useful here.
Is it possible to remove it from the grammar definitions?

Adarsh0427: removed `_` prefix from Grammar definition variables

backend/apps/common/search/query_parser.py
Adarsh0427: Renamed `create_condition` and `to_response` to `to_dict`

backend/apps/common/search/query_parser.py
Adarsh0427:
```python
def _parse_boolean_value(value: str) -> str:
        """Parse boolean string to standardized string format.

        Converts boolean-like strings to "True" or "False" for consistent
        storage in dictionaries without needing type conversion.

        Args:
            value: Boolean string to parse

        Returns:
            "True" or "False" as string

        Raises:
            QueryParserError: If value is not a valid boolean

        """
```

Adarsh0427:
> * also address/resolve coderabbit's comments (the ready for review code should have only human authored comments or bot's comments that you couldn't resolve yourself)

Understood, I will take care next time

backend/apps/common/search/query_parser.py
arkid15r:
```python
'author:"John Doe" stars:>100 archived:false some "free text"'
...
{"type": "string", "field": "author", "string": '"john doe"'},
{"type": "number", "field": "stars", "op": ">", "number": "100"},
{"type": "boolean", "field": "archived", "boolean": "False"},
{"type": "string", "field": "query", "string": "some"},
    {"type": "string", "field": "query", "string": '"free text"'},
```

I think it should return naturally expected type/value pairs, e.g. `100` instead of `"100"` for number, `True` / `False` instead of "True"/"False" for boolean, string stays as is.
Any suggestions why it might be not a good idea?

Adarsh0427: I thought in this way that strings are immutable so while doing any query operation, its value does not gets changed by any code mistake and return wrong response.
Now I think you are right because if we have handle any operations like `if query[stars] > 500` then we would have to convert its type from string.

Adarsh0427: Do I need to change the code to store `type/value pairs` ? 
The structure of `return` type of `query_parser` would be `List[Dict[str, str | int | bool]]`.

arkid15r: I think it could be unified with `type/field/op/value` instead of `type/field/op/[string/boolean/number]`

Adarsh0427: I got your point, so the structure of result should be:
```json
{
    "type": "str",
    "field": "str",
    "op": "str",
    "value": "[str | bool | int]"
}
```

arkid15r: Yes, `value` may be one of str, bool, int consistent with `type`.

Adarsh0427: I have updated the parser behavior to return value with its actual data type (int, bool, str) and also set an overflow for integer `below 0 or above 1e32`.
defined dataclass for `QueryCondition`:
```python
@dataclass
class QueryCondition:
    """Dataclass representing a parsed query condition."""

    type: str
    field: str
    value: str | int | bool
    op: str | None = None

    def to_dict(self) -> dict:
        """Convert the QueryCondition to a dictionary representation."""
        data = asdict(self)
        if data.get("op") is None:
            data.pop("op")
        return data
```

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,561 @@
...
+            QueryParserError: If field_schema contains invalid field names or types
+
+        """
+        try:
+            self.field_schema = self._validate_field_schema(field_schema, default_field)
+            self.default_field = default_field
+            self.case_sensitive = case_sensitive
+            self.strict = strict
+        except QueryParserError:
+            raise
```
arkid15r: This doesn't look right.

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,561 @@
...
+        except Exception as e:
```
arkid15r: Do we need to catch all here? It's not really a good practice.

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,561 @@
...
+            operator, clean_value = QueryParser._parse_comparison_pattern(value)
+            numeric_value = int(float(clean_value))
+            if (numeric_value > 2**64 or numeric_value < 0):
+                raise OverflowError
+        except QueryParserError:
+            raise
+        except (ValueError, OverflowError, ParseException) as e:
+            raise QueryParserError(
+                message=f"Invalid numeric value: {e!s}", error_type="NUMBER_VALUE_ERROR"
+            ) from e
```
arkid15r: This exception handling needs to be split/simplified. Normally you don't need to throw an exception just to catch one right away.

backend/apps/common/search/query_parser.py
```diff
@@ -0,0 +1,561 @@
...
+
+        """
+        try:
+            value = QueryParser._remove_quotes(value)
+            operator, clean_value = QueryParser._parse_comparison_pattern(value)
+            result = QueryParser.DATE_PATTERN.parseString(clean_value, parseAll=True)
+            result[0] = QueryParser._normalize_date(result[0])
+            return operator, result[0]
+        except QueryParserError:
+            raise
```
arkid15r: I suggest getting a better understanding of try/except. Please remove.

backend/apps/common/search/query_parser.py
Adarsh0427: will this work 
```python
try:
            value = QueryParser._remove_quotes(value)
            operator, clean_value = QueryParser._parse_comparison_pattern(value)
            numeric_value = int(float(clean_value))
            if numeric_value > 2**64 or numeric_value < 0:
                raise QueryParserError(
                    message=f"Numeric value out of range (0 to 2^64)",
                    error_type="NUMBER_VALUE_ERROR",
                )
        return operator, numeric_value
```

Adarsh0427: updated to
```python
try:
    value = QueryParser._remove_quotes(value)
    operator, clean_value = QueryParser._parse_comparison_pattern(value)
    numeric_value = int(float(clean_value))
    if numeric_value > 2**32 or numeric_value < 0:
        operator = ">" if numeric_value < 0 else operator
        numeric_value = max(0, min(numeric_value, 2**32))
  except ValueError as e:
      raise QueryParserError(
          message=f"Invalid numeric value: {e!s}", error_type="NUMBER_VALUE_ERROR"
      ) from e
  except OverflowError:
      raise QueryParserError(
          message="Numeric value overflow", error_type="NUMBER_VALUE_ERROR"
      )
  return operator, numeric_value
```

backend/apps/common/search/query_parser.py
Adarsh0427: Removed exception handling from this function as it was not required here.

backend/apps/common/search/query_parser.py
Adarsh0427: removed this `except`

backend/apps/common/search/query_parser.py
Adarsh0427: removed this `except`

arkid15r (approved):
Hopefully, this will serve as a strong foundation to build upon.

Thank you @Adarsh0427 👍

PR #2155, opened 2025-08-29
Integrate graphql-codegen to generate GraphQL types

### LINKED ISSUE #1600: Implement codegen schema validation

- [ ] Yes
- [x] No

rudransh-shrivastava:
Resolves #1600

## Proposed change
Configure graphql-codegen using this [documentation](https://www.apollographql.com/docs/react/development-testing/graphql-codegen) 
The codegen generates types and validates the queries.

Use `pnpm run graphql-codegen` to run the codegen.
## Checklist


rudransh-shrivastava:
Had to violate the DRY principle in `graphql-codegen.ts` for fetching csrf token.
I tried using `fetchCsrfTokenServer` at `frontend/src/server/fetchCsrfTokenServer.ts`. But this function gets the csrf url from environment variables. I will use the same script `pnpm run graphql-codegen` for both local and CI/CD runs. 

This made it tricky to use that function and would make the code more complicated as node does not load environment variables from a `.env` file directly.

frontend/src/server/queries/chapterQueries.ts
```diff
@@ -1,7 +1,7 @@
 import { gql } from '@apollo/client'
 
 export const GET_CHAPTER_DATA = gql`
-  query GetChapter($key: String!) {
+  query GetChapterData($key: String!) {
     chapter(key: $key) {
```
rudransh-shrivastava, line 5: The codegen expects unique names for all queries. I have only changed the conflicting query names. There still exist many query names that are inconsistent with the format I chose. I'm not sure if this is the best practice for naming queries, I need to know what you think @arkid15r

frontend/src/server/queries/programsQueries.ts
```diff
@@ -1,23 +1,5 @@
 import { gql } from '@apollo/client'
 
-export const GET_PROGRAM_DATA = gql`
-  query GetPrograms($page: Int!, $search: String, $mentorUsername: String) {
-    allPrograms(page: $page, search: $search, mentorUsername: $mentorUsername) {
```
rudransh-shrivastava, line 5: This was a query was not being used anywhere. It had syntax errors that were never caught due to it not being used.
The error:
```
Error 1: Cannot query field "allPrograms" on type "Query". Did you mean "myPrograms" or "getProgram"?
      at /home/psyduck/repos/Nest/frontend/src/server/queries/programsQueries.ts:3:5
```

frontend/src/server/queries/moduleQueries.ts
```diff
@@ -14,18 +14,15 @@ export const GET_MODULES_BY_PROGRAM = gql`
...
       mentors {
         id
-        githubUser {
-          id
-          login
-          avatarUrl
-        }
+        login
+        avatarUrl
       }
```
rudransh-shrivastava, line 19: This query was not being used anywhere. Fixed syntax errors to run codegen. I can remove it entirely if required.
Error:
```
Error 0: Cannot query field "githubUser" on type "MentorNode".
      at /home/psyduck/repos/Nest/frontend/src/server/queries/moduleQueries.ts:15:9
```

frontend/graphql-codegen.ts
```diff
@@ -0,0 +1,59 @@
+import { CodegenConfig } from '@graphql-codegen/cli'
+
+const PUBLIC_API_URL = process.env.PUBLIC_API_URL || 'http://localhost:8000'
+let csrfToken = null
+
```
rudransh-shrivastava, line 4: I don't think creating a global variable `csrfToken` like this is a good idea.
The other alternative I found was to use a function like `createCodegenConfig` that would query the csrf url once and return the config which would be exported.

Should I do it that way? @arkid15r

frontend/.prettierignore
```diff
@@ -1,2 +1,3 @@
 .pnpm-store/
 pnpm-lock.yaml
+src/types/__generated__/
```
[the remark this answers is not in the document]
rudransh-shrivastava: Should I not commit generated types?

rudransh-shrivastava:
I'm yet to integrate this in the CI/CD pipeline
Edit: its difficult as we discussed, skipping this.

rudransh-shrivastava:
I will also add generated types to `.gitignore` as per coderabbit's suggestion
Edit: Not required anymore, we are committing types

frontend/graphql-codegen.ts
rudransh-shrivastava: Another question, should we commit the generated types? CR suggests to add it to `.gitignore`. 

edit: Yeah, We should not commit it. This will reduce diffs in PRs and sonar code quality check/coderabbit won't flag it.

rudransh-shrivastava: generating these types require a server to be present. if we do it during build time we will need to make sure backend is up before generating types.

rudransh-shrivastava: I've fixed the first comment in this thread. Still doubtful about committing generated types though

frontend/graphql-codegen.ts
```diff
@@ -0,0 +1,52 @@
...
+      [`${PUBLIC_API_URL}/graphql/`]: {
+        headers: {
+          Cookie: `csrftoken=${csrfToken}`,
+          'x-csrftoken': `${csrfToken}`,
+        },
+      },
+    },
+  }
+}
+export default await createCodegenConfig()
```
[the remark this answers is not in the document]
rudransh-shrivastava: This does not work.

frontend/package.json
```diff
@@ -8,6 +8,7 @@
     "dev": "next dev --port 3000",
     "format": "prettier --log-level warn --write .",
     "format:check": "prettier --check .",
+    "graphql-codegen": "graphql-codegen --config graphql-codegen.ts",
```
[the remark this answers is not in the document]
rudransh-shrivastava: how is this possible if we run backend separately? if we invoke codegen before build it may fail as backend may or may not be available. Codegen requires a server  to query the GraphQL endpoint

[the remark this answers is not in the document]
rudransh-shrivastava: 1. to get a `schema.graphql` file, I run into the same problem. I need a backend server to generate that file.

I would like to know if there's a way to run graphql-codegen after backend is available? This is for local development purposes. I am assuming that for CI jobs -- all generated files are already present.

frontend/graphql-codegen.ts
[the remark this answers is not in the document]
rudransh-shrivastava: I have implemented this, please resolve the issue @coderabbitai

frontend/src/server/queries/programsQueries.ts
```diff
@@ -96,7 +78,7 @@ export const GET_PROGRAM_AND_MODULES = gql`
 `
 
 export const GET_PROGRAM_ADMIN_DETAILS = gql`
-  query GetProgramDetails($programKey: String!) {
+  query GetProgramAdminDetails($programKey: String!) {
```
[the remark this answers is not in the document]
rudransh-shrivastava: This query is being used in places. we cannot remove it.

frontend/graphql-codegen.ts
```diff
@@ -0,0 +1,52 @@
...
+    ignoreNoDocuments: true,
+    overwrite: true,
+    schema: {
+      [`${PUBLIC_API_URL}/graphql/`]: {
+        headers: {
+          Cookie: `csrftoken=${csrfToken}`,
+          'x-csrftoken': `${csrfToken}`,
+        },
+      },
+    },
```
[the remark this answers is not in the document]
rudransh-shrivastava: django does not reject them, it works fine.

rudransh-shrivastava:
@arkid15r Hey, this is ready for review.

I had to rename GraphQL queries for the codegen to work (unique names are required). I renamed them based on their variable name (`GET_USER_DATA` became `GetUserData` instead of `GetUser`). There is no official naming convention that I could find. 

I say that we name the queries the same as their variable names but in PascalCase because these generated typed queries will be used instead of current query string variables (see [Example code](https://www.apollographql.com/docs/react/development-testing/graphql-codegen#usage-1) line 6). 

I also think that we should refactor variable and query names as they are very inconsistent. The only constraint we have is that they should be unique.

Sources of documentation I followed: 
[Apollo GraphQL Codegen Guide](https://www.apollographql.com/docs/react/development-testing/graphql-codegen)
[Official Codegen Presets Documentation](https://the-guild.dev/graphql/codegen/plugins/presets/near-operation-file-preset)

.github/workflows/run-ci-cd.yaml
```diff
@@ -56,6 +56,8 @@ jobs:
             pre-commit-${{ runner.os }}-
 
       - name: Run pre-commit
+        env:
+          SKIP: graphql-codegen
```
arkid15r: This should probably be part of `make check` target if you skip it here.
Also this way we'll avoid adding local pre-commit hook

frontend/package.json
```diff
@@ -23,6 +24,7 @@
     "@fortawesome/free-regular-svg-icons": "^6.7.2",
     "@fortawesome/free-solid-svg-icons": "^6.7.2",
     "@fortawesome/react-fontawesome": "^3.0.1",
+    "@graphql-typed-document-node/core": "^3.2.0",
```
arkid15r, line 27: Does apollo pick up the generated types automagically based on the location?

arkid15r (changes requested):
Looks good, let's improve the workflow a bit:

.github/workflows/run-ci-cd.yaml
rudransh-shrivastava: I've made this change. Thank you

frontend/package.json
rudransh-shrivastava: No, We need to import the types in each file where `useQuery` hook is used. 
[Example](https://www.apollographql.com/docs/react/development-testing/graphql-codegen#usage-1)

Is it okay if I refactor all this in a different PR? I will need to update the tests too. I believe this refactor should complete the apollo migration. 

For now, the generated files are unused.

It's recommended by apollo to separate out queries (one generated file for each query/mutation file) for larger applications. If we don't do this, all the queries and mutation are generated in the `graphql.ts` file.

arkid15r (approved):
LGTM

PR #2162, opened 2025-09-01
Add auto review on feature branches to coderabbitai config.

ahmedxgouda:
## Proposed change

Add the PR description here.

## Checklist

- [ ] I've read and followed the [contributing guidelines](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md).

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -9,10 +9,10 @@ reviews:
   assess_linked_issues: true
   auto_apply_labels: false
   auto_review:
-    enabled: true
-    drafts: true
     base_branches:
-      - .*
+      - feature/*
+    drafts: true
+    enabled: true
   collapse_walkthrough: false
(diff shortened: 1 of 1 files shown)


.coderabbit.yaml
```diff
@@ -11,6 +11,8 @@ reviews:
   auto_review:
     enabled: true
     drafts: true
+    base_branches:
+      - .*
```
arkid15r: Doesn't this mean all branches?
The goal is to enable it for `feature/` branches while keeping it unchanged for `main`.

arkid15r: Please also order this properly.

ahmedxgouda: The variable is for the base branches other than `main`. `main` is included in the auto review by default.
image

arkid15r (approved):

PR #2208, opened 2025-09-04
Unify user summary card

### LINKED ISSUE #2167: Create unified UserSummary card
### Description
Combine the **User summary block** and the **contribution heat map** into a single block.  
The combined block should **not have any title**.

### Proposed Layout 
- **Avatar on the left**: approx. 1/3 of the block width.  
- **Summary text + heat map on the right**: stacked vertically, taking up the remaining 2/3.

_This layout idea is just a proposition and it might not work out, but we'd like to see how this looks like._

### Responsive Behavior
- On small screens, hide the heat map.  
- Only show the avatar and the summary text.

- [ ] Yes
- [x] No

Image

anurag2787:
Resolves #2167 

## Description
This PR combines the User summary block and the contribution heat map into a single block.

## Key Changes
- **Left side:** User avatar (approx. 1/3 of the card width).  
- **Right side:** Summary text on top, Contribution Heat Map below (approx. 2/3 of the card width).  
- Hided heat map in **small screens**.

image

image

## Checklist


anurag2787:
Hy @kasya I have combined the user summary and the contribution heat map into a single block could you please take a look and let me know if any further changes are needed or should I start writing tests for these updates?

kasya:
> Hy @kasya I have combined the user summary and the contribution heat map into a single block could you please take a look and let me know if any further changes are needed or should I start writing tests for these updates?

@anurag2787 Looks good so far! 

Could we move the avatar to the left and use the rest of the space to make the map and make it a bit larger (250x250)?
This way we can free more space for the heat map. The screenshot only has the avatar move aside, but the heat map size/position is not updated ⬇️ 
Screenshot 2025-09-04 at 6 13 05 PM

Would be good to have the map a bit larger... 🤔 We also need to start hiding the map earlier - if you look closely on screen size change the heat map image gets cut off. 

Also, can we make the avatar round? 

These are just some initial thoughts! Basically, I just don’t want any space to go to waste - I’d like us to use it as best as we can.
Thank you!

anurag2787:
Hy @kasya I have updated as per your suggestion can you please review Thanks!

Screenshot from 2025-09-07 19-02-51

kasya:
@anurag2787 this is great! 👍🏼 
One more suggestion/question: can we add padding/margin to move summary text a bit to the right to kinda match with the map? We'll need to only do this when we show the map, and remove on small screen sizes. I understand the image has some empty space, but it just looks weird when text is aligned with it. 
Screenshot 2025-09-07 at 11 04 31 AM

And if so, you can start working on the tests 👍🏼 thanks!

anurag2787:
Hy, I have fixed this according to the suggestion given.
image

arkid15r (dismissed):
It looks more compact now, thank you @anurag2787 👍

arkid15r (approved):

PR #2244, opened 2025-09-09
UI/ux mentorship program update

Rajgupta36:
## Proposed change

Resolves #(2194)

Media
image
image
image
image


frontend/src/components/ProgramCard.tsx
```diff
@@ -26,19 +47,30 @@ const ProgramCard: React.FC<ProgramCardProps> = ({ program, onEdit, onView, acce
...
+              <Tooltip content={program.name} placement="top" className="w-64">
+                <h3 className="line-clamp-2 h-12 overflow-hidden text-base font-semibold text-gray-600 dark:text-white">
+                  {program.name}
+                </h3>
+              </Tooltip>
+            ) : (
+              <h3 className="line-clamp-2 h-12 overflow-hidden text-base font-semibold text-gray-600 dark:text-white">
+                {program.name}
+              </h3>
+            )}
```
kasya: I don't really like this approach. This is one of the reasons why SonarCube is not happy with code duplication. 
Tooltip component from heroui has `isDisabled` property that we can use to not have this code duplication;
```suggestion
            <Tooltip content={program.name} placement="top" className="w-64" isDisabled={!isTitleLong}>
              <h3 className="line-clamp-2 h-12 overflow-hidden text-base font-semibold text-gray-600 dark:text-white">
                {program.name}
              </h3>
            </Tooltip>
```

frontend/src/components/ProgramCard.tsx
```diff
@@ -26,19 +47,30 @@ const ProgramCard: React.FC<ProgramCardProps> = ({ program, onEdit, onView, acce
...
-  const description =
-    program.description?.length > 100
-      ? `${program.description.slice(0, 100)}...`
-      : program.description || 'No description available.'
+  const description = program.description || 'No description available.'
+
+  const titleThresholds = { sm: 40, md: 60, lg: 80 }
+  const descriptionThresholds = { sm: 150, md: 200, lg: 250 }
+
+  const isTitleLong = program.name.length > titleThresholds[screenSize]
```
kasya: This doesn't work btw. Never saw a tooltip unless I hardcoded `true` here.
I'm not sure why we check for screen sizes to be honest. From my testing the card width doesn't change with the screen size. We might wrap items, but the card stays as is.

frontend/src/components/ProgramCard.tsx
```diff
@@ -47,14 +79,26 @@ const ProgramCard: React.FC<ProgramCardProps> = ({ program, onEdit, onView, acce
...
-          <div className="mb-2 text-xs text-gray-600 dark:text-gray-400">
+          <div className="mb-2 text-sm text-gray-600 dark:text-gray-400">
             {program.startedAt && program.endedAt
               ? `${formatDate(program.startedAt)} – ${formatDate(program.endedAt)}`
               : program.startedAt
                 ? `Started: ${formatDate(program.startedAt)}`
                 : 'No dates set'}
           </div>
-          <p className="mb-4 text-sm text-gray-700 dark:text-gray-300">{description}</p>
+          <div className="flex flex-1 flex-col justify-start">
```
kasya, line 65: Also, do you really think this looks ok? 🤔 
Screenshot 2025-09-09 at 7 28 04 PM

I feel like there's a huge need for some bottom padding

frontend/src/components/ProgramCard.tsx
```diff
@@ -47,14 +79,26 @@ const ProgramCard: React.FC<ProgramCardProps> = ({ program, onEdit, onView, acce
...
+            {isDescriptionLong ? (
+              <Tooltip content={description} placement="bottom" className="w-88">
```
kasya: Actually, let's remove showing full description on hover(keep truncation). That's just too much. Users can just click into the program to read the whole thing.

I do think, however, that maybe we should increase the whole card size and make it square.
this way it won't feel so crowded - right now it looks messy with all the information, labels.. And I actually wanted to show more things here 😅 
With increase in size we'll probably only be able to fit about 3 cards per row.. But we can try that.

frontend/src/components/SingleModuleCard.tsx
```diff
@@ -99,12 +94,6 @@ const SingleModuleCard: React.FC<SingleModuleCardProps> = ({
 
           {dropdownOpen && (
             <div className="absolute right-0 z-20 mt-2 w-40 rounded-md border border-gray-200 bg-white shadow-lg dark:border-gray-700 dark:bg-gray-800">
-              <button
```
kasya, line 102: Hmm... We needed to hide the whole thing, there's no point to keep the three dots with broken dropdown (this is clicked into/active state):
Screenshot 2025-09-09 at 7 58 40 PM

frontend/src/app/mentorship/programs/[programKey]/page.tsx
Comment (kasya) [inline frontend/src/app/mentorship/programs/[programKey]/page.tsx:1]: Can you also update the `save` button styling on creating the program? 
Screenshot 2025-09-09 at 8 01 15 PM

it should match that button color that you updated in this pr - `text-[#1D7BD7]`

frontend/src/app/my/mentorship/programs/[programKey]/page.tsx
Comment (kasya) [inline frontend/src/app/my/mentorship/programs/[programKey]/page.tsx:1]: Also, weirdly enough Create module button for saving has a different styling too. 
Screenshot 2025-09-09 at 8 03 34 PM

Can we unify look of all of these?

kasya (changes requested):
Left comments ⬇️

frontend/src/components/ProgramForm.tsx
```diff
@@ -218,7 +218,7 @@ const ProgramForm = ({
                 <button
                   type="submit"
                   disabled={loading}
-                  className="flex items-center justify-center gap-2 rounded-md border-1 border-[#0D6EFD] px-4 py-2 text-[#0D6EFD] hover:bg-[#0D6EFD] hover:text-white dark:text-sky-600 dark:hover:bg-sky-100"
+                  className="flex items-center justify-center gap-2 rounded-md border border-[#1D7BD7] bg-transparent px-6 py-2 whitespace-nowrap text-[#1D7BD7] transition-all hover:bg-[#1D7BD7] hover:text-white dark:hover:text-white"
```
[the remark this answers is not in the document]
Rajgupta36: any suggestion @kasya

Rajgupta36:
this is updated preview
Screenshot from 2025-09-10 12-26-57
Screenshot from 2025-09-10 13-51-00
Screenshot from 2025-09-10 14-01-16
Screenshot from 2025-09-10 14-01-38
Screenshot from 2025-09-10 14-12-41
Screenshot from 2025-09-10 14-13-15

frontend/src/components/ProgramForm.tsx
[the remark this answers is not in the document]
kasya: @Rajgupta36 we can keep it as is for now 👌🏼

PR #2287, opened 2025-09-16
fix params and remove refresh params

### LINKED ISSUE #2274: Technical debt: clean up refresh param handling in Program/Module forms
Currently, the form update logic for `Program` or `Module` is messy, relying on passing `refresh=true` as a parameter. This implementation exposes internal mechanics to the end user, which they should never see—especially on slower connections where the param might briefly appear. It also contributes to code duplication check failures.

Apollo provides built-in support for updating local frontend data after mutations ([docs](https://www.apollographql.com/docs/react/data/mutations?utm_source=chatgpt.com#updating-local-data)). There are several ways to handle this more cleanly than our current approach.

I’m particularly in favor of **updating the cache directly** ([docs](https://www.apollographql.com/docs/react/data/mutations?utm_source=chatgpt.com#updating-the-cache-directly)) to avoid unnecessary backend calls, but we need to ensure this only happens after the mutation completes successfully. This may require updating the mutation’s returned data.

Alternatively, we could define a custom `update()` function ([docs](https://www.apollographql.com/docs/react/data/mutations?utm_source=chatgpt.com#the-update-function)) if modifying the returned data isn’t desirable.

Rajgupta36:
## Proposed change

Resolves #2274 

Description
- [x] Removed refreshs param
- [x] update direct apollo cache


kasya:
@Rajgupta36 I believe this should go into your feature branch and not into main? 🤔

frontend/src/app/my/mentorship/programs/[programKey]/page.tsx
```diff
@@ -93,28 +110,11 @@ const ProgramDetailsPage = () => {
...
-        setModules(data.getProgramModules || [])
-      }
+    if (data?.getProgram) {
+      setProgram(data.getProgram)
+      setModules(data.getProgramModules || [])
     }
-
-    processResult()
-  }, [shouldRefresh, data, refetch, router, searchParams])
+  }, [data])
```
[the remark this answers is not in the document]
Comment (Rajgupta36) [reply frontend/src/app/my/mentorship/programs/[programKey]/page.tsx:91]: it maybe hide the last known data during transient states.

frontend/package.json
```diff
@@ -19,10 +19,10 @@
   },
   "dependencies": {
     "@apollo/client": "^3.14.0",
-    "@fortawesome/fontawesome-svg-core": "^7.0.1",
```
kasya, line 22: Why did you downgraded these? And all package.json changes 🤔

frontend/src/app/my/mentorship/programs/[programKey]/edit/page.tsx
```diff
@@ -103,7 +104,41 @@ const EditProgramPage = () => {
         status: formData.status,
       }
 
-      await updateProgram({ variables: { input } })
+      await updateProgram({
```
Comment (kasya) [inline frontend/src/app/my/mentorship/programs/[programKey]/edit/page.tsx]: We don't need to do this for edit mutations. 
Since you've added `id` to GraphQL queries - this now works as is. This falls under this [first point](https://www.apollographql.com/docs/react/data/mutations#include-modified-objects-in-mutation-responses) in cache update options.

frontend/src/app/my/mentorship/programs/[programKey]/page.tsx
```diff
@@ -77,7 +69,32 @@ const ProgramDetailsPage = () => {
             status: newStatus,
           },
         },
-        refetchQueries: [{ query: GET_PROGRAM_AND_MODULES, variables: { programKey } }],
+        update: (cache, { data: mutationData }) => {
```
Comment (kasya) [inline frontend/src/app/my/mentorship/programs/[programKey]/page.tsx]: Same here - not needed, as long as GraphQL query has `id`

frontend/src/app/my/mentorship/programs/[programKey]/modules/create/page.tsx
```diff
@@ -94,7 +94,32 @@ const CreateModulePage = () => {
         mentorLogins: parseCommaSeparated(formData.mentorLogins),
       }
 
-      await createModule({ variables: { input } })
+      await createModule({
+        variables: { input },
+        update: (cache, { data: mutationData }) => {
```
Comment (kasya) [inline frontend/src/app/my/mentorship/programs/[programKey]/modules/create/page.tsx:99]: For create logic we still need to have custom update module because as stated [here](https://www.apollographql.com/docs/react/data/mutations#include-modified-objects-in-mutation-responses:~:text=Returning%20modified%20objects,update%20function.) auto-update would not work with any new objects in a list (for example, list of modules in a program)

Rajgupta36:
> Hey Raj!
> 
> Thanks for working on this. I noticed that you added `id`s to GraphQL queries and it was the main change we needed for this to work.
> 
> I reverted changes in files for editing programs and modules - this does not need any custom logic now, since query has ID to match with cache on response. Next time please read docs more closely - it specifically describes this situation [here](https://www.apollographql.com/docs/react/data/mutations#include-modified-objects-in-mutation-responses:~:text=Returning%20modified%20objects,update%20function.)
> 
> We still need it for creating items though, so good call there.
> 
> I also left a question about changes in `package.json` file. And noticed that you didn't run `make check` since there were no generated types in your PR. Please follow our contributing guidelines - you are not a newbie here 👌🏼

Hi @kasya,
I was working on another PR: https://github.com/OWASP/Nest/pull/2288 .
Sorry for not marking it as a draft. Once I finish that PR, I’ll rebase this one. Thanks!

frontend/package.json
Rajgupta36: may force push cause   conflicts. i am resolving and sync it with feature/mentorship-portal

Rajgupta36: maybe changing base cause problem. i'm solving it

frontend/src/components/ModuleCard.tsx
```diff
@@ -76,9 +75,13 @@ const ModuleItem = ({ details }: { details: Module }) => {
 
   return (
     <div className="flex h-46 w-full flex-col gap-3 rounded-lg border-1 border-gray-200 p-4 shadow-xs ease-in-out hover:shadow-md dark:border-gray-700 dark:bg-gray-800">
-      <ActionButton onClick={handleClick}>
+      <button
+        type="button"
+        onClick={handleClick}
+        className="text-start font-semibold text-blue-400 hover:underline"
+      >
```
kasya: Why is this a button? 😮 Somehow missed that before..
This should be just a `Link` element. Works the same and the navigation is actually faster. 
```
<Link
    href={`${window.location.pathname}/modules/${details.key}`}
    className="text-start font-semibold text-blue-400 hover:underline"
  >
    <TruncatedText text={details?.name} />
</Link>
```

kasya (changes requested):
Left one comment! The rest looks fine and works as expected 👍🏼

kasya (approved):
Nice 💪🏼

PR #2299, opened 2025-09-20
refactor : added PLC0415 ruff config for specific files

### LINKED ISSUE #2297: Narrow down PLC0415 global ignore to per-file-ignores in pyproject.toml
**Description:**

Currently, our Ruff configuration globally ignores `PLC0415` https://docs.astral.sh/ruff/rules/import-outside-top-level/

We should narrow this down by removing the global ignore and specifying **only the files that actually need it** via `per-file-ignores` in `pyproject.toml`. This will keep the rule enabled everywhere else while still allowing necessary exceptions.

This will help us:

* Maintain stricter linting rules across the project.
* Prevent accidental top-level import violations.
* Improve long-term code health and maintainability.

**Acceptance Criteria:**

* Identify all files currently triggering `PLC0415`.
* Update `pyproject.toml` to include `per-file-ignores` for those files.
* Remove the global `PLC0415` ignore from the Ruff configuration.
* Confirm Ruff passes cleanly with the new configuration.

- [ ] Yes
- [x] No

ishanBahuguna:
## Proposed change
1. Removed `PLC0415` from the global ignore.  
2. Ran `ruff check` to identify the files triggering `PLC0415`.  
3. Found 51 errors across 28 files.  
4. Added the following per-file ignore in `pyproject.toml` for all 28 files:

Resolves #2297 

## Checklist

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -68,20 +68,21 @@ line-length = 99
 lint.select = [ "ALL" ]
 lint.extend-select = [ "I" ]
 lint.ignore = [
-  "ANN",     # https://docs.astral.sh/ruff/rules/#flake8-annotations-ann/
-  "ARG002",  # https://docs.astral.sh/ruff/rules/unused-method-argument/
-  "C901",    # https://docs.astral.sh/ruff/rules/complex-structure/
-  "COM812",  # https://docs.astral.sh/ruff/rules/missing-trailing-comma/
-  "D407",    # https://docs.astral.sh/ruff/rules/missing-dashed-underline-after-section/
-  "DJ012",   # https://docs.astral.sh/ruff/rules/django-unordered-body-content-in-model/
-  "FIX002",  # https://docs.astral.sh/ruff/rules/line-contains-todo/
-  "PLC0415", # https://docs.astral.sh/ruff/rules/import-outside-top-level/
+  "ANN",    # https://docs.astral.sh/ruff/rules/#flake8-annotations-ann/
+  "ARG002", # https://docs.astral.sh/ruff/rules/unused-method-argument/
+  "C901",   # https://docs.astral.sh/ruff/rules/complex-structure/
+  "COM812", # https://docs.astral.sh/ruff/rules/missing-trailing-comma/
+  "D407",   # https://docs.astral.sh/ruff/rules/missing-dashed-underline-after-section/
+  "DJ012",  # https://docs.astral.sh/ruff/rules/django-unordered-body-content-in-model/
+  "FIX002", # https://docs.astral.sh/ruff/rules/line-contains-todo/
+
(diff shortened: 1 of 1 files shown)


backend/pyproject.toml
```diff
@@ -68,20 +68,21 @@ line-length = 99
...
 lint.ignore = [
-  "ANN",     # https://docs.astral.sh/ruff/rules/#flake8-annotations-ann/
-  "ARG002",  # https://docs.astral.sh/ruff/rules/unused-method-argument/
-  "C901",    # https://docs.astral.sh/ruff/rules/complex-structure/
-  "COM812",  # https://docs.astral.sh/ruff/rules/missing-trailing-comma/
-  "D407",    # https://docs.astral.sh/ruff/rules/missing-dashed-underline-after-section/
-  "DJ012",   # https://docs.astral.sh/ruff/rules/django-unordered-body-content-in-model/
-  "FIX002",  # https://docs.astral.sh/ruff/rules/line-contains-todo/
-  "PLC0415", # https://docs.astral.sh/ruff/rules/import-outside-top-level/
+  "ANN",    # https://docs.astral.sh/ruff/rules/#flake8-annotations-ann/
```
arkid15r: Please keep the rest unchanged, revert the formatting changes.

backend/pyproject.toml
```diff
@@ -120,6 +121,92 @@ lint.per-file-ignores."**/tests/**/*.py" = [
...
+lint.per-file-ignores."apps/slack/common/handlers/users.py" = [
+  "PLC0415", # https://docs.astral.sh/ruff/rules/import-outside-top-level/
+]
+lint.per-file-ignores."apps/slack/events/member_joined_channel/contribute.py" = [
+  "PLC0415", # https://docs.astral.sh/ruff/rules/import-outside-top-level/
+]
+lint.per-file-ignores."apps/slack/utils.py" = [
+  "PLC0415", # https://docs.astral.sh/ruff/rules/import-outside-top-level/
+]
+lint.per-file-ignores."tests/apps/ai/management/commands/ai_update_chapter_chunks_test.py" = [
```
arkid15r: You need to verify that all non top-level imports are justified. The idea is to minimize them to must have cases only. This one, for example, looks like a good candidate for the removal.

ishanBahuguna: @arkid15r  Sorry for the inefficient approach earlier.
Just to confirm—should I go ahead and move the imports to the top of the files (in cases where having them inside classes/functions isn’t justified), and then remove PLC0415 from the ignore list?

arkid15r: Yes, for the cases where it's just an easy move to the top -- please do it. No all cases can be fixed though, so you will still have to keep some per file ignores, mostly for slack app.

arkid15r:
Resolved in https://github.com/OWASP/Nest/pull/3899

PR #2352, opened 2025-10-02
Fix slack and duplication errors

### LINKED ISSUE #2342: Fix the errors while syncing context and chunks
There are duplication errors while syncing context and chunks

Dishant1804:
## Proposed change

Resolves #2342 

fixed the erros of slack and integrity errors of the chunks 

## Checklist

### DIFF
--- a/backend/apps/ai/common/base/chunk_command.py
+++ b/backend/apps/ai/common/base/chunk_command.py
@@ -20,7 +20,7 @@ class BaseChunkCommand(BaseAICommand):
     def process_chunks_batch(self, entities: list[Model]) -> int:
         """Process a batch of entities to create or update chunks."""
         processed = 0
-        batch_chunks_to_create = []
+        batch_chunks_to_create = {}
         content_type = ContentType.objects.get_for_model(self.model_class)
 
         for entity in entities:
@@ -68,7 +68,9 @@ class BaseChunkCommand(BaseAICommand):
                     openai_client=self.openai_client,
                     save=False,
                 ):
-                    batch_chunks_to_create.extend(chunks)
+                    for chunk in chunks:
+                        key = (chunk.context_id, chunk.text)
+                        batch_chunks_to_create[key] = chunk
                     processed += 1
                     self.stdout.write(
                         self.style.SUCCESS(f"Created {len(chunks)} new chunks for {entity_key}")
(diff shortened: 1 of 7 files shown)


backend/apps/ai/common/base/chunk_command.py
```diff
@@ -77,7 +79,21 @@ def process_chunks_batch(self, entities: list[Model]) -> int:
                 self.stdout.write(f"Chunks for {entity_key} are already up to date.")
 
         if batch_chunks_to_create:
-            Chunk.bulk_save(batch_chunks_to_create)
+            context_ids = {context_id for context_id, _ in batch_chunks_to_create}
```
arkid15r: Any reason to not using (again) our backend-wide approach for handling models data -- model.update_data + model.bulk.save instead?

backend/apps/ai/common/base/context_command.py
```diff
@@ -35,10 +35,12 @@ def process_context_batch(self, entities: list[Model]) -> int:
             ):
                 processed += 1
                 entity_key = self.get_entity_key(entity)
-                self.stdout.write(f"Created context for {entity_key}")
+                self.stdout.write(f"Created/updated context for {entity_key}")
             else:
                 entity_key = self.get_entity_key(entity)
```
arkid15r, line 40: This line is repetitive.

arkid15r (changes requested):

backend/apps/ai/common/base/chunk_command.py
Dishant1804: I'm using bulk save, the integrity error was due to duplication -- I had 2 options one was to do it while checking in this way and other was to make the ignore_conflicts = true in the bulk save

backend/apps/ai/common/base/context_command.py
Dishant1804: fixed

backend/apps/ai/common/base/chunk_command.py
arkid15r: You already have a check in the Chunk::update_data method. Why it's not enough and why it can't be handled there?

Dishant1804: the update_data method will check for duplicates but the integrity error is caused due to race condition, while updating we are deleting the necessary chunks which are to be updated

so to avoid the race condition we are doing this
ignore_conflicts=true in the bulk_save was an option too I read about it on stackoverflow -- but I dont think that is viable
what do you think?

backend/apps/ai/common/base/chunk_command.py
```diff
@@ -58,12 +58,14 @@ def process_chunks_batch(self, entities: list[Model]) -> int:
                     continue
 
                 chunk_texts = Chunk.split_text(full_content)
-                if not chunk_texts:
+                unique_chunk_texts = list(dict.fromkeys(chunk_texts))
```
arkid15r: Could you explain why you need this fancy stuff instead of using set? Does the order matter here?

Dishant1804: no the order does not matter -- now I am using set it is fixed

arkid15r: Would just `set(chunk_texts)` work here without converting it to a list? It doesn't look right at all from performance perspective.

Dishant1804: completely forgot that the text order will matter when we are passing it to `create_chunks_and_embeddings` so we will need to use `unique_chunk_texts = list(dict.fromkeys(chunk_texts))` which was our implementation before the set

Dishant1804: the openAI embeddings only accepts list[str] or str as an input so we need to convert it to list

reference doc : https://platform.openai.com/docs/api-reference/embeddings/create?utm_source=chatgpt.com

arkid15r: I wouldn't be so sure about that -- https://github.com/openai/openai-python/blob/d69edeb39e6a9cc8d9822a2838b10ab4102b4cc6/src/openai/resources/embeddings.py#L50

PR #2374, opened 2025-10-06
Nestbot AI Direct message implementation

### LINKED ISSUE #2189: Implement Slack Direct Messaging and Mentions for NestBot AI Assistant

We need to implement a conversational interface for NestBot inside Slack, allowing users to interact with the NestBot AI Assistant via direct messages and mentions in channels. This feature will serve as the primary way for OWASP community members to get AI-powered responses directly in Slack.

**Additional context**

Image

Dishant1804:
## Proposed change

Resolves #2189 

- [x] added slack direct message feature  

## Checklist


backend/apps/slack/models/chat.py
```diff
@@ -0,0 +1,90 @@
...
+
+from django.db import models
+from django.utils import timezone
+
+from apps.common.models import TimestampedModel
+from apps.slack.models.member import Member
+from apps.slack.models.workspace import Workspace
+
+
+class Chat(TimestampedModel):
```
arkid15r: Isn't this redundant with `Conversation` model we already have?

backend/apps/slack/models/chat.py
```diff
@@ -0,0 +1,90 @@
...
+            chat = Chat.objects.get(user=user, workspace=workspace)
+        except Chat.DoesNotExist:
+            chat = Chat(user=user, workspace=workspace, is_active=True)
+
+        if save:
+            chat.save()
+
+        return chat
+
+    def add_to_context(self, user_message: str, bot_response: str | None = None) -> None:
```
arkid15r: These methods need to be migrated to conversation + message models we already have. Please don't `+` strings in Python.

arkid15r (changes requested):
Please use existing Slack app entities:

backend/apps/slack/models/chat.py
Dishant1804: the conversation is for the slack channel conversation -- this chat model is responsible for handling the human and bot interaction 

this is a good approach becaue if we try to combine this with conversation -- unnecessary complexity will be added and it will be difficult for future developer to figure out what exactly is happening, as there would be a lot of conditions because the slack direct messages and the slack channel messages behave differently

let me know what do want to continue with

backend/apps/slack/models/chat.py
Dishant1804: this is the chat model responsible for human and bot conversation, so I think we should add and get the context from here 

in case we want to migrate let me know the specific model conversation or message

backend/apps/slack/models/chat.py
arkid15r: If you check Slack [conversation object](https://docs.slack.dev/reference/objects/conversation-object) you'll see that it covers more cases than just channel conversation. I don't want introduce new redundant models for each specific case we may have. Unless Slack splits it on their side. 

Your perspective for future developer's complexity issues may be biased as it's your code, you wrote it and now need to defend your decisions. I suggest reconsidering it and not to multiply entities when you can keep it unified.

>  the slack direct messages and the slack channel messages behave differently

It'd be great to see some specific examples.

Dishant1804: I'll refactor my approach and use the conversation model

backend/apps/slack/models/chat.py
Dishant1804: fixed

backend/apps/slack/models/chat.py
Dishant1804: refactored the code --- we are now using our existing conversation model

PR #2399, opened 2025-10-08
Add entity_channel field to Reminder model and remove channel_id

### LINKED ISSUE #2390: Update Reminder model channel field
- Update `Reminder` model to have `EntityChannel` model as a field instead of channel_id.
- Update slack, nest handlers and schedulers that uses the channel id of `Reminder` model.
- Update tests.

- [x] Yes
- [ ] No

ahmedxgouda:
## Proposed change

Resolves #2390

Add the PR description here.

## Checklist

- [ ] I've read and followed the [contributing guidelines](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md).

### DIFF
--- a/backend/tests/apps/nest/handlers/calendar_events_test.py
+++ b/backend/tests/apps/nest/handlers/calendar_events_test.py
@@ -1,30 +1,46 @@
 """Test cases for Nest Calendar Events handlers."""
 
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 from django.core.exceptions import ValidationError
 from django.utils import timezone
 
-from apps.nest.handlers.calendar_events import schedule_reminder, set_reminder
-from apps.nest.models.google_account_authorization import GoogleAccountAuthorization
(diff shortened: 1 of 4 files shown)


TheCyberLeader:
Hey @ahmedxgouda nice work on this refactor overall. 

I noticed there's still an open issue flagged by CodeRabbit that looks blocking: the conversation lookup in calendar_events.py uses Conversation.objects.get(name=args.channel.lstrip("#"), workspace__slack_workspace_id=workspace_id), but Slack slash commands typically deliver a channel ID (e.g., C12345) rather than a name. This means the lookup will raise Conversation.DoesNotExist and reminders will always fail with something like "Channel 'C12345' does not exist in this workspace."

CodeRabbit suggested supporting both by filtering on slack_channel_id OR name__iexact as a fallback, worth addressing before this lands so it doesn't break the existing command behavior.

ahmedxgouda:
Hey @TheCyberLeader. Thank you for reviewing. As I remember, rabbit's suggestion was wrong, and the functionality is working as expected, but this PR is worth testing again after we merge the updates from main (it was last updated 4 months ago).

PR #2533, opened 2025-11-02
Improved Snapshot Generation

### LINKED ISSUE #1988: Improve snapshot generation

For chapters get new items method returns entities filtered by `is_active` status. However, it's not enough as sometimes repositories can still be empty.

The snapshot processing job needs to utilize `active_chapters` manager instead of using custom filter.
The same approach has to be applied for all kind of entities (projects, releases, etc)

- [ ] Yes
- [x] No

**Additional context**

Image

anurag2787:
Resolves #1988 

## Description
This PR improves how we generate snapshots by making the handling of active entities more consistent. Earlier, different parts of the code used different ways to filter chapters by is_active, and some didn’t properly handle cases where repositories were empty
Now, we’ve standardized the process using the manager patterns active_chapters active_projects active_releases

## Checklist


anurag2787:
@arkid15r for new_issue if we use OpenIssueManager that is predefined but it only returns issues from the last 90 days so So instead, can we directly use this ?
```
new_issues = Issue.objects.filter(
                created_at__gte=snapshot.start_at,
                created_at__lte=snapshot.end_at,
                state="open",
                repository__project__isnull=False,
            ).select_related("repository")
   ```

**Review by ahmedxgouda — CHANGES_REQUESTED** [review]
Please run `make check-test` first

Image

**Comment by arkid15r** [comment]
> @arkid15r for new_issue if we use OpenIssueManager that is predefined but it only returns issues from the last 90 days so So instead, can we directly use this ?
> 
> ```
> new_issues = Issue.objects.filter(
>                 created_at__gte=snapshot.start_at,
>                 created_at__lte=snapshot.end_at,
>                 state="open",
>                 repository__project__isnull=False,
>             ).select_related("repository")
> ```

The current naming/implementation is bad. Open issues manager should filter by state only. The dates should be handled separately by `.recent_issues()` or similar manager.
We can add it as you suggested but the issue managers need to be improved.

File: backend/apps/github/models/managers/user.py [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+    def get_queryset(self):
+        """Get queryset of users excluding bots and system accounts.
+
+        Filters out bot users and system accounts to ensure valid user data
+        for snapshot processing.
+        """
+        return (
+            super()
+            .get_queryset()
+            .select_related("organization")
```
ahmedxgouda: This is not correct and it is giving this error: 

Image

ahmedxgouda (changes requested):
Please ensure that the command is working before requesting a review.

backend/apps/github/models/managers/user.py
anurag2787: I have fixed this now sorry for the inconvenience

ahmedxgouda (dismissed):
LGTM

anurag2787:
Hi @kasya @arkid15r The PR is ready for review.
One more request could you please move this issue from GSoC 2025 to GSoC 2026?
Thanks!

backend/apps/github/models/user.py
```diff
@@ -28,6 +29,9 @@
 class User(NodeModel, GenericUserModel, TimestampedModel, UserIndexMixin):
     """User model."""
 
+    objects = models.Manager()
+    snapshot_users = SnapshotUserManager()
```
arkid15r: The naming could have been better here -- focus on issue traits, not specific context where you need it.

backend/apps/owasp/management/commands/owasp_process_snapshots.py
```diff
@@ -66,37 +66,25 @@ def process_snapshot(self, snapshot: Snapshot) -> None:
...
-                is_pre_release=False,
+            new_releases = Release.active_releases.filter(
+                created_at__gte=snapshot.start_at,
+                created_at__lte=snapshot.end_at,
             )
-            new_users = self.get_new_items(
-                User,
-                snapshot.start_at,
-                snapshot.end_at,
+            new_users = User.snapshot_users.filter(
```
arkid15r: You can notice this discrepancy here, active_ vs snapshot_ items.

arkid15r (changes requested):

anurag2787:
Hi @arkid15r i have updated the naming Could you please review and let me know if any further changes required

PR #2573, opened 2025-11-08
refactor: update breadcrumbs component to support dynamic title rende…

### LINKED ISSUE #1425: Improve breadcrumbs component

Kunal1522:
## Tasks

- [x] Improve the breadcrumbs component to display each project/chapter/user/etc page H1 header content
- [x] Added Fallback to BreadCrumbWrapper in case of some rendering issue
- [x] Removed the `/repositories` to appear in breadcrumb as the page does not really exist
- [x] Added Unit Testing for `BreadCrumbs.tsx`, `BreadCrumbsWrapper.tsx`

Resolves #1425 

---

## Proposed change

Display actual entity names in breadcrumbs instead of URL slugs.

Currently, breadcrumbs show URL-formatted text like "Bjornkimminich" instead of the real name "Björn Kimminich" from the database. This PR fixes that by having detail pages pass their GraphQL data to the breadcrumb system through a simple interface.

Pages now pass entity names like this:

The breadcrumb checks which field is set and displays that name. If no data is provided, it falls back to formatting the URL slug nicely.

Also refactored breadcrumb logic into reusable components (`BreadCrumbs`, `BreadCrumbsWrapper`, `useBreadcrumbs` hook) and added 72 tests.

---

## Detailed PR

**Problem**
Breadcrumbs were showing URL slugs instead of actual entity names from the database.

**Solution**
1.  **Created Reusable Breadcrumb System**
    Three main pieces:
    * `BreadCrumbs.tsx` - Pure UI component that renders breadcrumb items.
    * `BreadCrumbsWrapper.tsx` - Renders breadcrumbs. It takes an array of items and displays them nicely with proper styling and navigation links.
    * `useBreadcrumbs` hook - Contains all the logic for building breadcrumbs. It handles both simple routes (`/projects/zap`) and nested routes (`/organizations/x/repositories/y`).

2.  **Pages Pass Real Names via Props**
    Each detail page already fetches GraphQL data. Now they pass the entity name to `PageLayout`:

    The interface accepts different entity types:

    The hook checks which field is present and uses it. If none is provided, it falls back to formatting the URL slug.

3.  **Special Handling for Nested Routes**
    Repository routes (`/organizations/[org]/repositories/[repo]`) get special detection. The breadcrumb displays actual org and repo names, and skips the "Repositories" segment since that page doesn't exist.

**Testing**
Added 72 tests covering rendering, auto-generation, real name display, fallbacks, and edge cases. All pass.

### BEFORE

image

### AFTER

image

### BEFORE

image

### AFTER

image

---

## Checklist


Kunal1522:
@coderabbitai generate docstrings

frontend/src/hooks/useBreadcrumbs.ts
```diff
@@ -0,0 +1,64 @@
...
+export function useBreadcrumbs(breadcrumbData?: BreadcrumbData): BreadCrumbItem[] {
+  const pathname = usePathname()
+  const breadcrumbs: BreadCrumbItem[] = [{ title: 'Home', path: '/' }]
+
+  if (!pathname) return breadcrumbs
+
+  const segments = pathname.split('/').filter(Boolean)
+
+  const isNestedRepoRoute =
+    segments.length === 4 && segments[0] === 'organizations' && segments[2] === 'repositories'
```
kasya: Where is this `4` coming from? What if we have more then 4 nested items? 🤔 
We want to make it as dynamic as possible

frontend/src/components/BreadCrumbs.tsx
```diff
@@ -32,35 +27,27 @@ export default function BreadCrumbs() {
...
-          {segments.map((segment, index) => {
-            const href = homeRoute + segments.slice(0, index + 1).join(homeRoute)
-            const label = upperFirst(segment).replaceAll('-', ' ')
-            const isLast = index === segments.length - 1
+          {items.map((item, index) => {
+            const isLast = index === items.length - 1
 
             return (
-              <BreadcrumbItem key={href} isDisabled={isLast}>
+              <HeroUIBreadcrumbItem key={item.path}>
```
kasya, line 34: HeroUIBreadcrumbItem doesn't use isDisabled prop, but `BreadCrumbsWrapper.tsx`on line 58 does, causing inconsistent behavior. 🤔 

Think on adding `isDisabled={isLast}` to match BreadCrumbsWrapper

frontend/src/hooks/useBreadcrumbs.ts
```diff
@@ -0,0 +1,64 @@
+import upperFirst from 'lodash/upperFirst'
+import { usePathname } from 'next/navigation'
+
+export interface BreadCrumbItem {
+  title: string
+  path: string
+}
+
+interface BreadcrumbData {
```
kasya: `BreadcrumbData` interface is duplicated in two files, violating DRY. 
It's added here and in `PageLayout.tsx`. 

Also, we are trying to get away from using `interface` and want to use `type` instead. Please move this into `types/` folder and use where needed to avoid code duplication

frontend/src/hooks/useBreadcrumbs.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const segments = pathname.split('/').filter(Boolean)
+
+  const isNestedRepoRoute =
+    segments.length === 4 && segments[0] === 'organizations' && segments[2] === 'repositories'
+
+  if (isNestedRepoRoute) {
+    breadcrumbs.push({ title: 'Organizations', path: '/organizations' })
+
+    const orgTitle = breadcrumbData?.orgName || upperFirst(segments[1]).replaceAll('-', ' ')
```
kasya: I can see you use the same logic several times. Can we move this into a function instead? Lines 32, 38, 48.

frontend/src/components/BreadCrumbsWrapper.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { usePathname } from 'next/navigation'
+
+const ROUTES_WITH_PAGE_LAYOUT = [
+  /^\/members\/[^/]+$/,
+  /^\/projects\/[^/]+$/,
+  /^\/chapters\/[^/]+$/,
+  /^\/committees\/[^/]+$/,
+  /^\/organizations\/[^/]+$/,
+  /^\/organizations\/[^/]+\/repositories\/[^/]+$/,
+]
```
kasya: I'm also not super happy with these hardcoded values. If we ever change our project structure - this might fail. Is there a better way to find a match? 🤔

frontend/src/components/BreadCrumbsWrapper.tsx
```diff
@@ -0,0 +1,78 @@
+'use client'
+
+import { faChevronRight } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { Breadcrumbs, BreadcrumbItem as HeroUIBreadcrumbItem } from '@heroui/react'
+import upperFirst from 'lodash/upperFirst'
+import Link from 'next/link'
+import { usePathname } from 'next/navigation'
+
+const ROUTES_WITH_PAGE_LAYOUT = [
```
kasya: I also think this route detection logic is somewhat duplicated: regex patterns in here vs string matching in useBreadcrumbs 🤔

frontend/__tests__/unit/components/BreadCrumbsWrapper.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    test('returns null for project detail pages', () => {
+      ;(usePathname as jest.Mock).mockReturnValue('/projects/zap')
+
+      const { container } = render(<BreadCrumbsWrapper />)
+      expect(container.firstChild).toBeNull()
+    })
+
+    test('returns null for member detail pages', () => {
+      ;(usePathname as jest.Mock).mockReturnValue('/members/bjornkimminich')
```
kasya: Can we use users/project names that we already added to cspell to not add new items? Or just some fake ones, that would be even better.

kasya:
@Kunal1522 I pushed some changes to address issues I found while reviewing your PR. The one I noticed was not connected to your changes, but other 2 were from running `make check`.

Could you also address 5 newly added issues found by SonarQube? Here's the [list](https://sonarcloud.io/project/issues?id=OWASP_Nest&pullRequest=2573&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)

kasya:
@Kunal1522 Hi! Are you planning to address the comments in this PR? The deadline for this issue passed a while ago (Nov 8th). If the updates aren’t made within the next day, I’ll need to reassign the issue to someone else. Thanks!

Kunal1522:
Hi, I’m sorry for the late reply. I have my end-term exams going on right now, so I won’t be able to contribute until 25th November.

However I went through the PR comments and completed all the requested changes, including the SonarQube ones.

Before committing, the only part  I wanted to improve was  making the breadcrumbs more dynamic and easier to extend.

Right now, the only path that appears in the breadcrumb but does not actually exist is: organisations/[slug]/repositories. I had to hard-code it, but similar cases may be added in the future. 

One solution I found is to run a script at build time that walks through the app directory and stores all existing routes in a JSON file. This JSON can be checked whenever needed. This would help in two ways:

validating multi-level slugged routes,(so i would not need regex pattern matching )
preventing breadcrumbs for paths that do not actually exist (paths without a page.tsx or layout.tsx).

To make things more extensible, I also moved the configuration to a types file. So if more pages are added later, we can just update the types, and everything will adjust automatically.
image

Please assign this issue to someone else  if they volunteer  otherwise i will fix this after 26th November.

kasya:
@Kunal1522 I haven't reassigned this to anyone else. It's Nov 26th so let me know if you can get back to this tomorrow. Otherwise, I'll reassign! Hope the exams went well.
Thanks!

Kunal1522:
thanks i appreciate that .yeah i have started reworking on it now maybe will raise pr in 1-2 days

frontend/src/utils/breadcrumb.ts
kasya, line 1: I found one issue with this logic.

For Snapshots page our pathname will include the hyphen that we need to keep in the crumb or substitute with something else (like `/` for example)

Image

Kunal1522:
Hi i added a check where it detects the date and does not separate it.also  some new related test cases to it as well

Issue #2570, opened 2025-11-08
Minor documentation changes needed


cognitivegears:
**Describe the bug**

In `CONTRIBUTING.md` in the getting started steps there are a few things that did not work exactly as described. The commands to copy the .env file only work on bash-like shells (as well, the current step one are unnecessary in any case for each of them). Also, when signing up with Algolia, the steps as listed are slightly confusing based on the current sign-up process.

**To Reproduce**

Run the getting started steps using  `nushell` or other non-bash shell.

Sign up for Algolia as a new user, there is now an App automatically created during sign-up (although it does try to get you to run an import wizard, you can skip this step.) Also, in the documentation, checking for the permissions as written is not straightforward, as the permission isn't named the same.

**Expected behavior**

New developer successfully setup, including environment files copied and Algolia setup confirmed.

**Are you going to work on fixing this?**

- [x] Yes
- [ ] No

**Screenshots**

none

**Desktop (please complete the following information):**

- OS: MacOS Tahoe running nushell
- Browser: Zen (not applicable to issue)
- Version: 1.17.5b (not applicable to issue

**Smartphone (please complete the following information):**

**Additional context**

PoojasPatel013:
Is this issue still open to work?

PR #2585, opened 2025-11-10
docs: Minor onboarding clarifications

cognitivegears:
A couple of minor onboarding clarification updates. When I ran though the process, there were a few small differences in how I needed to run the setup:

    The method to copy the environment file only works on bash-like shells. Replacing it with just a general cp command has broader compatibility.

    The Algolia setup was slightly different than described. Updated the section based on what I saw during signup.

Fixes Issue: https://github.com/OWASP/Nest/issues/2570
Checklist

    [ X ] I've read and followed the [contributing guidelines](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md).
    [ X ] I've run make check-test locally; all checks and tests passed.


arkid15r (approved):
Thanks for improving that @cognitivegears

PR #2606, opened 2025-11-11
Extend Nest API `chapter` and `project`

### LINKED ISSUE #2569: Update Nest API `chapter` and `project`
`chapter` and `project` do not return the leader's names.

Update `chapter` and `project` endpoints to return the leader`s names.

- [x] Yes
- [ ] No

rudransh-shrivastava:
Resolves #2569

## Proposed change
- add `leaders` field.

## Checklist

### DIFF
--- a/.github/ansible/production/nest.yaml
+++ b/.github/ansible/production/nest.yaml
@@ -63,7 +63,7 @@
         ansible.builtin.command:
           cmd: crontab /tmp/production_crontab
 
-      - name: Retart services
+      - name: Restart services
         shell:
           cmd: docker compose up -d --pull always
 
--- a/.github/ansible/production/proxy.yaml
+++ b/.github/ansible/production/proxy.yaml
@@ -20,9 +20,9 @@
         dest: ~/docker-compose.yaml
         mode: '0644'
 
-    - name: Retart services
+    - name: Restart services
       shell:
-        cmd: docker compose up -d --pull always
+        cmd: docker compose up -d --pull always && docker compose restart
 
     - name: Prune docker images
       shell:
--- a/.github/ansible/staging/nest.yaml
+++ b/.github/ansible/staging/nest.yaml
@@ -67,7 +67,7 @@
         ansible.builtin.command:
           cmd: crontab /tmp/staging_crontab
 
-      - name: Retart services
+      - name: Restart services
         shell:
           cmd: docker compose up -d --pull always
 
--- a/.github/ansible/staging/proxy.yaml
+++ b/.github/ansible/staging/proxy.yaml
@@ -20,9 +20,9 @@
         dest: ~/docker-compose.yaml
         mode: '0644'
 
-    - name: Retart services
+    - name: Restart services
       shell:
-        cmd: docker compose up -d --pull always
+        cmd: docker compose up -d --pull always && docker compose restart
 
     - name: Prune docker images
       shell:
--- a/.github/workflows/run-ci-cd.yaml
+++ b/.github/workflows/run-ci-cd.yaml
@@ -43,7 +43,7 @@ jobs:
           python-version: '3.13'
 
       - name: Set up pre-commit cache
-        uses: actions/cache@v5
+        uses: actions/cache@9255dc7a253b0ccc959486e2bca901246202afeb
         with:
           path: ~/.cache/pre-commit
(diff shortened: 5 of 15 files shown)


backend/apps/api/rest/v0/chapter.py
```diff
@@ -38,8 +38,14 @@ class ChapterDetail(ChapterBase):
     """Detail schema for Chapter (used in single item endpoints)."""
 
     country: str
+    leaders: list[str]
     region: str
 
+    @staticmethod
+    def resolve_leaders(obj):
+        """Resolve leaders."""
+        return obj.leaders_raw
```
rudransh-shrivastava: @arkid15r Should I use `EntityMember` instead of `leaders_raw`?

rudransh-shrivastava: @arkid15r Should I use `EntityMember` instead of `leaders_raw`?

rudransh-shrivastava: 👀 duplicate comment.

arkid15r: Yes, we need to migrate to the EntityMember data

// please ignore the next comments -- I need to check something w/ coderabbit

arkid15r: @coderabbitai suggest 10 good first issues for backend and frontend parts of OWASP Nest. They should be relevant and easy to implement (1-3 points). Create just a single issue with the suggestions -- don't create issues for each suggestion.

backend/apps/api/rest/v0/chapter.py
```diff
@@ -41,8 +41,14 @@ class ChapterDetail(ChapterBase):
     """Detail schema for Chapter (used in single item endpoints)."""
 
     country: str
+    leaders: list[str]
     region: str
 
+    @staticmethod
+    def resolve_leaders(obj):
```
arkid15r, line 48: These are technically leader names (not leader objects).
I'm not sure how to handle compound objects within API yet (add minimum data + a separate endpoint vs adding all data).

rudransh-shrivastava: Yeah, I'll update it to `leader_names` for now. But I think it's wrong to name it `leader_names`. What do I do?

I found [Stripe API Docs / Expand](https://docs.stripe.com/expand) interesting. 
> The API has an Expand feature that allows you to retrieve linked objects in a single call, effectively replacing the object ID with all its properties and values. For example, say you wanted to access details on a customer tied to a given Checkout Session. You would retrieve the Checkout Session and pass the customer property to the expand array, which tells Stripe to include the entire Customer object in the response

rudransh-shrivastava: I'm not sure how to implement this, but I can look into it if we go down this path :)

backend/apps/api/rest/v0/project.py
```diff
@@ -40,6 +40,12 @@ class ProjectDetail(ProjectBase):
     """Detail schema for Project (used in single item endpoints)."""
 
     description: str
+    leader_names: list[str]
+
+    @staticmethod
+    def resolve_leader_names(obj):
+        """Resolve leader names."""
+        return [leader.member_name for leader in obj.entity_leaders]
```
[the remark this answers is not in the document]
rudransh-shrivastava: `entity_leaders` does not support prefetching.

backend/apps/api/rest/v0/chapter.py
arkid15r: Let's keep it simple and use compact leader representation objects: rename it to leaders and provide objects with member key and name -- should be enough for now.

backend/apps/api/rest/v0/project.py
```diff
@@ -40,6 +41,15 @@ class ProjectDetail(ProjectBase):
...
     description: str
+    leaders: list[Leader]
+
+    @staticmethod
+    def resolve_leaders(obj):
+        """Resolve leaders."""
+        return [
+            Leader(key=leader.member.login if leader.member else None, name=leader.member_name)
+            for leader in obj.entity_leaders
+        ]
```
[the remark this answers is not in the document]
rudransh-shrivastava: I don't think its worth it for this case.

backend/apps/api/rest/v0/chapter.py
rudransh-shrivastava: Got it!

PR #2710, opened 2025-11-24
Fix running e2e backend

### LINKED ISSUE #2709: Fix the running e2e backend instance
The e2e backend instance gives internal server error when opening anything.

ahmedxgouda:
## Proposed change

Resolves #2709

- Added db environment variables to the docker compose backend service.
- Added `csrf_decorate` that skips the csrf protection if the environment is `e2e`.
- Fixed rest API internal error.
- Updated the docs to elaborate how to update the e2e db data.

## Checklist

### DIFF
--- a/backend/apps/common/management/commands/dump_data.py
+++ b/backend/apps/common/management/commands/dump_data.py
@@ -9,12 +9,12 @@ from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 from psycopg2 import ProgrammingError, connect, sql
 
-DB = settings.DATABASES["default"]
-HOST = DB.get("HOST", "localhost")
-PORT = str(DB.get("PORT", "5432"))
-USERNAME = DB.get("USER", "")
-PASSWORD = DB.get("PASSWORD", "")
-NAME = DB.get("NAME", "")
+DEFAULT_DATABASE = settings.DATABASES["default"]
+DB_HOST = DEFAULT_DATABASE.get("HOST", "localhost")
+DB_PORT = str(DEFAULT_DATABASE.get("PORT", "5432"))
+DB_USER = DEFAULT_DATABASE.get("USER", "")
+DB_PASSWORD = DEFAULT_DATABASE.get("PASSWORD", "")
+DB_NAME = DEFAULT_DATABASE.get("NAME", "")
 
 
 class Command(BaseCommand):
@@ -39,49 +39,35 @@ class Command(BaseCommand):
                 "public.slack_conversations",
                 "public.slack_messages",
             ],
-            help=(
-                "Table pattern to include. "
-                "Defaults: public.owasp_*, public.github_*, public.slack_members, "
-                "public.slack_workspaces, public.slack_conversations, public.slack_messages."
-            ),
+            help="Table pattern to include",
         )
 
     def handle(self, *args, **options):
         output_path = Path(options["output"]).resolve()
-        tables = options["tables"] or []
         output_path.parent.mkdir(parents=True, exist_ok=True)
+        tables = options["tables"] or []
 
-        temp_db = f"temp_{NAME}"
         env = os.environ.copy()
-        env["PGPASSWORD"] = PASSWORD
-        self.stdout.write(self.style.NOTICE(f"Creating temporary database: {temp_db}"))
+        env["PGPASSWORD"] = DB_PASSWORD
+
+        temp_db = f"temp_{DB_NAME}"
         try:
-            # 1) Create temp DB from template
-            self._execute_sql(
-                "postgres",
-                [f"CREATE DATABASE {temp_db} TEMPLATE {NAME};"],
-            )
-            # 2) Get tables with email field
-            self.stdout.write(self.style.NOTICE("Fetching tables with email fields…"))
+            self._execute_sql("postgres", [f"CREATE DATABASE {temp_db} TEMPLATE {DB_NAME};"])
 
-            table_list = self._execute_sql(
-                temp_db,
-                [self._table_list_query()],
-            )
+            self.stdout.write(self.style.SUCCESS(f"Created temporary DB: {temp_db}"))
 
-            # 3) Hide email fields
-            self.stdout.write(self.style.NOTICE("Hiding email fields in temp DB…"))
-            self._execute_sql(temp_db, self._hide_emails_queries([row[0] for row in table_list]))
-            # 4) Dump selected tables
-            self.stdout.write(self.style.NOTICE(f"Creating dump at: {output_path}"))
+            table_list = self._execute_sql(temp_db, [self._table_list_query()])
+            self._execute_sql(temp_db, self._remove_emails([row[0] for row in table_list]))
+            self.stdout.write(self.style.SUCCESS("Removed emails from temporary DB"))
+
             dump_cmd = [
                 "pg_dump",
                 "-h",
-                HOST,
+                DB_HOST,
                 "-p",
-                PORT,
+                DB_PORT,
                 "-U",
-                USERNAME,
+                DB_USER,
                 "-d",
                 temp_db,
                 "--compress=9",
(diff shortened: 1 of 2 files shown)


ahmedxgouda:
I think I am assigned to the issue :) 
image

CONTRIBUTING.md
```diff
@@ -441,6 +441,34 @@ Follow these steps to setup your e2e testing environment:
 
 **Please note that you only need to do these steps once.**
 
+#### Updating e2e Test Data
```
arkid15r: My data file update comment was not about the instructions for the data loading. It was about how to maintain the `nest-e2e..sql.gz` file.

backend/apps/common/utils.py
```diff
@@ -67,6 +68,13 @@ def clean_url(url: str | None) -> str | None:
     return url.strip().rstrip(".,;:!?") or None
 
 
+def csrf_decorate(view):
```
arkid15r: Is there a way to avoid settings based on the running environment? We want it to be as close to production as possible.

docker-compose/e2e.yaml
```diff
@@ -12,6 +12,12 @@ services:
       db:
         condition: service_healthy
     env_file: ../backend/.env.e2e.example
+    environment:
```
arkid15r: The `env_file` values don't work here?

backend/apps/common/utils.py
ahmedxgouda: The last time I test sending fuzzing request in the fuzzing PR, it didn't work even with setting csrf header, but it was before a  dedicated backend instance for e2e. I will remove this for now and try again in the fuzzing PR.

backend/Makefile
```diff
@@ -66,6 +66,10 @@ dump-data-e2e:
 	@echo "Dumping Nest e2e data"
 	@CMD="pg_dumpall -U nest_user_e2e --clean | gzip -9 > backend/data/nest-e2e.sql.gz" $(MAKE) exec-db-command-e2e
 
+dump-data-local:
```
arkid15r: Let's look at this one more time. Are we going to use our full data snapshot for e2e testing (e.g. the entire data from `nest.json.gz`)? If so I'd prefer to keep it simple and have just a single data file -- without `nest-e2e.sql.gz` introduction.

backend/apps/api/decorators/cache.py
```diff
@@ -36,6 +36,8 @@ def cache_response(
     def decorator(view_func):
         @wraps(view_func)
         def _wrapper(request, *args, **kwargs):
+            if settings.IS_E2E_ENVIRONMENT:
```
arkid15r: Why e2e needs this change?

backend/apps/api/rest/v0/__init__.py
```diff
@@ -46,7 +46,7 @@
 }
 
 api_settings_customization = {}
-if settings.IS_LOCAL_ENVIRONMENT:
+if settings.IS_LOCAL_ENVIRONMENT or settings.IS_E2E_ENVIRONMENT:
```
arkid15r: I suggest creating a separate section for this.

backend/Makefile
ahmedxgouda: Maybe we can use `nest.sql.gz` instead of `nest.json.gz`. As it is extremely faster to load.

backend/apps/api/decorators/cache.py
ahmedxgouda: We didn't setup cache for e2e, so when it tries to access cache it gives 500 internal error. I think there is an option to setup redis cache in CI/CD. Maybe we can do that or keep it simple.

backend/Makefile
```diff
@@ -83,12 +68,18 @@ index-data:
 	@CMD="python manage.py algolia_update_synonyms" $(MAKE) exec-backend-command
 
 load-data:
+	@echo "Recreating Nest schema"
```
arkid15r: This script drops the DB meaning local settings like superuser login/password are lost.

backend/apps/api/decorators/cache.py
arkid15r: Yes, let's add cache service for e2e instead. The closer to production architecture the better -- for both local and CI/CD cases.

backend/data/dump.sh
arkid15r, line 1: This is not an implementation that we'd like to have for long term. I don't think we have any .sh scripts with more than 5-10 lines of code logic. I suggest migrating this to Python and adding tests as we normally do.

backend/data/dump.sh
```diff
@@ -0,0 +1,53 @@
...
+if [ -z "$UPDATES" ]; then
+    echo "No email fields found to hide."
+else
+    echo "Hiding email addresses…"
+    echo "$UPDATES" | psql -h "$DJANGO_DB_HOST" -U "$DJANGO_DB_USER" -d "$TEMP_DB"
+fi
+
+# 3. Dump the DB
+echo "Creating dump…"
+pg_dump -h "$DJANGO_DB_HOST" -U "$DJANGO_DB_USER" -d "$TEMP_DB" | gzip -9 > ./data/nest.sql.gz
```
arkid15r: It seems to create the entire DB dump while our current command dumps only what we need (github/owasp/slack apps data)

arkid15r (changes requested):

backend/apps/api/decorators/cache.py
ahmedxgouda: Ok I will add the cache in another PR after this one.

arkid15r: Let's configure the cache backed for e2e via Django settings (locmem for now)

backend/apps/common/management/commands/dump_data.py
```diff
@@ -0,0 +1,158 @@
...
+            "--output",
+            default=str(Path(settings.BASE_DIR) / "data" / "nest.sql.gz"),
+            help="Output dump path (default: data/nest.sql.gz)",
+        )
+        parser.add_argument(
+            "-t",
+            "--table",
+            action="append",
+            dest="tables",
+            default=["public.owasp_*", "public.github_*", "public.slack_*"],
```
arkid15r: We don't dump the entire slack app data atm.

backend/apps/common/management/commands/dump_data.py
arkid15r, line 1: This ai code is still difficult to maintain. Also there are not tests for this file.

backend/apps/common/management/commands/dump_data.py
```diff
@@ -0,0 +1,158 @@
...
+            except CalledProcessError:
+                # Best-effort cleanup
+                self.stderr.write(
+                    self.style.WARNING(f"Failed to drop temp DB {temp_db} (ignored).")
+                )
+
+    def _hide_emails(self) -> str:
+        # Uses a DO block to UPDATE every column named 'email' in non-system schemas
+        return """
+DO $$
```
arkid15r: Is there a more readable/maintainable way to get rid of email addresses?

backend/apps/common/management/commands/dump_data.py
```diff
@@ -0,0 +1,158 @@
...
+        SELECT quote_ident(n.nspname) AS schemaname,
+               quote_ident(c.relname) AS tablename,
+               quote_ident(a.attname) AS colname
+        FROM pg_attribute a
+        JOIN pg_class c ON c.oid = a.attrelid
+        JOIN pg_namespace n ON n.oid = c.relnamespace
+        WHERE a.attname = 'email'
+          AND a.attnum > 0
+          AND NOT a.attisdropped
+          AND n.nspname NOT IN ('pg_catalog','information_schema','pg_toast')
```
arkid15r: These magic names and pretty much the entire PL/SQL spell doesn't fit our clean and readable code approach. Yes, it's probably performs the best but it's not the only goal we have.

Issue #2713, opened 2025-11-24
Light Mode Icon Is Misleading and Confusing


dhirukumar:
Description

The current Light Mode icon used in the UI looks similar to a settings gear instead of a sun or light indicator. This causes confusion because users may think the icon opens settings rather than toggling the theme.

To Reproduce

Go to any page in the application.
Locate the theme toggle button (top-right of the navbar).
Observe the Light Mode icon (gear-like sun shape).
See that it resembles a settings/gear icon instead of a light mode indicator.

Expected behavior

The Light Mode icon should clearly indicate theme switching, not settings.
A correct icon should look like a sun, light bulb, or recognizable light theme symbol.

Screenshots
Old (Misleading) Icon

<img width="311" height="46" alt="Image" src="https://github.com/user-attachments/assets/96a64c87-9998-4884-9580-a8a733e70e2e" />

New (Recommended) Icon

<img width="417" height="51" alt="Image" src="https://github.com/user-attachments/assets/b644aa31-a572-4f38-8398-e2c81ca5299e" />

(Already fixed in PR: change-lightmode-icon)

dhirukumar:
hii @arkid15r  @kasya  can you plese assign me this because i already make this changes

kasya:
@dhirukumar Hi!
This is actually a sun icon in FA. 
What icon are you proposing?

dhirukumar:
Hi @kasya,
This is the faLightbulb icon — it looks very similar to the Font Awesome lightbulb icon, so it aligns well with standard UI patterns and avoids the confusion caused by the previous one.

kasya:
@dhirukumar oh, sorry! Missed that the second screenshot had a new icon already in the description 😅 
Ok, let's try the bulb icon. What if we do `solid` icon? This would suggest it is ON because of the white fill 🤔 So that would match the light theme more... But it's up to you to play around with those.

PR #2742, opened 2025-11-27
Changed light mode icon because it was misleading

dhirukumar:
Description

This PR updates the Light Mode toggle icon in the UI.

What Was Changed

Replaced the old sun/gear-style icon (which looked like a settings icon):
Added the new light bulb icon (more intuitive for Light Mode):

before
Screenshot 2025-11-27 at 5 09 47 PM

After
Screenshot 2025-11-27 at 5 09 37 PM

🧠 Why This Change

The previous icon was visually misleading and confused users as it closely resembled the settings icon.
The new icon provides clearer UX and better represents Light Mode functionality.

🧪 Testing

Verified UI behavior in both Light and Dark modes
Ensured toggle works with no functional regressions

📝 Checklist

I've read and followed the [contributing guidelines](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md)
I've run make check-test locally; all checks and tests passed.

Closes https://github.com/OWASP/Nest/issues/2713


kasya:
@dhirukumar why did you close previous PR? 🤔 This makes it much harder to review since reviewer leaves comments to be addressed. 

I'll leave this one be for this first time, but next time I'll just close this duplicate.
Please don't close open PRs "to keep it clean".

dhirukumar:
Hi @kasya , sorry for the inconvenience caused.

I closed the previous PR because I was also working on the Skeleton feature at the same time, and switching branches with stashes caused some files to get mixed into the PR that weren’t related to the actual change. While running make check-test, some extra files were added for making the check test pass, and I got a bit stuck.
To avoid confusion, I thought it would be cleaner to start again from scratch on a new branch — but I now understand this made the review harder.

Thank you for pointing it out.
I won’t close active PRs next time, and I’ll make sure to keep branches clean and focused on a single change.
Sorry again

sameersharmadev:
Hey! Not sure if it’s my place to say 😅, but a light-bulb icon can be a bit confusing here since in many modern UIs it usually represents tips/help/insights.
Maybe we can try using a cleaner sun icon instead of the previous one so it doesn't look like a cogwheel? Maybe something like the one on Lucide. I think a Sun/Moon is more universal for theme toggles.
Do let me know what you guys think.

kasya:
@sameersharmadev I agree! After reviewing this PR I decided to revert the lighbulb and try something else to keep universal Sun/Moon design.

@dhirukumar I know your original suggestion was to use lightbulb, but it is confusing to have a lightbulb and the moon. I tried to play around with changing style of the lightbulb for light/dark theme (on/off lightbulb) but did not really like it.

So I decided to introduce custom icons via SVG. We will still rely on FA icons for most cases, but this should be a good fallback when we can not find suitable FA icon. 

I pushed some changes to this PR. Hope you don't mind! We'll still count it as completed by you since you've put a good effort in it.

dhirukumar:
@kasya ,Thanks for the update!
To be honest, even I felt from the beginning that the lightbulb icon wasn’t a 100% perfect fit — but after going through all the FA icons, I couldn’t find a suitable sun icon that matched well with the moon. I also considered introducing an external icon, but I wasn’t sure if mixing one FA icon and one external icon would be the right approach.

If you believe adding the custom SVG icons improves the consistency, then that’s great — I’m happy with it. I also prefer having a proper sun icon, so introducing external SVGs makes sense to me.

theAnuragMishra:
Why dont we just switch to a better icon set? md icons on react-icons offer some better fits.

arkid15r:
> Why dont we just switch to a better icon set? md icons on react-icons offer some better fits.

Please create a separate issue for that -- we can definitely consider a better unified solution if it exists.
Thank you!

PR #2762, opened 2025-12-01
Use markdown-it for mentorship portal summaries

kasya:
Update to use markdown text for Program and Module summary/description.

## Checklist

### DIFF
--- a/frontend/src/app/globals.css
+++ b/frontend/src/app/globals.css
@@ -147,6 +147,12 @@
     color: #1d7bd7;
   }
 
+  a .md-wrapper ul,
+  a .md-wrapper ol,
+  a .md-wrapper li {
+    color: inherit;
+  }
+
   .navlink {
     color: inherit;
   }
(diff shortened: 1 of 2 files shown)


frontend/src/components/MarkdownWrapper.tsx
kasya: `DOMPurify.sanitize()` acts as a security barrier that strips out all malicious content before it reaches the DOM. Even though markdown-it is configured with `html: true` (which allows raw HTML in the markdown to pass through), DOMPurify removes any dangerous elements like `<script>` tags, event handlers (onclick, etc.), and other XSS vectors before the HTML is rendered via dangerouslySetInnerHTML. 

This means users can't inject executable JavaScript or other harmful code—DOMPurify ensures only safe HTML makes it to the page.

arkid15r (approved):

arkid15r (approved):

Issue #2781, opened 2025-12-03
Improve frontend/src/app/settings/** test coverage


arkid15r:
**Describe the solution you'd like**

Add tests for better coverage of src/app/settings/** -- then remove the coverage [exclusion](https://github.com/OWASP/Nest/blob/cad6e7ba66b09c4f8a66acb18c0892336471fe68/frontend/jest.config.ts#L11)

**Are you going to work on implementing this?**

- [ ] Yes
- [x] No

Swyamk:
hey @arkid15r i would love to work on this issue , can u assign it to me.

Swyamk:
Hey @arkid15r, I had a quick question about this issue. Apart from enabling coverage, should I also extend the existing tests to cover more edge cases and failure scenarios? Just wanted to confirm the expected scope before proceeding. Thanks!

arkid15r:
> Hey [@arkid15r](https://github.com/arkid15r), I had a quick question about this issue. Apart from enabling coverage, should I also extend the existing tests to cover more edge cases and failure scenarios? Just wanted to confirm the expected scope before proceeding. Thanks!

Yes, that's the idea 👍

Shofikul-Isl4m:
I would like to work on this.  can you assign this to me?

PR #2788, opened 2025-12-03
Enable coverage for API Keys page and improve deterministic test quality

Swyamk:
## Summary

This PR enables Jest coverage for the `src/app/settings/**` directory so that the existing API Keys page tests are now properly included in the frontend coverage reports. While the API Keys page already had good test coverage, those tests were being ignored due to a coverage exclusion in the Jest configuration.

Along with fixing the coverage issue, the API Keys test suite itself has been improved to be more deterministic, easier to maintain, and more reliable overall. As a result, frontend coverage now accurately reflects the work that was already in place.

Fixed: #2781 

---

## Improvements Made

All suggested improvements have now been applied:

### 1. Cleaned up toast mocking
- Removed repeated usage of `require('@heroui/toast')` across multiple tests
- Imported `addToast` once at the top of the file instead
- Simplified toast-related assertions and reduced duplication

### 2. Removed a duplicate validation test
- Deleted the duplicate test that checked the 101-character name case
- Strengthened the remaining validation test by also asserting that the mutation is **not** called on validation failure:
  ```ts
  expect(mockCreateMutation).not.toHaveBeenCalled()

### DIFF
--- a/frontend/__tests__/unit/pages/ApiKeysPage.test.tsx
+++ b/frontend/__tests__/unit/pages/ApiKeysPage.test.tsx
@@ -76,7 +76,9 @@ describe('ApiKeysPage Component', () => {
     })
   }
 
-  const setupMocks = (overrides = {}, mutationLoading = { create: false, revoke: false }) => {
+  const setupMocks = (overrides = {}, mutationLoading?: { create: boolean; revoke: boolean }) => {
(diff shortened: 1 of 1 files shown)


Swyamk:
@kasya  Could you please review the PR when you get a chance? I believe the implementation is complete from my end

kasya (approved):
@Swyamk thanks for adding these!

Issue #2892, opened 2025-12-13
Extend nest_update_badges.py to handle OWASP project leader badges via EntityMember model


arkid15r:
**Is your feature request related to a problem? Please describe.**
Currently, the backend management command `nest_update_badges.py` only handles badge assignment for users with the OWASP staff role. Project leaders, despite their significance, do not receive the appropriate "OWASP project leader" badge through this logic. This limits recognition for project leadership roles in the community.

**Describe the solution you'd like**
Enhance `nest_update_badges.py` so that it assigns project leader badges for all users in the database with the project leader role. This should use the `EntityMember` model, identifying all records where `role = LEADER`. Badge assignment should cover every project leader present in the database, ensuring the badge is displayed accordingly.

**Are you going to work on implementing this?**

- [ ] Yes
- [x] No

**Additional context**
- The system already assigns badges for staff via this command; logic for project leaders should be added in a similar pattern.
- Related models: `EntityMember` with the `LEADER` role value.
- Ensure future extensibility for other roles as well.

Swyamk:
@arkid15r I would like to work on this issue. Can u please assign it to me .

mrkeshav-05:
@arkid15r , Can I work on this.

Shofikul-Isl4m:
could you please assign this to me?

HarshitVerma109:
@arkid15r 
Could you please assign this issue to me?

PR #2894, opened 2025-12-13
feat: Add automatic Project Leader badge assignment

Swyamk:
## Summary

This PR extends the existing badge synchronization logic to recognize **OWASP Project Leaders**.  
Previously, the `nest_update_badges` command only handled OWASP Staff badges, so project leaders were not automatically awarded a badge.

With this change, active and reviewed project leaders now receive a **Project Leader** badge automatically, and the badge is removed if the user is no longer a leader. The implementation follows the same pattern as the existing staff badge logic and is safe to run multiple times.

Fixed: #2892 

## Changes Made

- Added `update_project_leader_badge()` to `nest_update_badges.py`
- Uses the `EntityMember` model to identify users with the `LEADER` role
- Filters only active and reviewed members
- Creates the **Project Leader** badge if it does not exist
- Assigns badges to eligible users and removes them from former leaders
- Extended backend tests to cover project leader badge assignment and idempotency

### DIFF
--- a/backend/apps/nest/management/commands/nest_update_badges.py
+++ b/backend/apps/nest/management/commands/nest_update_badges.py
@@ -2,12 +2,14 @@
 
 import logging
 
+from django.contrib.contenttypes.models import ContentType
 from django.core.management.base import BaseCommand
 
 from apps.github.models.user import User
 from apps.nest.models.badge import Badge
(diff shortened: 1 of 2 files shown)


arkid15r:
Sorry, I don't have time for [cosplay code](https://github.com/arkid15r/cosplay-coding).

PR #2898, opened 2025-12-14
imporove test covarage to 100 percent for EntityActions.tsx

### LINKED ISSUE #2850: Improve test covarage of EntityActions.tsx
current test covarage for frontend/src/components/EntityActions.tsx is low . So have to increase its test covarage

Shofikul-Isl4m:
## Proposed change

Resolves #2850
image

Improved test covarage to 100 percent for EntityActions.tsx

## Checklist

- [x] I read and followed the [contributing guidelines](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md)
- [x] I ran `make check-test` locally and all tests passed

### DIFF
--- a/frontend/__tests__/unit/components/EntityActions.test.tsx
+++ b/frontend/__tests__/unit/components/EntityActions.test.tsx
@@ -456,7 +456,8 @@ describe('EntityActions', () => {
     it('prevents event propagation on button click', () => {
       const mockParentClick = jest.fn()
       render(
-        <div role="button" tabIndex={0} onClick={mockParentClick} onKeyDown={() => {}}>
+        // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
+        <div onClick={mockParentClick}>
           <EntityActions type="program" programKey="test-program" />
         </div>
       )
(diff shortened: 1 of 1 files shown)


kasya (changes requested):
@Shofikul-Isl4m Hi!
Please address Sonar comment. There are 2 new [issues](https://sonarcloud.io/project/issues?id=OWASP_Nest&pullRequest=2898&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true) introduced in this PR.

Image

Shofikul-Isl4m:
@kasya  Hi,
The Issue: For testing event bubbling, I need a wrapper element around EntityActions to capture clicks.

SonarCloud flags a Code Smell if I use `<div role="button">`, suggesting I use a native `<button>` or `<input>` instead.
I cannot use `<button>` because EntityActions renders a button itself, and nesting buttons is invalid HTML that breaks React tests.I cannot use `<input>` because it is a void element and cannot contain children.

If I use a plain div with onClick (to satisfy SonarCloud and HTML validity), ESLint throws accessibility errors
(jsx-a11y/click-events-have-key-events).

Proposed Solution: Is it acceptable to use a plain div with onClick and disable the ESLint accessibility rules for these specific test lines?
// eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
`<div onClick={mockParentClick}>`
  `<EntityActions ... />`
`</div>`
Reasoning: The accessibility rules ensure clickable elements work for keyboard/screen reader users. However, this div is a test-only helper that never renders in the real app, so no real user will ever interact with it. Disabling the rule here is safe and resolves the conflict.

Shofikul-Isl4m:
@kasya so should i do this?

kasya:
> @kasya so should i do this?

@Shofikul-Isl4m yes, for this particular case I feel like that's fine, you're right! Feel free to update this 👌🏼

PR #2901, opened 2025-12-14
Fixed naming convention error for totalCount

### LINKED ISSUE #2887: Rename this field "totalCount" to match the regular expression ^[_a-z][_a-z0-9]*$
**Describe the bug**
So in backend/tests/apps/github/management/commands/github_sync_user_test.py. There is field named as totalCount, it will be good to use camelCase expression if the backend was in javascript but its in python and according to python's naming convention using snake_case which will support the required regular expression ^[_a-z][_a-z0-9]*$ & will be more efficient. The camelCased totalCount violates this rule.

**To Reproduce**

1. Go to 'backend/tests/apps/github/management/commands/github_sync_user_test.py'
2. rename totalCount to total_count

**Expected behavior**

- [x] Yes
- [ ] No

**Additional Context:**
https://sonarcloud.io/project/issues?impactSeverities=LOW&issueStatuses=OPEN%2CCONFIRMED&id=OWASP_Nest&open=AZp-i-mzHhyenAko7tNZ

anirudhprmar:
## Proposed change

Resolves #2887 

The problem was totalCount didn't matched the required regular expression ^[_a-z][_a-z0-9]*$, meaning it wasn't snake_case as python supports that. When trying to solve for this issue, i renamed all the files using totalCount as field name but when i ran make check-test, i was getting some additional unnecessary error. I went deep and found out that PyGithub utilization in repository.test file was giving run time errors as according to docs PyGithub intentionally uses camelCase for totalCount to match GitHub's JSON response patterns, even though it breaks Python conventions.

To solve this there were only 2 ways:
1. either to keep the total_count that'll fix the linting issue and keep totalCount too, for PyGithub compatibility. 
2. use getattr(commits, 'totalCount', 0) dynamically wherever we've used PyGithub library. 

So, I implemented the first solution. Willing to hear suggestions on what should be done, keep it as it as or go with getattr().

## Checklist

- [X] I read and followed the [contributing guidelines](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md)
- [X] I ran `make check-test` locally and all tests passed

### DIFF
--- a/backend/apps/slack/common/handlers/users.py
+++ b/backend/apps/slack/common/handlers/users.py
@@ -72,7 +72,6 @@ def get_blocks(
         ),
     ]
 
-    blocks = []
     for idx, user in enumerate(users):
         user_name_raw = user.get("idx_name") or user.get("idx_login", "")
         user_name = truncate(escape(user_name_raw), presentation.name_truncation)
--- a/backend/tests/apps/slack/common/handlers/users_test.py
+++ b/backend/tests/apps/slack/common/handlers/users_test.py
@@ -37,8 +37,9 @@ class TestGetUsersBlocks:
         """Tests the happy path, ensuring user data is formatted correctly into blocks."""
         mocker.patch("apps.github.index.search.user.get_users", return_value=mock_users_data)
         blocks = get_blocks(search_query="john")
-        assert len(blocks) > 1
-        user_block_text = blocks[0]["text"]["text"]
+        user_block_text = blocks[1]["text"]["text"]
+
+        assert "OWASP users that I found" in blocks[0]["text"]["text"]
         assert "1. <https://github.com/johndoe|*John Doe*>" in user_block_text
         assert "Company: OWASP" in user_block_text
         assert "Location: San Francisco" in user_block_text


anirudhprmar:
> @anirudhprmar Please address all CodeRabbit comments

yeah, on it

anirudhprmar:
@kasya, the changes coderabbit suggests revert back to the original implementation, which makes this entire issue irrelevant. Also i noticed that changing totalCount to total_count is creating more problems and errors. This time i even tried to implement use getattr(commits, 'totalCount', 0) but this still fails the make check-tests

I don't think this issue is a good fix. 
Would really appreciate some feedback on this one.

anirudhprmar:
> @kasya, the changes coderabbit suggests revert back to the original implementation, which makes this entire issue irrelevant. Also i noticed that changing totalCount to total_count is creating more problems and errors. This time i even tried to implement use getattr(commits, 'totalCount', 0) but this still fails the make check-tests
> 
> I don't think this issue is a good fix. Would really appreciate some feedback on this one.

@arkid15r can you look into this issue. I don't think we should proceed with this issue as it's not helping a lot instead adding unnecessary complexities.

arkid15r (approved):

PR #2914, opened 2025-12-14
Enhance Precise Location Manipulation on Chapters Map

### LINKED ISSUE #2895: Enhance Precise Location Manipulation on Chapters Map
Currently, on the chapters map in the frontend, precise location manipulation has several usability quirks:
- When hovering over interactive buttons, focus on the map is lost and the overlay (`Click to interact with map`) reappears, interrupting the experience.
- Disabling the shared location mode does not remove the Pin displaying the precise location.
- Enabling shared location mode repeatedly can result in multiple pins being added, which is only visible when fully zoomed in.

- Ensure that button hover does not cause loss of map focus and does not show `Click to interact with map` button
- Disabling shared location mode should fully remove the precise location pin from the map (ideally - reset it)
- Activating shared location mode multiple times should never create more than one pin representing the precise location, regardless of zoom level or interaction count (should work if we address prev. point with reset)
- Make sure this behavior is present on all pages where we have maps (chapters, home page, individual chapter's page, etc).

- [ ] Yes
- [x] No

**Additional context**
This enhancement will improve UX for chapter organizers and users manipulating their location accurately on the map, making interactions predictable and free of visual clutter or interruptions.

anurag2787:
Resolves #2895

## Proposed change
This PR fixes the following issues with precise location on Chapter maps
- Fixed the issue when hovering over interactive buttons, focus on the map is lost and the overlay (Click to interact with map) reappears
- Fixed the behavior so disabling shared location mode fully removes and resets the precise location pin from the map.

## Checklist

- [x] I read and followed the [contributing guidelines](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md)
- [x] I ran `make check-test` locally and all tests passed

### DIFF
--- a/backend/tests/apps/github/management/commands/github_sync_user_test.py
+++ b/backend/tests/apps/github/management/commands/github_sync_user_test.py
@@ -16,7 +16,7 @@ class MockPaginatedList:
 
     def __init__(self, items):
         self._items = items
-        self.totalCount = len(items)
+        self.totalCount = len(items)  # NOSONAR (for consistency with PyGithub).
 
     def __iter__(self):
         """Return an iterator for the items."""


frontend/src/components/CardDetailsPage.tsx
```diff
@@ -176,6 +176,7 @@ const DetailsCard = ({
               <ChapterMapWrapper
                 geoLocData={geolocationData}
                 showLocal={true}
+                showLocationSharing={true}
```
kasya, line 179: This didn't work for me on an individual chapter's page. It's still showing my hardcoded IP location instead of the precise geolocation.

Image

frontend/src/app/community/snapshots/[id]/page.tsx
```diff
@@ -138,6 +138,7 @@ const SnapshotDetailsPage: React.FC = () => {
             <ChapterMapWrapper
               geoLocData={snapshot.newChapters}
               showLocal={false}
+              showLocationSharing={true}
```
Comment (kasya) [inline frontend/src/app/community/snapshots/[id]/page.tsx:141]: This also doesn't work for me right now. It just refreshes the chapter without any movement to a precise location

kasya (changes requested):
@anurag2787 Thanks for working on these updates so quickly!
A couple of things I noticed:
1. Precise location button works differently on chapters page and on the home page. On chapters page resetting shared location moves back to a general view. On the home page, however, it stays in that zoomed in local view (as if I'm still on the precise location view, but without the pin).

2 more I shared in comment below 🔽

anurag2787:
The issue has been fixed
[Screencast from 2025-12-15 14-08-20.webm](https://github.com/user-attachments/assets/985076b9-4721-449a-b6ba-51b4745b9200)

frontend/src/components/CardDetailsPage.tsx
anurag2787: Hy @kasya for individual chapter it actually focusing on the individual chapter location but currently fixed it so when user share location it will properly shows both the user location pin and the chapter pins

PR #2934, opened 2025-12-16
Avoid duplicate backend image builds in Docker Compose

### LINKED ISSUE #2927: Update docker compose to build the backend image only once
The backend and worker services use the same build context and they build the same image in the same time.

We should build the image once only in one service and reuse the image in the other service. We should do this in all environments (i.e. local, production, and staging).

- [ ] Yes
- [x] No

**Additional context**

anurag2787:
Resolves #2927 

## Proposed change
Build the backend image once and used the same image for the worker as both services use the same code so building it twice is unnecessary.

## Checklist

### DIFF
--- a/docker-compose/local.yaml
+++ b/docker-compose/local.yaml
@@ -6,7 +6,7 @@ services:
         python manage.py migrate &&
         python manage.py runserver 0.0.0.0:8000
       '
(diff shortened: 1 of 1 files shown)


arkid15r (approved):
Looks good, let's try this 👍

PR #3007, opened 2025-12-22
updated mentorship_update_comments.py to not sync unlabeled modules

### LINKED ISSUE #2860: Update `mentorship_update_comments` Command to Skip Modules Without Labels
The `mentorship_update_comments` Django management command currently syncs comments for all modules, regardless of whether they have labels attached.  

To improve efficiency and prevent unnecessary processing, we need to update the command so that **modules without labels are skipped** during the sync.

---

### **Requirements**
- Update the `mentorship_update_comments` command logic:
  - Identify modules that do not have any labels attached.
  - Skip comment syncing for these modules.
- Ensure that modules with labels continue to sync comments as usual.
- Add any necessary logging or messages to indicate skipped modules (optional but helpful for debugging).
- Ensure that the command continues to work as expected for all other modules.

kart-u:
## Proposed change

Resolves #2860 

Added a condition to skip sync for unlabeled modules.

## Checklist

### DIFF
--- a/backend/apps/mentorship/management/commands/mentorship_update_comments.py
+++ b/backend/apps/mentorship/management/commands/mentorship_update_comments.py
@@ -41,13 +41,6 @@ class Command(BaseCommand):
         self.stdout.write(self.style.SUCCESS("Starting mentorship issue processing job..."))
 
         modules_with_labels = published_modules.exclude(labels=[]).select_related("project")
-        modules_without_labels = published_modules.filter(labels=[])
-
-        if modules_without_labels.exists():
-            for module in modules_without_labels:
-                self.stdout.write(
-                    self.style.WARNING(f"Skipping. Module '{module.name}' has no labels.")
-                )
 
         if not modules_with_labels.exists():
             self.stdout.write(


ahmedxgouda (dismissed):
LGTM

kasya (changes requested):
@kart-u Thanks for getting this done so fast!
Do you think there might be a cleaner or more performant solution?

kart-u:
hello @kasya 
Thanks for pointing that out, we can skip modules without labels right at the beginning instead of checking later 
I will now push updated code

backend/apps/mentorship/management/commands/mentorship_update_comments.py
```diff
@@ -40,7 +40,22 @@ def process_mentorship_modules(self) -> None:
...
 
-        for module in published_modules:
+        modules_with_labels = published_modules.exclude(labels=[]).select_related("project")
+        modules_without_labels = published_modules.filter(labels=[])
+
+        if modules_without_labels.exists():
+            for module in modules_without_labels:
+                self.stdout.write(
+                    self.style.WARNING(f"Skipping. Module '{module.name}' has no labels.")
+                )
```
kasya: I think there's still a better solution...

Do we really need to see this printed out? Do we care about this info at all? 🤔 This will require more computation and will slow down sync a little bit. Or a lot, depending on number of modules available.

kart-u: Yes you are absolutely right , I had some doubts about this as well. It was included to satisfy a issue requirement, but if the output isn’t useful , I’m fine with removing it

kasya (approved):

PR #3089, opened 2025-12-30
Prevent crash from invalid theme class during initialization

### LINKED ISSUE #3082: Local dev crash: InvalidCharacterError caused by non-CSS-safe theme value
**Describe the bug**
When running OWASP Nest locally, the frontend crashes during initial render with an `InvalidCharacterError`.
The error occurs because a theme value containing spaces (e.g. `"Default Theme"`) is applied as a CSS class, which is not valid.

This issue appears only in local development and prevents the UI from rendering.

**To Reproduce**

1. Set up the project locally following the documentation
2. Run the application using `make run` or Docker Compose local setup
3. Open http://localhost:3000
4. The app crashes before rendering

**Expected behavior**
The application should render normally in local development, even if an unexpected or invalid theme value exists.

- [x] Yes
- [ ] No

**Screenshots**
N/A (runtime error occurs before UI renders)

**Desktop (please complete the following information):**

- OS: Windows (WSL2 Ubuntu)
- Browser: Chrome
- Version: Latest

**Smartphone (please complete the following information):**

- N/A

**Additional context**
- The production site (https://nest.owasp.org/) works as expected
- This issue appears to be limited to local development / contributor onboarding
- Improving this would help new contributors avoid setup blockers

SuyashJain17:
## Proposed change

Resolves #3082

This PR prevents a runtime crash caused by invalid theme values being applied as CSS class names during initial render.

`next-themes` applies the persisted theme value directly to the `class` attribute. If the stored value contains spaces (for example `"Default Theme"`), this results in an `InvalidCharacterError` and the app crashes before rendering.

The fix validates persisted theme values and falls back safely to a CSS-safe default during initialization.

## Checklist

- [x] **Required:** I read and followed the contributing guidelines
- [ ] **Required:** I ran `make check-test` locally and all tests passed
- [x] I used AI for code, documentation, or tests in this PR

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -416,6 +416,82 @@ If you are adding new functionality, include relevant test cases.
 
 ## Contributing Workflow
 
+The following diagram illustrates the complete contribution workflow:
+
+```mermaid
+flowchart TD
+    Start([Start]) --> CreateIssue[Create New Issue]
+    Start --> FindIssue[Find Existing Issue]
+    CreateIssue --> GetAssigned["**Get Assigned to Issue**<br/>PRs will be automatically<br/>closed if you're not assigned"]
+    FindIssue --> GetAssigned
+    GetAssigned --> ResolveIssue[**Resolve Issue**<br/>work on code/docs/tests updates]
+
+    ResolveIssue --> RunChecks{**Run `make check-test`**<br/>locally! This is a required step -- you will not be assigned to new issues if you ignore this}
+    RunChecks -->|Fails| WP1[ ]
+    RunChecks -->|Passes| PushChanges[**Push Changes to<br/>GitHub Fork Branch**]
+    WP1 -.-> ResolveIssue
+
+    PushChanges --> HasPR{PR Exists?}
+    HasPR -->|No| CreateDraftPR[Create Draft PR]
+    HasPR -->|Yes| WaitAutoChecks[**Wait for Automated<br/>Checks to Finish**]
+    CreateDraftPR --> WaitAutoChecks
+
+    WaitAutoChecks --> CheckAutoTools{All **CodeRabbit and <br/>SonarQube** Comments<br/>Resolved?}
+    CheckAutoTools -->|No| MarkDraft[Make Sure PR Is **Marked as a Draft**]
+    CheckAutoTools -->|Yes| MarkReady[Mark PR as Ready<br/>for Review]
+    MarkDraft --> WP2[ ]
+    WP2 -.-> ResolveIssue
+
+    MarkReady --> RequestReview[Request Review from<br/>Project Maintainers]
+    RequestReview --> WaitMaintainer[Wait for Maintainers'<br/>Comments]
+
+    WaitMaintainer --> HasMaintainerComments{**Maintainers' Comments<br/>Resolved**?}
+    HasMaintainerComments -->|No| MarkDraft
+    HasMaintainerComments -->|Yes| CheckCI{**CI/CD<br/>Passing?**}
+
+    CheckCI -->|Yes| ReadyMerge([PR Ready for Merge])
+    CheckCI -->|No| MarkDraft
+
+    style Start fill:#4caf50,stroke:#2e7d32,stroke-width:2px,color:#ffffff
+    style ReadyMerge fill:#4caf50,stroke:#2e7d32,stroke-width:2px,color:#ffffff
+    style ResolveIssue fill:#ff9800,stroke:#f57c00,stroke-width:2px,color:#000000
+    style RunChecks fill:#2196f3,stroke:#1565c0,stroke-width:2px,color:#ffffff
+    style CheckAutoTools fill:#2196f3,stroke:#1565c0,stroke-width:2px,color:#ffffff
+    style HasMaintainerComments fill:#2196f3,stroke:#1565c0,stroke-width:2px,color:#ffffff
+    style CheckCI fill:#2196f3,stroke:#1565c0,stroke-width:2px,color:#ffffff
+    style MarkDraft fill:#ff9800,stroke:#f57c00,stroke-width:2px,color:#000000
+    style MarkReady fill:#4caf50,stroke:#2e7d32,stroke-width:2px,color:#ffffff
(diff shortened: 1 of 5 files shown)


SuyashJain17:
Thanks for the feedback. Agreed that forcing the theme breaks ModeToggle.

I’ll update the PR to validate persisted theme values and preserve theme switching instead of forcing a single theme.

SuyashJain17:
I noticed a low-severity SonarCloud warning suggesting `globalThis.window`.
I kept `typeof window === 'undefined'` here as it’s the standard SSR pattern
in Next.js and aligns with the CodeRabbit feedback.

Please let me know if you’d prefer a different approach.

frontend/src/wrappers/provider.tsx
```diff
@@ -7,22 +8,51 @@ import { ThemeProvider as NextThemesProvider } from 'next-themes'
...
+    const stored = localStorage.getItem('__nest_theme__')
+
+    if (stored && VALID_THEMES.includes(stored as ValidTheme)) {
+      return stored as ValidTheme
+    }
+
+    if (stored) {
+      localStorage.removeItem('__nest_theme__')
+    }
+  } catch {
```
arkid15r: I don't think it's a good idea.

frontend/src/wrappers/provider.tsx
```diff
@@ -7,22 +8,51 @@ import { ThemeProvider as NextThemesProvider } from 'next-themes'
 import React, { Suspense } from 'react'
 import apolloClient from 'utils/helpers/apolloClient'
 
-// <AppInitializer> is a component that initializes the Django session.
-// It ensures the session is synced with Django when the app starts.
-// AppInitializer is mounted once. Its job is to call useDjangoSession(),
-// which syncs the GitHub access token (stored in the NextAuth session) with the Django session.
+const VALID_THEMES = ['dark', 'light'] as const
```
arkid15r: Don't remove existing comments.

arkid15r: Why do you need const twice here?

arkid15r (changes requested):
I don't think I understand the code and the issues it fixes.
Also have some doubts about specific code parts:

SuyashJain17:
Thanks for the detailed feedback - that helps.

The issue this PR fixes is a runtime crash caused by next-themes applying a persisted theme value directly as a CSS class during initial render. If the stored value contains spaces (for example "Default Theme"), the browser throws an InvalidCharacterError and the app crashes before rendering.

The cleanup logic only removes the persisted value when it’s invalid (i.e. not one of ['dark', 'light']). This allows next-themes to fall back to a safe default and prevents the crash. Valid user preferences are not touched.

I agree the intent wasn’t clear enough from the code. I’ll:

- Add an explicit comment explaining the crash scenario and why the cleanup is needed

- Move the theme validation helpers away from the Django session–related comments

- Simplify the typing to avoid confusion around const / as const

Please let me know if you’d prefer a different approach (for example sanitizing instead of removing). I’m happy to adjust.

SuyashJain17:
Closing this PR since it contains multiple unrelated changes from an earlier branch.
I’ll open a clean, scoped PR for the assigned issue. Thanks!

arkid15r:
You should never close a PR unless you're asked to. PR is a "shell" for your code to be reviewed/merged. As an author you can do whatever you need w/ your branch linked to this PR.

SuyashJain17:
Thanks for clarifying, that makes sense - appreciate it.

I closed the PR earlier because the branch had accumulated multiple unrelated commits from experimentation, and I didn’t want to risk merging unintended changes.

I understand now that the PR can remain open while I clean up the branch history. I’ll rebase this PR onto a clean state and update it accordingly.

SuyashJain17:
Removed the dependency cleanup commit so this PR focuses only on the theme initialization crash fix.

frontend/src/wrappers/provider.tsx
```diff
@@ -7,22 +8,72 @@ import { ThemeProvider as NextThemesProvider } from 'next-themes'
...
+/**
+ * next-themes applies the persisted theme value directly as a CSS class
+ * during initial render.
+ *
+ * If the stored value contains spaces (e.g. "Default Theme"), the browser
+ * throws InvalidCharacterError and the app crashes before rendering.
+ *
+ * To prevent this, we validate the persisted value and remove it only
+ * if it is invalid, allowing next-themes to safely fall back to default.
+ */
```
kasya: @SuyashJain17 can you provide steps how to replicate this issue you're trying to fix? 👀 
Current default theme is set to "dark", so I'm not sure how it's possible to have it set to something like "Default Theme", unless user goes in and manually updates it in local storage. 

Is that the scenario you're working with?

Even if so, I believe toggling the theme toggle again - would update the value. 
However, even if we'd want to add extra validation, I believe the solution can be much much simpler - like a 1-2 lines change 🤔

SuyashJain17: Thanks for the follow-up — adding concrete details from the browser console.

### Reproduction (local via Docker)

1. Start the app using the standard Docker setup:
   ```bash
   docker compose -f docker-compose/local/compose.yaml up --build
   ```
2. Open the frontend at http://localhost:3000
3. Open the browser Developer Console
4. Reload the page

### Explanation

`next-themes` applies the persisted theme value directly as a CSS class
during initial render. If the value contains spaces (e.g. `"Default Theme"`),
the browser throws `InvalidCharacterError` because class tokens cannot contain
whitespace. This happens before React finishes mounting, causing a crash
before hydration and before the theme toggle can run.

The fix can be minimal: remove the persisted value only when it contains
invalid CSS characters, allowing `next-themes` to fall back safely.

SuyashJain17:
Thanks for the review!
SonarCloud is now green (Quality Gate passed).
I’ve simplified the fix to a minimal guard as suggested, and the crash no longer reproduces locally.
Happy to adjust further if needed.

arkid15r:
I don't think we'll be merging this.
Please feel free to update the troubleshooting section of https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md#troubleshooting though

Issue #2386, opened 2025-10-07
Fix sporadic 404 flash on /about page


arkid15r:
**Describe the bug**

It's rarely reproducible but it 100% exists. When you open https://nest.owasp.org/about page it sometimes shows quick 404 page while still loading the contents and then presents the loaded content.

**To Reproduce**

Open https://nest.owasp.org/about many times. Be lucky

**Expected behavior**

No 404 page shown

**Are you going to work on fixing this?**

- [ ] Yes
- [x] No

**Screenshots**

No screenshot, sorry.

abhaysagarofficial:
Hey @arkid15r , can you assign me this issue to me. I would love to work on this issue.

anurag2787:
@arkid15r Deadline for this issue has been passed still their is no update Could you please assign this issue to me?

arkid15r:
Reopening this as it still happens in prod.

HarshitVerma109:
@arkid15r
Could you please assign this issue to me?

PR #3124, opened 2025-12-31
Fix 404 flash on About page

HarshitVerma109:
**Description:**

This PR fixes a bug where the About page would incorrectly show a 404 error for a split second before the real content loaded.

**Changes:**

Refactor: Updated frontend/src/app/about/page.tsx to derive data directly from the Apollo query response instead of using local state. This ensures the component correctly handles the loading state and prevents the premature error message.

**Related Issue:**

Closes [#2386](https://github.com/OWASP/Nest/issues/2386)

### DIFF
--- a/frontend/src/app/about/page.tsx
+++ b/frontend/src/app/about/page.tsx
@@ -60,36 +60,30 @@ const getMilestoneIcon = (progress: number) => {
 }
 
 const About = () => {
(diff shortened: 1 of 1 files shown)


HarshitVerma109:
Understood, apologies for missing the template. I will be sure to follow it in future contributions

PR #3130, opened 2026-01-01
Fix ChapterMap locked state with explicit unlock

### LINKED ISSUE #3012: Improve ChapterMap UX: Locking, Interactions, and Screen Layout
### 1. Interaction Locking

* **Problem:** On mobile and desktop, when the map is "locked," interactions like dragging, zooming, and keyboard navigation should be disabled.
* **Goal:** Ensure that when the map is locked, all interactive features are disabled, and they are re-enabled only after the user explicitly unlocks the map.

### 2.. Wide Screen Layout / Empty Space

* **Problem:** On very wide screens, map tiles repeat horizontally, but using `fitBounds` and current `maxBounds` settings can leave empty white space at the edges.
* **Goal:** Adjust `minZoom` dynamically and/or update bounds logic to prevent empty spaces while allowing tile repetition.

**Proposed Fixes Implemented in Branch:**

* Disabled interactions when map is locked.
* Added proper event handling for mobile and desktop.
* Set `minZoom` and `maxZoom` for better wscreen experience.
* Enabled cluster and marker click to unlock map smoothly.
* Removed horizontal repetition and fixed map display to fit within the screen.

###Are you going to work on this

- [x] yes
- [ ] no

https://www.loom.com/share/60de02be06c94726bd1c628afd3e4e87

kris70lesgo:
## Proposed change

Resolves #3012 

This pr fixes the ChapterMap locked/unlocked interaction behavior by enforcing a fully locked map state on initial load and enabling map interactions only through an explicit user action.

### Summary of changes
- Disable all map interaction handlers by default (drag, zoom, scroll, keyboard).
- Enable interactions only when the user explicitly clicks the **Unlock map** button.
- Centralize map interaction toggling logic to avoid inconsistent state.
- Ensure visual and interactive cues accurately reflect the locked state.
- Update unit tests to assert full interaction enable/disable lifecycle.

This approach prevents accidental map interactions while scrolling and aligns the behavior with explicit user intent.

## Checklist


kris70lesgo:
@kasya can u review the pr ?

kris70lesgo:
@kasya, thanks for the review.
All the issues by coderabbit and sonar has been resolved 
I’ve now run `make check-test` end-to-end 
All checks pass except `pnpm run graphql-codegen`, which fails locally because the backend isn’t running (CSRF token fetch). 
Please let me know if you’d like me to spin up the full backend locally as well, but otherwise this should be ready for review now

frontend/src/components/ChapterMap.tsx
kasya, line 1: Previously the map would lock back on loosing focus. After your changes I'm only able to lock the map back with  `Escape` button but I believe we need to keep that lock on loosing focus as well. 🤔

kasya, line 1: Also I'm still able to click into pins when map is closed:

Image

kasya (changes requested):
Hi @kris70lesgo! Found a couple of issues you'll need to address with this PR ⬇️

kasya:
@kris70lesgo Hi! Do you plan to update this PR with requested changes? There are also conflicts that needs to be resolved.
If not - I will have to re-assign this issue to someone else.

kris70lesgo:
Hi @kasya, thanks for the follow-up.
Given the current time constraints on my side, I won’t be able to finish and validate the requested changes with the quality and speed this pr deserves
I think it would be better to re-assign the issue so it doesn’t block progress.
If it’s okay, I’d still like to continue working on the fix in parallel
Thanks for the opportunity to work on it

PR #3278, opened 2026-01-10
ci: enable Docker layer caching for cspell spellcheck job

### LINKED ISSUE #3269: CI: Spellcheck job does not benefit from Docker layer caching
The `spellcheck` CI job builds a Docker image for `cspell` on every run, but the Docker layer cache is not effective in GitHub Actions.  
Each CI job runs on a fresh runner, and the current `docker build --cache-from cspell` setup does not restore any cache across runs. As a result, the image (including pnpm dependencies) is rebuilt from scratch every time, adding unnecessary overhead to CI.

Make the spellcheck job benefit from caching in CI. Two possible approaches:

1. **Keep Docker, fix caching**  
   Use Docker Buildx with GitHub Actions cache (`cache-from` / `cache-to` with `type=gha`) so Docker layers can be reused across runs.

2. **Run cspell directly on the runner**  
   Run `cspell` using `setup-node` and pnpm on the runner, similar to other frontend checks, leveraging the existing pnpm cache and removing Docker overhead entirely.

Either approach would ensure the spellcheck job no longer performs cold builds on every CI run.

- Keeping the current setup, which always results in cold Docker builds and longer CI times.  
- Pulling a prebuilt image from a registry, which still requires proper cache configuration to be effective.

- [x] Yes
- [ ] No

**Additional context**  
Other CI jobs in this workflow already use effective caching strategies (e.g., pnpm cache, Docker Buildx cache, Trivy DB cache). Aligning the spellcheck job with these patterns would improve CI consistency and performance without changing functionality.

SuyashJain17:
## Proposed change

Resolves #3269

This PR optimizes the Docker-based spellcheck job by enabling Docker Buildx with GitHub Actions cache for the `cspell` image.

Previously, the spellcheck job rebuilt the Docker image from scratch on every CI run (~26s on recent `main` runs). With this change, the first run populates the cache and subsequent runs can reuse Docker layers, avoiding repeated cold builds.

### Summary of changes
- Enable Docker Buildx and GitHub Actions cache in the `spellcheck` job
- Update the Makefile to reuse the cached `cspell:ci` image when available
- Preserve existing Docker-based spellcheck behavior with no functional changes

### Notes
- The first CI run will still be a cold build (expected)
- Subsequent runs should show reduced build time due to Docker layer reuse

## Checklist

- [ ] **Required:** I ran `make check-test` locally and all tests passed  
  _(Not applicable — this change only affects CI configuration and Docker build caching)_
- [x] I used AI for code, documentation, or tests in this PR


cspell/Makefile
```diff
@@ -1,10 +1,9 @@
 check-spelling: cspell-check
 
 cspell-install:
-	@DOCKER_BUILDKIT=1 docker build \
-		--cache-from cspell \
-		cspell \
-		-t cspell
+	@docker image inspect cspell:ci >/dev/null 2>&1 && \
```
arkid15r, line 6: What's goal of these changes?

SuyashJain17: The goal is to avoid rebuilding the cspell Docker image from scratch on every CI run.

The workflow now builds the image once using Buildx with the GitHub Actions cache and tags it as cspell:ci.
The Makefile change simply reuses that image when it’s already available and falls back to a local build otherwise, so CI benefits from caching without changing local behavior.

PR #3295, opened 2026-01-11
Fix brittle OWASP event date parsing and add regression tests

### LINKED ISSUE #3286: Improve date range parsing in Event model to handle varied delimiters and crossovers
**Describe the bug**
The `parse_dates` method in `backend/apps/owasp/models/event.py` relies on brittle string manipulation to parse date ranges. Currently, it uses hardcoded splitting on specific characters like `,` and `-` and assumes a very strict "Month Day-Day, Year" structure. If the input data varies even slightly (different dash types, extra spaces, or month-crossovers), the parser fails or returns `None`.

**To Reproduce**
1. Pass a date string using an em-dash or en-dash instead of a hyphen (e.g., `"May 26—30, 2025"`).
2. The current `if "-" in dates` check fails, skipping range logic entirely.
3. Pass a date string with irregular spacing (e.g., `"May 26 - 30 , 2025"`).
4. The manual splitting and integer casting `int(day_range.split("-")[-1])` becomes prone to `ValueError`.
5. Try a month-crossover range like `"May 30 - June 2"`. The current logic cannot handle this format at all.

**Expected behavior**
The parser should be versatile enough to handle various delimiters (hyphens, dashes, spaces) and different date formats (like month-crossovers or international formatting) without crashing or returning `None`. It should leverage standard libraries like `dateutil` consistently rather than manual string hacking to ensure stability.

- [x] Yes
- [ ] No

**Additional context**
I noticed a few technical debt items while looking at this:
- **Positional Fragility:** The code assumes the month is always at `parts[0]`, which breaks non-US formats.
- **Library Inconsistency:** `dateutil.parser` is imported but ignored for the complex range logic in favor of manual string manipulation.
- **Modernization:** This refactor will address the existing `TODO` in the source code (line 104) regarding method modernization.

I have a drafted solution using Regex for normalization and `dateutil.parser` with a default context to handle these edge cases. I'd also like to add a set of unit tests to make sure this remains stable.

saichethana28:
## Proposed change

Resolves #3286 

This PR addresses several edge cases in OWASP event date parsing that were causing failures due to slight variations in input format. The earlier logic depended on manual string splitting and assumed a very specific format (for example, a single dash type, fixed spacing, and same-month ranges). In practice, event dates often include em/en dashes, extra spaces, ISO-style dates, or ranges that cross months or years, which caused the parser to return incorrect results or `None`.

This change:
- Normalizes different dash characters before parsing
- Handles ISO-style single dates explicitly
- Parses date ranges using `dateutil.parser` with the start date as context, allowing month and year crossovers
- Removes assumptions about fixed string positions
- Adds unit and integration tests covering the cases above to prevent regressions

The behavior for previously valid inputs remains unchanged.

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -23,6 +23,7 @@ Add the PR description here.
 
 ## Checklist
 
-- [ ] **Required:** I read and followed the [contributing guidelines](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md)
-- [ ] **Required:** I ran `make check-test` locally and all tests passed
-- [ ] I used AI for code, documentation, or tests in this PR
+- [ ] **Required:** I followed the [contributing workflow](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md#contributing-workflow)
+- [ ] **Required:** I verified that my code works as intended and resolves the issue as described
+- [ ] **Required:** I ran `make check-test` locally: all warnings addressed, tests passed
+- [ ] I used AI for code, documentation, tests, or communication related to this PR
--- a/.github/workflows/run-ci-cd.yaml
+++ b/.github/workflows/run-ci-cd.yaml
@@ -145,7 +145,6 @@ jobs:
           scan-type: repo
           skip-setup-trivy: true
           trivy-config: trivy.yaml
-          trivyignores: trivyignore.yaml
 
   scan-ci-dependencies:
     name: Run CI Dependencies Scan
@@ -171,7 +170,6 @@ jobs:
           scan-type: fs
           skip-setup-trivy: true
           trivy-config: trivy.yaml
-          trivyignores: trivyignore.yaml
 
   run-backend-tests:
     name: Run backend tests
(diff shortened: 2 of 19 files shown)


backend/apps/owasp/models/event.py
```diff
@@ -66,12 +67,7 @@ def __str__(self) -> str:
 
     @staticmethod
     def upcoming_events():
-        """Get upcoming events.
-
-        Returns
-            QuerySet: A queryset of upcoming Event instances ordered by start date.
-
-        """
+        """Get upcoming events."""
```
arkid15r: Any good reason for removing docstrings?

saichethana28: Thanks for calling this out.
I shortened docstrings where they were repeating the method name and obvious return behavior, without adding extra context. The idea was to keep documentation focused on places where the behavior isn’t selfevident (like parse_dates).
Happy to restore the longer version here if consistency is preferred.

arkid15r: It's not a longer version. It's a template we use across BE code base even though not all methods are compliant yet.
Please revert.

backend/apps/owasp/models/event.py
```diff
@@ -117,41 +120,36 @@ def parse_dates(dates: str, start_date: date) -> date | None:
...
+                    day_val = day_match.group(1)
+                    if len(end_str) <= max_day_length or "," in end_str:
+                        return start_date.replace(day=int(day_val))
 
-                # Use the start_date to determine the month if provided
-                if start_date:
-                    start_date_parsed = start_date
-                    month = start_date_parsed.strftime("%B")  # Full month name (e.g., "May")
+                default_dt = datetime.combine(start_date, datetime.min.time())
+                return parser.parse(end_str, default=default_dt).date()
```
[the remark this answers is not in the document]
saichethana28: Fixed year crossover handling by incrementing the year when the parsed end date precedes the start date. Backend tests pass.

saichethana28:
@coderabbitai review

arkid15r (approved):
LGTM

PR #3303, opened 2026-01-11
Hide contribution section when no contributions exist

### LINKED ISSUE #3262: Do not show contribution heatmap for 0 contribution chapters/projects
It looks weird for new projects for example

Do not render the component if entity has 0 contributions.

- [ ] Yes
- [x] No

**Additional context**

https://nest.owasp.org/projects/browser-security-project

Image

saichethana28:
## Proposed change

Resolves #3262

This PR ensures that the contribution activity section (stats + heatmap) is **not rendered when an entity has zero contributions**, preventing confusing or empty UI states for new or inactive projects/chapters.

Instead of guarding this at an individual page level, the logic is centralized inside the shared `DetailsCard` component. A single, explicit `hasContributions` check determines whether contribution-related UI should be mounted at all. This keeps the behavior consistent across all consumers of `DetailsCard` and avoids duplicating conditional logic in multiple routes.

### Why this approach

- **Single source of truth:** Contribution visibility is owned by `DetailsCard`, the component responsible for rendering it.
- **Consistent behaviour:** Applies uniformly to both projects and chapters without page-specific conditionals.
- **Non-invasive:** No backend, API, or GraphQL changes; rendering behavior only.
- **Future-safe:** If additional entities or pages reuse `DetailsCard`, the correct behaviour comes “for free”.

### Implementation details

- Introduced a small derived boolean (`hasContributions`) based on:
  - presence of contribution stats, or
  - non-empty contribution data
- The entire contribution section (stats + heatmap container) is skipped when this condition is false.
- Existing behaviour is unchanged when contribution data exists.

This results in a cleaner UI for entities with no activity while preserving all existing contribution visualizations for active ones.

## Checklist

- [x] **Required:** I ran `make check-test` locally (frontend-only change; CI covers full environment)

### DIFF
--- a/frontend/src/types/card.ts
+++ b/frontend/src/types/card.ts
@@ -85,14 +85,6 @@ export interface DetailsCardProps {
   tags?: string[]
   type: CardType
   userSummary?: JSX.Element
-
-  /**
-   * Added to support hiding contribution UI when no contributions exist
-   */
-  contributionStats?: unknown
-  contributionData?: Record<string, unknown>
-  startDate?: string
-  endDate?: string
 }
 
 export interface UserCardProps {


saichethana28:
Hi @arkid15r 
Just to clarify on approach,  
In my initial issue comment I mentioned guarding the render at the page level. While implementing, I realised the contribution UI is fully owned by `DetailsCard`, so moving the check there avoids duplicating the same condition across multiple pages (projects, chapters, and any future consumers).
This keeps the change frontend-only, doesn’t affect data fetching or APIs, and centralises the behaviour in the component responsible for rendering it.
Happy to adjust if you’d prefer this logic to stay page specific instead. Thanks!

kasya:
Closed in favor of #3314

PR #3344, opened 2026-01-14
feat(frontend): add PostHog analytics for staging and production

### LINKED ISSUE #2838: Add PostHog support

Add PostHog to OWASP Nest production and staging environments (skip other like test, local, etc)

- [ ] Yes
- [x] No

**Additional context**

https://posthog.com/docs/getting-started/install

SuyashJain17:
## Proposed change

Resolves #2838

This PR integrates **PostHog analytics** into the frontend and enables it **only for staging and production environments**.

### Summary of changes
- Added `posthog-js` as a frontend dependency
- Introduced a `PostHogProvider` to initialize and wrap the app
- Captured page views manually to align with Next.js routing
- Added environment variables for PostHog configuration
- Ensured analytics is disabled for local/test environments

### Notes
- `pnpm-lock.yaml` was regenerated after rebasing on `main` and adding `posthog-js`, which resulted in a larger diff.  
- No functional changes were made beyond the PostHog integration.

Happy to adjust the lockfile or pnpm version if a different approach is preferred.

## Checklist

- [ ] **Required:** I ran `make check-test` locally: all warnings addressed, tests passed


Shofikul-Isl4m:
image
you also have to adress sonarcloud issue .

frontend/pnpm-lock.yaml
Shofikul-Isl4m, line 1: Could you please tell me what you have done here? Based on my experience, there might be a potential issue with this approach. Though, the code owners can surely tell if it's wrong or not.

frontend/src/wrappers/PostHogProvider.tsx
arkid15r, line 1: Why do we need this provider?

SuyashJain17: The provider mainly exists for two reasons. First, with the App Router the app behaves like an SPA, so client-side navigations don’t trigger full page loads. The provider renders a small PostHogPageView component that listens to route changes and manually captures $pageview events, otherwise only the initial load would be tracked.

Second, we’re using the PostHog React SDK (usePostHog()), which requires a PostHogProvider in the tree. This keeps initialization and environment gating in one place and allows us to safely use PostHog hooks later (feature flags, custom events, etc.) without relying on a global singleton.

The intent was to keep analytics logic centralized and aligned with how other providers (like Sentry) are handled in the app.

arkid15r: Their official documentation suggests nothing similar.
Will it work w/o it?

SuyashJain17: I agree - following the docs, PostHog will work without a custom provider.
I added the provider mainly to handle client-side navigation in the App Router and to keep things ready for using usePostHog() later (for events, flags, etc.).
If it makes more sense to keep this closer to the official setup and simpler, I’m fine refactoring it to remove the provider.

arkid15r: Yes, I'd prefer minimum working setup. Please simplify it.

Thank you!

SuyashJain17: Okay, Working on it !

SuyashJain17:
Thanks for the feedback, @arkid15r.
I’ve removed the custom provider and switched to the official PostHog Next.js setup using instrumentation-client.ts.
Cleaned up the commit history and kept the diff minimal. Ready for another look.

SuyashJain17:
Hello @arkid15r, I’ve updated the PR in line with the changes you requested. When you have some time, could you please review it again?

arkid15r:
Closing in favor of https://github.com/OWASP/Nest/pull/3537

PR #3547, opened 2026-01-25
Migrated project issue to m2m field

### LINKED ISSUE #3310: Migrate Project issues property to M2M relation field
Currently, the issues property is just another query to execute, which is fine when we are fetching one project, but when we will fetch multiple projects we will execute this query for each project which we will have N+1 queries.
For example:
```graphql
query {
   recentProjects(limit: 1000) {
      recentIssues {
         id
      }
   }
}
```
We will have a query for fetching the projects and one query for each project to fetch its issues, so we will have 1001 queries!

1. Add an M2M field to the `Project` model for the issues.
2. Remove the issues property.
3. Update any related properties (e.g. issues_count) to use this new field.
4. Update the GraphQL `ProjectNode` to use this field instead.
5. Update the `owasp_aggregate_projects` command to populate the issues in the new field (we should use the same query in the issues property).
6. Update related tests.

- [ ] Yes
- [x] No

anurag2787:
## Proposed change
This PR fixes the N+1 query problem when fetching project issues by moving issues to a proper many-to-many field on Project
Resolves #3310

## Checklist

### DIFF
--- a/backend/apps/owasp/api/internal/nodes/project.py
+++ b/backend/apps/owasp/api/internal/nodes/project.py
@@ -90,17 +90,19 @@ class ProjectNode(GenericEntityNode):
                     "level",
                     "milestone",
                     "repository",
-                ).prefetch_related(
+                )
+                .prefetch_related(
                     "assignees",
                     "labels",
-                ).order_by("-created_at")[:RECENT_ISSUES_LIMIT],
-                to_attr="_recent_issues",
+                )
+                .order_by("-created_at")[:RECENT_ISSUES_LIMIT],
+                to_attr="recent_issues",
             ),
         ],
     )
     def recent_issues(self, root: Project) -> list[IssueNode]:
         """Resolve recent issues."""
-        return root._recent_issues
+        return root.recent_issues
 
     @strawberry_django.field
     def recent_milestones(self, root: Project, limit: int = 5) -> list[MilestoneNode]:
--- a/backend/apps/owasp/management/commands/owasp_aggregate_projects.py
+++ b/backend/apps/owasp/management/commands/owasp_aggregate_projects.py
@@ -96,7 +96,7 @@ class Command(BaseCommand):
                 ),
             )
             project.issues.set(issues)
-            project.issues_count = issues.count()
+            project.issues_count_value = issues.count()
 
             project.pushed_at = max(pushed_at)
             if released_at:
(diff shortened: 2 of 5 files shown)


backend/apps/owasp/api/internal/nodes/project.py
```diff
@@ -79,7 +79,16 @@ def languages(self, root: Project) -> list[str]:
...
+    @strawberry_django.field(
+        prefetch_related=[
+            "issues__author",
+            "issues__level",
+            "issues__milestone",
+            "issues__repository",
+            "issues__assignees",
+            "issues__labels",
+        ]
+    )
```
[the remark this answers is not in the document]
anurag2787: Moving all select_related / prefetch_related into the resolver would re-introduce a GraphQL level N+1 problem

ahmedxgouda, line 102: This is not the right way to provide a hint for a queryset with an `order_by` clause. You should use `Callable[Info, Prefetch]` type. See this [PR](https://github.com/OWASP/Nest/pull/3560) and the [docs](https://strawberry.rocks/docs/django/guide/optimizer#optimization-hints)

backend/apps/owasp/models/mixins/project.py
```diff
@@ -51,7 +51,7 @@ def idx_is_active(self) -> bool:
     @property
     def idx_issues_count(self) -> int:
         """Return issues count for indexing."""
-        return self.open_issues.count()
+        return self.issues.count()
```
ahmedxgouda: Let's use the new populated field in the model.

backend/apps/owasp/management/commands/owasp_aggregate_projects.py
```diff
@@ -85,6 +87,18 @@ def handle(self, *_args, **options) -> None:
                 if repository.topics:
                     topics.update(repository.topics)
 
+            # Populate issues from all repositories.
+            repository_ids = project.repositories.filter(
+                is_empty=False,
+                is_fork=False,
+                is_template=False,
+            ).values_list("id", flat=True)
+            issues = Issue.objects.filter(
```
ahmedxgouda: This query can be simpler. Why are we extracting the IDs? Can't we just use `repository__in`?

ahmedxgouda: I believe the original query does not have these filters.

arkid15r:
Closing as not planned for now, sorry for unnecessary assignment.

Issue #3311, opened 2026-01-12
Migrate Project pull_requests property to M2M relation field


ahmedxgouda:
**Is your feature request related to a problem? Please describe.**
Currently, the pull_requests property is just another query to execute, which is fine when we are fetching one project, but when we will fetch multiple projects we will execute this query for each project which we will have N+1 queries.
For example:
```graphql
query {
   recentProjects(limit: 1000) {
      recentPullRequests {
         id
      }
   }
}
```
We will have a query for fetching the projects and one query for each project to fetch its pull requests, so we will have 1001 queries!

**Describe the solution you'd like**
1. Add an M2M field to the `Project` model for the pull_requests.
2. Remove the pull_requests property.
3. Update any related properties (e.g. pull_requests_count) to use this new field.
4. Update the GraphQL `ProjectNode` to use this field instead.
5. Update the `owasp_aggregate_projects` command to populate the pull_requests in the new field (we should use the same query in the pull_requests property).
6. Update related tests.

**Are you going to work on implementing this?**

- [ ] Yes
- [x] No

OM-JADHAV25:
Hii @arkid15r 
I would like to work on this issue.
If that's okay, please assign it to me !!
Thanks !!!

anurag2787:
Hi @arkid15r could you please assign this to me?

kart-u:
Hi @arkid15r 
I would like to work on this , can i get assigned??

saichethana28:
Hi @arkid15r, I’d like to work on this. I’ve reviewed the pull_requests property and am ready to implement the ManyToManyField along with a data migration to ensure existing PR links are preserved. I'll also update the backend-tests to match the new relationship.

saichethana28:
Actually, just digging into the GraphQL side of this, the N+1 issue makes a lot of sense. I'm happy to handle the ProjectNode updates and refactor the owasp_aggregate_projects command to use bulk_create for the new M2M links. It’ll be much cleaner than the current property approach.

SuyashJain17:
Hi @arkid15r,
I understand this issue and the N+1 pattern here.
Happy to work on i.

Aditya23-lab:
Hi @arkid15r @kasya 

I took a look at how `pull_requests` is currently resolved and the N+1 issue is pretty clear, especially when `recentPullRequests` is queried across many projects.

[...]

PR #3560, opened 2026-01-25
Migrate Project pull_requests property to M2M relation field

mrkeshav-05:
### Proposed change
This PR replaces the `pull_requests` property with a `ManyToManyField` to enable efficient prefetching and eliminate the N+1 query problem.
A new migration, `0073_add_project_pull_requests_m2m.py`, has been added to introduce the M2M field. Additionally, the Docker database volume name was temporarily modified to avoid conflicts during migration and testing.

Resolves: #3311 

## Checklist

### DIFF
--- a/backend/tests/apps/owasp/api/internal/nodes/project_test.py
+++ b/backend/tests/apps/owasp/api/internal/nodes/project_test.py
@@ -1,6 +1,6 @@
 """Test cases for ProjectNode."""
 
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, Mock
 
 from apps.github.api.internal.nodes.issue import IssueNode
 from apps.github.api.internal.nodes.milestone import MilestoneNode
@@ -9,7 +9,6 @@ from apps.github.api.internal.nodes.release import ReleaseNode
 from apps.github.api.internal.nodes.repository import RepositoryNode
 from apps.owasp.api.internal.nodes.project import ProjectNode
 from apps.owasp.api.internal.nodes.project_health_metrics import ProjectHealthMetricsNode
-from apps.owasp.models.project import Project
 from tests.apps.common.graphql_node_base_test import GraphQLNodeBaseTest
 
 
--- a/backend/tests/apps/owasp/management/commands/owasp_aggregate_projects_test.py
+++ b/backend/tests/apps/owasp/management/commands/owasp_aggregate_projects_test.py
@@ -1,10 +1,8 @@
-import contextlib
 from unittest import mock
 
 import pytest
 
 from apps.github.models.pull_request import PullRequest
-from apps.github.models.repository import Repository
 from apps.owasp.management.commands.owasp_aggregate_projects import Command, Project
 
 
@@ -141,9 +139,13 @@ class TestOwaspAggregateProjects:
         )
         mock_active_projects.order_by.return_value = mock_active_projects
 
+        mock_pull_requests = mock.MagicMock()
+        mock_pull_requests.filter.return_value = MockQuerySet([])
+
         with (
             mock.patch.object(Project, "active_projects", mock_active_projects),
             mock.patch("builtins.print"),
+            mock.patch.object(PullRequest, "objects", mock_pull_requests),
         ):
             command.handle(offset=0)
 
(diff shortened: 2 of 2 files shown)


ahmedxgouda:
Please resolve the conflicts @mrkeshav-05

backend/apps/owasp/management/commands/owasp_aggregate_projects.py
```diff
@@ -110,6 +112,23 @@ def handle(self, *_args, **options) -> None:
...
+                    "milestone",
+                    "repository__organization",
+                    "repository",
+                )
+                .prefetch_related(
+                    "assignees",
+                    "labels",
+                )
+            )
+            project.pull_requests.set(pull_requests)
```
ahmedxgouda: We don't need `select_related` and `prefetch_related` here. `project.pull_requests.set` only needs the ids no other fields.

mrkeshav-05:
Hi @ahmedxgouda , 
Now the endpoint is working 
image

backend/apps/owasp/api/internal/nodes/project.py
```diff
@@ -103,7 +103,16 @@ def recent_milestones(self, root: Project, limit: int = 5) -> list[MilestoneNode
...
+    @strawberry_django.field(
+        prefetch_related=[
+            "pull_requests__author",
+            "pull_requests__milestone",
+            "pull_requests__repository__organization",
+            "pull_requests__repository",
+            "pull_requests__assignees",
+            "pull_requests__labels",
+        ],
+    )
```
ahmedxgouda, line 121: This didn't solve the N+1 queries problem yet. The problem that order_by returns a new queryset the bypasses the the optimization. The prefetch_related decorator may accept `Callable[Info, Prefetch]` as argument. Consider reading the [docs](https://strawberry.rocks/docs/django/guide/optimizer#optimization-hints)

ahmedxgouda (changes requested):
Thank you for working on this, yet we still have N+1 queries. Please see the below comment.

backend/apps/owasp/api/internal/nodes/project.py
mrkeshav-05: I have refactored the resolver to use a Prefetch callable so the ordered queryset is prefetched instead of being recreated in the resolver, which prevents bypassing the optimizer.

backend/apps/owasp/api/internal/nodes/project.py
```diff
@@ -103,10 +105,27 @@ def recent_milestones(self, root: Project, limit: int = 5) -> list[MilestoneNode
...
+                )
+                .prefetch_related(
+                    "assignees",
+                    "labels",
+                )
+                .order_by("-created_at"),
+                to_attr="_recent_pull_requests",
+            ),
+        ],
+    )
```
ahmedxgouda, line 121: I have tested the resolver and yes the N+1 queries problem is solved! but the response is really slow. Maybe move the limit to the prefetch query? Also do we need these M2M fields of PullRequest to pass it to `prefetch_related`? Does the PullRequestNode return them? Also consider nested relationships (e.g. author__owasp_profile).

ahmedxgouda (changes requested):
Thank you! We are really close! Please see the bellow comment

backend/apps/owasp/api/internal/nodes/project.py
mrkeshav-05: Thanks for the suggetion. I experimented with applying the limit inside the Prefetch queryset, but that would make the limit global across all projects rather than per project, which changes the semantics of recentPullRequests.
For now, I kept the limit in Python to preserve correctness while avoiding N+1, and reduced over-fetching by narrowing select_related to only fields exposed by PullRequestNode (including author__owasp_profile)

ahmedxgouda: image
image

Here are pull requests from two projects. I believe the optimizer is applying the limit to the project pull requests only not globally. 
Any way I tested this query and it is taking 743ms only (it was taking 3 seconds without the limit applied to the prefetch).
```graphql
{
  recentProjects(limit: 1000) {
    id
    name
    recentPullRequests {
      id
      repositoryName
      organizationName
      author {
        isOwaspStaff
      }
      
    }
  }
}
```

ahmedxgouda:
@mrkeshav-05 Could you increase the coverage?

mrkeshav-05:
@ahmedxgouda I’ve increased the test coverage for this change.

arkid15r:
Closing as not planned for now, sorry for unnecessary assignment.

PR #3572, opened 2026-01-26
fix: replace status role with output in CardSkeleton

Adarshkumar0509:
Update CardSkeleton loading placeholder to improve layout and configurability by adjusting default props and markup for more accurate skeletons during load.


Adarshkumar0509:
I will love to work on this issue. Could you please assign me?

PR #3608, opened 2026-01-28
feat: Enable Pagination for Pull Requests on Issue and Module Pages

### LINKED ISSUE #3573: Pagination for Pull Requests
## 🎯 Issue Description

The Pull Request lists on the **Issue Details Page** and **Public Module Page** are currently limited to a fixed number of items (e.g., 4 or 5), with no option to view the remaining Pull Requests. This prevents users from accessing the full contribution history.

## 🛠️ Proposed Fix

Implement pagination to allow users to view all linked Pull Requests:
- **Backend**: Update the API to support `limit` and `offset` for fetching Pull Requests.
- **Frontend**: Add a **"Show More"** button to load additional Pull Requests on demand.
- **UI**: Ensure smooth loading and consistent behavior across both pages.

- [x] Yes
- [ ] No

HarshitVerma109:
## Proposed change

Resolves #3573 

This PR implements pagination for Pull Requests to fix visibility limits on the **Issue Details Page** and **Public Module Page**.
Previously, these lists were capped at a fixed number (4 or 5) with no way to view the rest.

**Key Changes:**
- **Frontend**: Added "Show More" functionality using `fetchMore` to load additional PRs on demand without page refreshes.
- **Backend**: Updated [IssueNode](cci:2://file:///home/harsh/Nest/backend/apps/github/api/internal/nodes/issue.py:10:0-64:9) to support `limit` and `offset` arguments for efficient data fetching.

## Checklist

### DIFF
--- a/frontend/__tests__/unit/components/CardDetailsPage.test.tsx
+++ b/frontend/__tests__/unit/components/CardDetailsPage.test.tsx
@@ -1013,6 +1013,21 @@ describe('CardDetailsPage', () => {
       expect(screen.getByRole('button', { name: /Show less/i })).toBeInTheDocument()
     })
 
+    it('shows Loading on Show more and disables both controls when isFetchingMore is true', () => {
+      render(
+        <CardDetailsPage
+          {...defaultProps}
+          type="module"
+          pullRequests={mockPullRequests as unknown as PullRequest[]}
+          onLoadMorePullRequests={jest.fn()}
+          onResetPullRequests={jest.fn()}
+          isFetchingMore={true}
+        />
+      )
+      expect(screen.getByRole('button', { name: /Loading/i })).toBeDisabled()
+      expect(screen.getByRole('button', { name: /Show less/i })).toBeDisabled()
+    })
+
     it('calls scrollToAnchor when MetricsScoreCircle is clicked', () => {
       const { scrollToAnchor } = jest.requireMock('utils/scrollToAnchor')
 
--- a/frontend/__tests__/unit/pages/ModuleIssueDetailsPage.test.tsx
+++ b/frontend/__tests__/unit/pages/ModuleIssueDetailsPage.test.tsx
@@ -180,6 +180,26 @@ describe('ModuleIssueDetailsPage', () => {
     expect(screen.getAllByAltText('Loading indicator')[0]).toBeInTheDocument()
   })
 
+  it('does not show full-page spinner when loading with cached issue (e.g. fetchMore)', () => {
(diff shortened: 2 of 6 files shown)


HarshitVerma109:

HarshitVerma109:

HarshitVerma109:

HarshitVerma109:

HarshitVerma109:

HarshitVerma109:

HarshitVerma109:
@kasya , please review it and let me know if any changes are required.

HarshitVerma109:

HarshitVerma109:

preeeetham:
Good feature — paginating pull requests on issue and module pages is the right approach to avoid loading everything upfront. A few things stood out from the diff:

The `IssueNode.pull_requests` resolver has a potential name collision. The method is called `pull_requests` and it accesses `self.pull_requests.all()` — but `self` here refers to the IssueNode instance, and the method itself is called `pull_requests`. In regular Python this would be infinite recursion, but strawberry-django resolvers receive `root` as the actual model instance. However, the resolver signature uses `self` instead of `root` as the first parameter, which means `self.pull_requests` refers to the method, not the model's related manager. The prefetch decorator puts data on the model instance, but the resolver accesses `self` which is the Node class. This should be using `root` as the parameter name and accessing `root.pull_requests.all()` to avoid ambiguity.

The `hasMorePRs` logic in `ModuleDetailsPage` is quite complex and has a subtle issue. The useEffect sets `setHasMorePRs(prCount >= limit)` only when `prCount <= limit`, but this condition is always true when `prCount >= limit` is evaluated inside it — so `hasMorePRs` only ever gets set on the first load. After a `fetchMore`, the `prCount` will be the total accumulated count (e.g., 8 after two fetches), and the check `prCount <= limit` will be false, so `hasMorePRs` never gets set to `false`. The "Show more" button will keep appearing even after all PRs are loaded. The `fetchMore` callback does set it via `if (newPRs.length < limit) setHasMorePRs(false)`, but that's in the `.then` of the promise — if the component re-renders between the fetch completing and the state update, you could see a flash.

The `onLoadMorePullRequests` prop is an inline function that creates a new reference on every render, which means `CardDetailsPage` will re-render unnecessarily every time the parent re-renders. Wrapping it in `useCallback` would prevent this.

HarshitVerma109:
@arkid15r, I’ve resolved the merge conflicts. Please review the changes and let me know if any further adjustments are needed.

PR #3678, opened 2026-01-30
refactor(ci): standardize security scans via Makefile targets

### LINKED ISSUE #3615: Refactor CI to use Makefile targets for security scans
### Context
With PR #3525 merged, we now have the Makefile targets set up for security scanning (renamed to `make security-scan-images` in the final merge).

However, looking at `.github/workflows/run-ci-cd.yaml`, we can see that we are still running Trivy image scans using inline configuration (around lines 916-937). Alternatively, you can search for "Setup Trivy" in that file. This means we have duplicate logic between our local dev environment and CI.

### Goal
Update `run-ci-cd.yaml` to simply invoke the Makefile targets instead of defining the Trivy configuration manually.

### Why
1. **Standardization:** Ensures we are running the same commands locally as we do in CI.
2. **Maintenance:** If we need to change flags or ignore specific CVEs later, we only have to update the Makefile, not the YAML.
3. **Cleanup:** Removes a big chunk of hardcoded config from the workflow file.

### Plan
- [ ] Update `scan-staging-images` job to run `make security-scan-images`.
- [ ] Update `scan-production-images` job to run `make security-scan-images`.
- [ ] Remove the hardcoded `trivy-action` blocks.

cc: @SpruhaCK (following up on our discussion in #3525)

hassaansaleem28:
## Proposed change

Resolves #3615 

## Description
This PR refactors the CI pipeline to use `make` targets for security scanning instead of inline configuration in the YAML workflow. This aligns our local development environment with CI/CD and solves the hardcoded configuration issues identified in #3615.

## Key Changes
- **Dependabot Tracking:** Added `docker/trivy/Dockerfile` to allow Dependabot to track and update the Trivy version automatically.
- **Performance:** Implemented Docker volume caching (`trivy-cache`) in the Makefile targets to prevent re-downloading the vulnerability DB on every run.
- **CI vs Local Logic:** Updated `backend/Makefile` and `frontend/Makefile` to support:
  - **Local Mode:** `TRIVY_EXIT_CODE=0` (Runs only misconfigs+vulns and not secrets).
  - **CI Mode:** `TRIVY_EXIT_CODE=1` (Fails the build on Critical/High vulnerabilities, runs misconfigs + secrets + vulns).

## Testing
- [x] Verified locally that `make security-scan-images` runs successfully.
- [x] Verified "Strict Mode" by running `make security-scan-backend-image TRIVY_EXIT_CODE=1` (confirmed 0 critical vulnerabilities).

## Checklist

### DIFF
--- a/backend/apps/api/rest/v0/issue.py
+++ b/backend/apps/api/rest/v0/issue.py
@@ -90,7 +90,12 @@ def list_issues(
     if filters.state:
         issues = issues.filter(state=filters.state)
 
-    return issues.order_by(ordering or "-created_at", "-updated_at")
+    primary_order = ordering or "-created_at"
+    order_fields = [primary_order]
+    if primary_order not in {"updated_at", "-updated_at"}:
+        order_fields.append("-updated_at")
+
+    return issues.order_by(*order_fields)
 
 
 @router.get(
--- a/backend/apps/api/rest/v0/milestone.py
+++ b/backend/apps/api/rest/v0/milestone.py
@@ -92,7 +92,12 @@ def list_milestones(
     if filters.state:
         milestones = milestones.filter(state=filters.state)
 
-    return milestones.order_by(ordering or "-created_at", "-updated_at")
+    primary_order = ordering or "-created_at"
+    order_fields = [primary_order]
+    if primary_order not in {"updated_at", "-updated_at"}:
+        order_fields.append("-updated_at")
+
+    return milestones.order_by(*order_fields)
 
 
 @router.get(
--- a/backend/apps/api/rest/v0/pagination.py
+++ b/backend/apps/api/rest/v0/pagination.py
@@ -32,20 +32,14 @@ class CustomPagination(PaginationBase):
         """Paginate the queryset and return standardized output."""
         page = pagination.page
         page_size = pagination.page_size
-
-        # Calculate pagination.
         total_count = queryset.count()
-        # Ensure total_pages is at least 1 for consistent metadata.
         total_pages = max(1, (total_count + page_size - 1) // page_size)
-
-        # Validate that the requested page is within the valid range.
         if page > total_pages:
             message = f"Page {page} not found. Valid pages are 1 to {total_pages}."
             raise Http404(message)
 
         offset = (page - 1) * page_size
 
-        # Get the page items.
         items = list(queryset[offset : offset + page_size])
 
         return {
--- a/backend/apps/api/rest/v0/release.py
+++ b/backend/apps/api/rest/v0/release.py
@@ -91,7 +91,12 @@ def list_release(
     if filters.tag_name:
         releases = releases.filter(tag_name=filters.tag_name)
 
-    return releases.order_by(ordering or "-published_at", "-created_at")
+    primary_order = ordering or "-published_at"
+    order_fields = [primary_order]
+    if primary_order not in {"created_at", "-created_at"}:
+        order_fields.append("-created_at")
+
+    return releases.order_by(*order_fields)
 
 
 @router.get(
--- a/backend/tests/apps/ai/agent/agent_test.py
+++ b/backend/tests/apps/ai/agent/agent_test.py
@@ -1,0 +1,96 @@
+"""Tests for the AgenticRAGAgent class."""
+
+from unittest.mock import MagicMock
+
+from apps.ai.agent.agent import AgenticRAGAgent
+
+
+class TestAgenticRAGAgent:
+    """Tests for AgenticRAGAgent."""
+
+    target_module = "apps.ai.agent.agent"
+
+    def test_init(self, mocker):
+        """Test AgenticRAGAgent initialization."""
+        mock_nodes = mocker.patch(f"{self.target_module}.AgentNodes")
+        mock_state_graph = mocker.patch(f"{self.target_module}.StateGraph")
+
+        mock_graph_instance = MagicMock()
+        mock_state_graph.return_value = mock_graph_instance
+
+        agent = AgenticRAGAgent()
+
+        mock_nodes.assert_called_once()
+        assert agent.nodes is not None
+        assert agent.graph is not None
+
+    def test_run(self, mocker):
+        """Test the run method executes the RAG workflow."""
+        mocker.patch(f"{self.target_module}.AgentNodes")
+        mock_state_graph = mocker.patch(f"{self.target_module}.StateGraph")
+
+        mock_graph_instance = MagicMock()
+        mock_compiled_graph = MagicMock()
+        mock_state_graph.return_value = mock_graph_instance
+        mock_graph_instance.compile.return_value = mock_compiled_graph
+
+        mock_compiled_graph.invoke.return_value = {
+            "answer": "Test answer",
+            "context_chunks": [{"text": "chunk1"}],
+            "evaluation": {"score": 0.9},
+            "extracted_metadata": {"key": "value"},
+            "history": ["step1", "step2"],
+            "iteration": 2,
+        }
+
+        agent = AgenticRAGAgent()
+        result = agent.run("Test query")
+
+        mock_compiled_graph.invoke.assert_called_once()
+        assert result["answer"] == "Test answer"
+        assert result["context_chunks"] == [{"text": "chunk1"}]
+        assert result["evaluation"] == {"score": 0.9}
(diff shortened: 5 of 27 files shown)


Makefile
```diff
@@ -111,9 +111,11 @@ security-scan-code-trivy:
...
 		--rm \
 		-v "$(PWD):/src" \
+		-v trivy-cache:/root/.cache/trivy \
 		-w /src \
-		aquasec/trivy fs \
-			--config trivy.yaml \
+		$$(grep -E '^FROM aquasec/trivy:' docker/trivy/Dockerfile | sed 's/^FROM //') \
+		fs \
+		--config trivy.yaml
 			.
```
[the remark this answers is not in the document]
hassaansaleem28: Done

hassaansaleem28:
@SpruhaCK I’ve updated the PR with the Makefile fixes. CI checks pass (the SonarCloud failure is a known false positive requiring admin approval).

I would love your feedback! If any changes are required, please just tag me.

SpruhaCK:
Hi @hassaansaleem28,
Currently reviewing the code. I think we also need to add trivy section in .github/dependabot.yml so that docker images will be tracked just like semgrep example (code line 43-46)

- package-ecosystem: docker
    directory: /docker/trivy
    schedule:
      interval: daily

hassaansaleem28:
@arkid15r All feedback addressed (CI updated). 
@SpruhaCK Dependabot now tracks the Trivy image. 
Ready for review.

arkid15r:

.github/workflows/run-ci-cd.yaml
```diff
@@ -151,11 +151,11 @@ jobs:
...
-        uses: aquasecurity/trivy-action@b6643a29fecd7f34b3597bc6acb0a98b03d33ff8
-        with:
-          scan-type: fs
-          skip-setup-trivy: true
-          trivy-config: trivy.yaml
+        run: |
+          make backend-security-scan-code
+          make frontend-security-scan-code
+          make backend-security-scan-secrets
+          make frontend-security-scan-secrets
```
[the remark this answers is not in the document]
hassaansaleem28: That was actually correct. Thanks!

.github/workflows/run-ci-cd.yaml
```diff
@@ -145,17 +145,12 @@ jobs:
...
-        uses: aquasecurity/trivy-action@b6643a29fecd7f34b3597bc6acb0a98b03d33ff8
-        with:
-          scan-type: fs
-          skip-setup-trivy: true
-          trivy-config: trivy.yaml
+        run: |
+          make backend-security-scan-code TRIVY_EXIT_CODE=1
+          make frontend-security-scan-code TRIVY_EXIT_CODE=1
+          make backend-security-scan-secrets TRIVY_EXIT_CODE=1
+          make frontend-security-scan-secrets TRIVY_EXIT_CODE=1
```
arkid15r: I don't think we need a separate target for that if we could control it via options similar to `TRIVY_EXIT_CODE` --  `--scanners vuln` depending on the env.

backend/Makefile
```diff
@@ -157,13 +157,60 @@ save-backup:
...
+backend-security-scan-code:
+	@echo "Scanning Backend Code (Vulns + Config)..."
+	@docker run --rm \
+		-v $(PWD)/backend:/app \
+		-v trivy-cache:/root/.cache/trivy \
+		$(TRIVY_IMAGE_TAG) \
+		fs \
+				--scanners vuln,config \
+		--exit-code $(TRIVY_EXIT_CODE) \
+		--severity CRITICAL,HIGH \
```
arkid15r: It looks like you deleted the configuration I pushed. Please revert.
Where possible params must be taken from trivy configuration file.

docker/frontend/Dockerfile
```diff
@@ -61,6 +65,18 @@ RUN cd /tmp && \
     rm -rf package tar-7.5.7.tgz && \
     grep -q 'version.*7.5.7' "${TAR_DIR}/package.json"
 
+# Fix CVE-2025-64756: Update npm's bundled glob to 11.1.0 in runner stage
```
arkid15r: Is this related to the PR?

frontend/next.config.ts
```diff
@@ -33,7 +33,7 @@ const nextConfig: NextConfig = {
   productionBrowserSourceMaps: true,
   serverExternalPackages: ['import-in-the-middle', 'require-in-the-middle'],
   transpilePackages: ['@react-leaflet/core', 'leaflet', 'react-leaflet', 'react-leaflet-cluster'],
-  ...(isLocal ? {} : { output: 'standalone' }),
+  output: 'standalone' ,
```
arkid15r: I believe this is not what we want for dev by default. You probably should set some flag during image local build process.

arkid15r (changes requested):

docker/frontend/Dockerfile
hassaansaleem28: while not part of the makefile refactor itself, this change is necessary because this pr enforces the trivy security scans.

My local frontend security scan was failing because trivy flags the current glob package (v10.4.5) as a High vulnerability. Since the goal of this PR is to enforce these checks, the pipeline effectively blocks itself without this fix.

should i remove this and do it in a separate pr depends on if you prefer, but it is currently a blocker for this one passing.

frontend/next.config.ts
hassaansaleem28: right
I added a flag for that.

hassaansaleem28:
@arkid15r Thanks for the detailed feedback. I've pushed updates to address all your points:

- Reverted Config Flag: I reverted the changes to --config trivy.yaml as you requested. The pipeline now correctly respects the config file precedence.

- Makefile Refactor: I removed the separate *-secrets targets. Instead, I introduced a TRIVY_SCANNERS environment variable. The Makefile now uses a single target that defaults to vuln,misconfig for local runs, and CI overrides this variable to include secret scanning. This keeps it DRY and flexible.

- Standalone Output: I removed the hardcoded output: 'standalone' from next.config.ts. I added a FORCE_STANDALONE build argument to the Dockerfile. Now, standalone mode is only enabled explicitly during the Docker build process, keeping the local dev environment standard.

- Trivy Config: I renamed security-checks to scanners and config to misconfig in trivy.yaml to resolve the deprecation warning flagged by the bot.

- Trivy Ignores: I added trivyignore.yaml to temporarily suppress specific high-severity vulnerabilities in cross-spawn and node-tar and reverted that code in frontend/Dockerfile. These are deep dependency issues that require complex package.json overrides. To keep this PR focused on fixing the CI/CD pipeline itself, I've marked them with TODO comments to be addressed in a dedicated frontend dependency PR.

Happy to make the changes if needed again. Learning alot in this.

Makefile
```diff
@@ -23,6 +27,9 @@ clean-docker: \
 	clean-docs-docker \
 	clean-frontend-docker
 
+clean-trivy-cache:
+	@rm -rf $(CURDIR)/.trivy-cache
```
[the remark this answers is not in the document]
arkid15r: this sounds reasonable, will check and fix in next PR

trivyignore.yaml
```diff
@@ -1,23 +1,9 @@
...
-# https://avd.aquasec.com/nvd/cve-2026-23745
-  - id: CVE-2026-23745
-  
-# node-tar: Arbitrary file overwrite via Unicode path collision (High)
-# https://avd.aquasec.com/nvd/cve-2026-23950
-  - id: CVE-2026-23950
-
-# node-tar: Arbitrary file creation via path traversal (High)
-# https://avd.aquasec.com/nvd/cve-2026-24842
-  - id: CVE-2026-24842
```
arkid15r: You didn't run `make check`

hassaansaleem28: right, I missed that step this time. I've made a note to strictly run make check before every push from now on. Thanks for headsup.

hassaansaleem28:
> This is good but could have been better -- see [8c0cf8b](https://github.com/OWASP/Nest/commit/8c0cf8b2d4dd8d48e2db6df2414e0225db1d0d5e)
> 
> It has some room for optimization but overall should work. Feel free to improve.

Thanks for merging and for the follow up commit! I just reviewed `8c0cf8b` it is a much, much cleaner approach. I appreciate you showing the optimized way.

I'll ensure the future PRs meets this standard.

PR #3764, opened 2026-02-03
perf: optimize isMerged resolver using Prefetch

### LINKED ISSUE #3658: Optimize `isMerged` field in `IssueNode`
When querying the `isMerged` field, we get N+1 queries.

Image

The current optimization hint is not enough for optimizing the query. We should use `Prefetch` instead.

- [ ] Yes
- [x] No

**Additional context**

Davda-James:
## Proposed change
Previously, recentIssues executed one base query, but resolving isMerged caused N additional queries to check merge state per issue. Using Prefetch removes this N+1 behavior by eagerly loading merged pull requests.

Resolves #3658 

Below is my local test after updating code:
```this is before updation```
earlier

```this is after updating```
optimized

## Checklist

### CODE REVIEW

**Comment by arkid15r** [comment]
@Davda-James please address CR's comments first.

File: backend/apps/github/api/internal/nodes/issue.py [hunk 1]
```diff
@@ -48,6 +48,13 @@ def labels(self, root: Issue) -> list[str]:
...
     def is_merged(self, root: Issue) -> bool:
         """Return True if this issue has at least one merged pull request."""
+        pull_requests = getattr(root, "prefetched_pull_requests", None)
+        if pull_requests is not None:
+            return any(
+                pr.state == "closed" and pr.merged_at is not None
+                for pr in pull_requests
+        )
+        # fallback to ORM if not prefetched
         return root.pull_requests.filter(state="closed", merged_at__isnull=False).exists()
```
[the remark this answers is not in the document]
Davda-James: done now every query path is optimized

backend/apps/github/api/internal/queries/issue.py
```diff
@@ -48,6 +50,14 @@ def recent_issues(
 
         queryset = queryset.filter(**filters)
 
+        queryset = queryset.prefetch_related(
+            Prefetch(
```
rudransh-shrivastava: Prefetching at the query level loads `pull_requests` for every `recent_issues` call.
I think it's better to keep the logic to the resolver using [Exists](https://docs.djangoproject.com/en/6.0/ref/models/expressions/#exists-subqueries) and [annotate](https://strawberry.rocks/docs/django/guide/optimizer#optimization-hints) instead of `Prefetch` in this case.

Davda-James: @rudransh-shrivastava mypy checks that are failing is not due to my change, it is failing even on main due to "Module datetime has no attribute UTC" issue. I have not looked into them. Also I used exists  and outerref instead of prefetch in new commit, pls review it if any suggestions.

backend/apps/github/api/internal/nodes/issue.py
```diff
@@ -45,10 +45,10 @@ def labels(self, root: Issue) -> list[str]:
         """Resolve label names for the issue."""
         return [label.name for label in root.labels.all()]
 
-    @strawberry_django.field(prefetch_related=["pull_requests"])
+    @strawberry_django.field
     def is_merged(self, root: Issue) -> bool:
         """Return True if this issue has at least one merged pull request."""
-        return root.pull_requests.filter(state="closed", merged_at__isnull=False).exists()
+        return bool(getattr(root, "is_merged", False))
```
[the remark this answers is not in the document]
Davda-James: added fallback orm when annotation is not present.

backend/apps/github/api/internal/queries/issue.py
```diff
@@ -48,6 +49,13 @@ def recent_issues(
 
         queryset = queryset.filter(**filters)
 
+        merged_pr_exists = PullRequest.objects.filter(
+            related_issues__id=OuterRef("pk"),
```
rudransh-shrivastava: This still runs for every `recent_issues` call (Example: query: recent_issues > body). Please move the annotate logic to the decorator/resolver instead. We might also not need the `if is_merged is not None` check then.

Davda-James: Should I make a class method like this  
```python
def get_queryset(cls, queryset, info):
  return queryset.annotate(
      is_merged=Exists(
          PullRequest.objects.filter(
              related_issues__id=OuterRef("pk"),
              state="closed",
              merged_at__isnull=False,
          )
      )
  )
```
in resolver 
then we would not need any check and resolver is_merged may just ```return bool(root.is_merged)``` @rudransh-shrivastava

rudransh-shrivastava: I was thinking something like this in the decorator itself:
```python
    @strawberry_django.field(
        annotate={
            "is_pr_merged": Exists(
                PullRequest.objects.filter(
                    merged_at__isnull=False,
                    related_issues=OuterRef("pk"),
                    state="closed",
                )
            )
        }
    )
```
@Davda-James

Davda-James: ok got it, done, pls review it if further modifications or optimization is required.

backend/apps/github/api/internal/nodes/issue.py
```diff
@@ -45,10 +47,20 @@ def labels(self, root: Issue) -> list[str]:
...
+                    related_issues=OuterRef("pk"),
+                    state="closed",
+                )
+            )
+        }
+    )
     def is_merged(self, root: Issue) -> bool:
         """Return True if this issue has at least one merged pull request."""
-        return root.pull_requests.filter(state="closed", merged_at__isnull=False).exists()
+        return bool(root.is_merged)
```
arkid15r: Do we actually need bool casting here?

Davda-James: Actually we don't require it because in earlier there was None case so I used bool casting but I don't think now we require it, let me remove it.

Davda-James: done

PR #3767, opened 2026-02-04
Fix N+1 query issue in releases field

### LINKED ISSUE #3662: Optimize `releases` field in `RepositoryNode`
When querying the releases field, we get N+1 queries.

Image

The current optimization hint is not enough for optimizing the query. We should use Prefetch instead.

- [ ] Yes
- [x] No

**Additional context**

HarshitVerma109:
## Proposed change

Resolves #3662 

Use `Prefetch` to fix N+1 query issue in releases field.

## Checklist


HarshitVerma109:

HarshitVerma109:

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -76,11 +78,22 @@ def recent_milestones(self, root: Repository, limit: int = 5) -> list[MilestoneN
...
+    @strawberry_django.field(
+        prefetch_related=[
+            lambda _: Prefetch(
+                "releases",
+                queryset=Release.objects.filter(
+                    is_draft=False,
+                    is_pre_release=False,
+                    published_at__isnull=False,
+                ).order_by("-published_at"),
+                to_attr="prefetched_releases",
```
ahmedxgouda, line 98: This optimizes N+1 for simple releases query, but what about nested relations? (e.g. author, repository).

ahmedxgouda (changes requested):
Thank you for working on this. We are near.

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -76,11 +78,26 @@ def recent_milestones(self, root: Repository, limit: int = 5) -> list[MilestoneN
...
+            lambda _: Prefetch(
+                "releases",
+                queryset=Release.objects.filter(
+                    is_draft=False,
+                    is_pre_release=False,
+                    published_at__isnull=False,
+                ).select_related(
+                    "author",
+                    "repository",
+                    "repository__organization",
```
ahmedxgouda: You don't need to specify `repository` in select_related here as you already specified it with organization. Also, what about nested relations of author?

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -76,11 +78,26 @@ def recent_milestones(self, root: Repository, limit: int = 5) -> list[MilestoneN
...
+                ).order_by("-published_at"),
+                to_attr="prefetched_releases",
+            )
+        ]
+    )
     def releases(self, root: Repository) -> list[ReleaseNode]:
         """Resolve recent releases."""
-        # TODO(arkid15r): rename this to recent_releases.
-        return root.published_releases.order_by("-published_at")[:RECENT_RELEASES_LIMIT]
+        return root.prefetched_releases[:RECENT_RELEASES_LIMIT]
```
ahmedxgouda: Can you move the limit to the prefetch? I think it will be faster. Also, why did you remove the comment?

ahmedxgouda (changes requested):

ahmedxgouda (changes requested):
You optimized N+1 queries problem for the repositories query, but the repository query is failing now.
Image

HarshitVerma109:
@ahmedxgouda, I think the problem is completely solved now.

ahmedxgouda:
@HarshitVerma109
This is spamming. I will take a look into it soon.
image

HarshitVerma109:
Sorry for the pings! I didn't intend to spam; I was just worried the message might have been missed. I’ll be more patient in the future and wait for your feedback. Thanks for your time!

ahmedxgouda:
@HarshitVerma109 It is not about the pings, it is about deleting and posting the same comment. Any way, sorry for the delays.

ahmedxgouda (approved):
LGTM

HarshitVerma109:

preeeetham:
Good catch on the N+1 for releases — the Prefetch with select_related and prefetch_related for project_set is a solid approach. A few things stood out though:

The lambda wrapping the Prefetch (`prefetch_related=[lambda _: Prefetch(...)]`) is unusual. PR #3770 fixes a similar N+1 in the same file for `issues` but uses Prefetch directly without a lambda. Having two different patterns in the same class for the same kind of optimization will be confusing for future contributors. If you need the lambda for deferred evaluation, it'd be worth adding a comment explaining why — otherwise removing it would keep things consistent.

The bigger concern is the `project_name` resolver change. The old code accessed `root.repository.project` (a direct FK/OneToOne), but the new code uses `root.repository.project_set.all()[0]`. This changes the semantics from a single deterministic FK lookup to picking the first item from a reverse queryset. If a repo ever has multiple projects, `[0]` picks whichever the DB returns first, which is non-deterministic without an explicit ordering. Adding `.order_by("name")` or `.first()` would make it explicit.

One more thing — not introduced by this PR, but since you're touching this line anyway: `project.name.lstrip(OWASP_ORGANIZATION_NAME)` is a subtle bug. Python's `lstrip()` strips individual *characters* from the set, not the prefix as a string. So `"OWASP Test".lstrip("OWASP")` strips all leading O, W, A, S, P characters and gives `" Test"` (with a leading space), which happens to look correct by coincidence. But something like `"OWASP WASPINATOR".lstrip("OWASP")` would give `"INATOR"` instead of `" WASPINATOR"`. Using `removeprefix()` (Python 3.9+) would fix this properly: `project.name.removeprefix(OWASP_ORGANIZATION_NAME)`.

PR #3770, opened 2026-02-04
fix: optimize issues field in RepositoryNode

### LINKED ISSUE #3660: Optimize `issues` field in `RepositoryNode`
When querying the issues field, we get N+1 queries.

Image

The current optimization hint is not enough for optimizing the query. We should use Prefetch instead.

- [ ] Yes
- [x] No

**Additional context**

Utkarsh-0304:
## Proposed change

Resolves #3660 

This PR optimises the `issues` field in `RepositoryNode`, eliminating the N + 1 query problem by utilising the Prefetch object. Also, adds related tests.

## Checklist


backend/apps/github/api/internal/queries/repository.py
```diff
@@ -31,7 +31,7 @@ def repository(
 
         """
         try:
-            return Repository.objects.select_related("organization").get(
+            return Repository.objects.select_related("organization").filter(
                 key__iexact=repository_key,
                 organization__login__iexact=organization_key,
             )
```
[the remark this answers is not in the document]
Utkarsh-0304: @cubic-dev-ai I explicitly want to return a queryset

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -45,11 +47,19 @@ class RepositoryNode(strawberry.relay.Node):
 
     organization: OrganizationNode | None = strawberry_django.field()
 
-    @strawberry_django.field(prefetch_related=["issues"])
+    @strawberry_django.field(
+        prefetch_related=[
+            Prefetch(
+                "issues",
+                queryset=Issue.objects.all().order_by("-created_at"),
```
arkid15r: Why do you need `all()` here?

backend/apps/github/api/internal/queries/repository.py
```diff
@@ -31,7 +31,7 @@ def repository(
 
         """
         try:
-            return Repository.objects.select_related("organization").get(
+            return Repository.objects.select_related("organization").filter(
```
arkid15r: What's the goal of this change?
It returns an iterable, not a single item.

backend/apps/github/api/internal/nodes/repository.py
Utkarsh-0304: Yes, it's not necessary. I will address this issue in my next commit.

backend/apps/github/api/internal/queries/repository.py
Utkarsh-0304: Previously, it was not functioning with `get()`, and the new prefetch object defined in the node was not being applied; that’s why I switched to `filter()` to return an iterable, which then worked fine.

I will now prefetch `issues`  within the `repository` query and revert `filter()` to `get()` in my next commit.

rudransh-shrivastava (changes requested):
Hey @Utkarsh-0304 , I tested this and it works.

Let's keep the prefetching logic to the decorator only to avoid redundant/unnecessary fetching:

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -45,11 +47,19 @@ class RepositoryNode(strawberry.relay.Node):
...
+                queryset=Issue.objects.order_by("-created_at"),
+                to_attr="recent_issues",
+            )
+        ]
+    )
     def issues(self, root: Repository) -> list[IssueNode]:
         """Resolve recent issues."""
         # TODO(arkid15r): rename this to recent_issues.
-        return root.issues.order_by("-created_at")[:RECENT_ISSUES_LIMIT]
+        return getattr(root, "recent_issues", [])[:RECENT_ISSUES_LIMIT]
```
rudransh-shrivastava: Is `getattr` required here? The decorator should always ensure the attribute `recent_issues` exists.

I think `getattr` was added because the prefetching logic exists in `repository` query too, is that a correct assumption?

backend/apps/github/api/internal/queries/repository.py
```diff
@@ -31,9 +33,19 @@ def repository(
...
         """
         try:
-            return Repository.objects.select_related("organization").get(
-                key__iexact=repository_key,
-                organization__login__iexact=organization_key,
+            return (
+                Repository.objects.select_related("organization")
+                .prefetch_related(
+                    Prefetch(
+                        "issues",
```
rudransh-shrivastava: Prefetching here seems redundant as the field decorator already handles this. Also, prefetching here will always prefetch `issues` even when they're not queried in GraphQL. 

The decorator approach looks good to me.

rudransh-shrivastava (changes requested):
Forgot to add the comments:

backend/apps/github/api/internal/nodes/repository.py
Utkarsh-0304: I am using `get_attr` to handle `AttributeError` if prefetch fails for some reason, so it falls back to an empty list.

backend/apps/github/api/internal/queries/repository.py
Utkarsh-0304: The repository query results in an empty list (due to `get_attr`) when not prefetching here.

rudransh-shrivastava: but we are prefetching in the decorator right? I tried removing this part locally and it seems to work.

Utkarsh-0304: image
image

rudransh-shrivastava: wait yes, I think its correct, didn't see this: https://github.com/OWASP/Nest/pull/3843/changes

rudransh-shrivastava (dismissed):
Thank you 👍

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -45,11 +47,19 @@ class RepositoryNode(strawberry.relay.Node):
 
     organization: OrganizationNode | None = strawberry_django.field()
 
-    @strawberry_django.field(prefetch_related=["issues"])
+    @strawberry_django.field(
+        prefetch_related=[
+            Prefetch(
+                "issues",
+                queryset=Issue.objects.order_by("-created_at"),
```
ahmedxgouda: What about nested relations of issue? They need to be optimized also (e.g. repositoryName).

backend/apps/github/api/internal/nodes/repository.py
ahmedxgouda: Let's move the limit to the prefetch queryset

ahmedxgouda (changes requested):
Thank you. Please see the below comments

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -45,11 +48,23 @@ class RepositoryNode(strawberry.relay.Node):
...
+                "issues",
+                queryset=Issue.objects.select_related("repository", "repository__organization")
+                .prefetch_related(
+                    Prefetch("labels", queryset=Label.objects.all(), to_attr="label_names")
+                )
+                .order_by("-created_at")[:RECENT_ISSUES_LIMIT],
+                to_attr="recent_issues",
+            )
+        ]
+    )
```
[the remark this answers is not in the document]
Utkarsh-0304: @coderabbitai, however, it is functioning and slicing effectively even when inside queryset.

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -45,11 +47,21 @@ class RepositoryNode(strawberry.relay.Node):
 
     organization: OrganizationNode | None = strawberry_django.field()
 
-    @strawberry_django.field(prefetch_related=["issues"])
+    @strawberry_django.field(
+        prefetch_related=[
+            Prefetch(
+                "issues",
+                queryset=Issue.objects.select_related("repository", "repository__organization")
```
ahmedxgouda: repository here is select_related is redundant. Also what about `author`?

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -45,10 +47,22 @@ class RepositoryNode(strawberry.relay.Node):
 
     organization: OrganizationNode | None = strawberry_django.field()
 
-    @strawberry_django.field(prefetch_related=["issues"])
+    @strawberry_django.field(
+        prefetch_related=[
+            Prefetch(
+                "issues",
+                queryset=Issue.objects.select_related("repository__organization")
```
ahmedxgouda: What about `author`?

ahmedxgouda (changes requested):

ahmedxgouda:
I made some updates to the code and removed the `Prefetch` of the repository query @Utkarsh-0304. Please see the comments and ensure that the code works after the updates.

ahmedxgouda:
@Utkarsh-0304 Please look at cubic comments!

backend/apps/github/api/internal/nodes/repository.py
```diff
@@ -45,10 +47,22 @@ class RepositoryNode(strawberry.relay.Node):
...
+                to_attr="recent_issues",
+            )
+        ]
+    )
     def issues(self, root: Repository) -> list[IssueNode]:
         """Resolve recent issues."""
         # TODO(arkid15r): rename this to recent_issues.
+        if hasattr(root, "recent_issues"):
+            return getattr(root, "recent_issues", [])
         return root.issues.order_by("-created_at")[:RECENT_ISSUES_LIMIT]
```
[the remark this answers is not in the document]
ahmedxgouda: @Utkarsh-0304 The fallback here is for the repository query not for repositories, so we don't need to do prefetch_related or select_related here.

ahmedxgouda (dismissed):
LGTM

Utkarsh-0304:
Hi @arkid15r @kasya, this pull request is now ready for review.

preeeetham:
Nice work on this N+1 fix — the Prefetch with select_related for author and repository is the right approach. A couple things I'd flag:

The queryset inside the Prefetch uses `[:RECENT_ISSUES_LIMIT]` to slice directly at the DB level, which is efficient, but sliced querysets in Django can't be filtered or annotated further. If strawberry-django or any middleware tries to re-filter the prefetched queryset downstream, it'll raise an `AssertionError: Cannot filter a query once a slice has been taken`. Might be safer to apply the limit in the resolver method itself rather than inside the Prefetch, unless you've confirmed nothing touches the queryset after prefetching.

The shared `LABELS_PREFETCH` with `to_attr="label_names"` is a clean idea for reuse across IssueNode and RepositoryNode, but the name is slightly misleading since it stores `Label` objects, not plain strings. Something like `prefetched_labels` would make the code more self-documenting and avoid confusion for anyone reading `getattr(root, "label_names", [])` and expecting a list of strings.

Also worth noting — the fallback path when `recent_issues` isn't set still does `root.issues.order_by(...)[:LIMIT]` without any select_related or prefetch, so any query that doesn't go through the Prefetch decorator still hits the original N+1. Might be worth mentioning that as a known limitation in the PR description, or adding select_related in the fallback too.

PR #3774, opened 2026-02-04
Add user locking during API key creation

### LINKED ISSUE #3634: race conditions could result in more api keys created for a user than is allowed
**Describe the bug**
The `select_for_update()` is called on rows already existing. so the transaction won't prevent other transactions from runnign at the same time and getting the same count. This could lead to a user being able to create more than maximum keys allowed.

**Expected behavior**
We should probably lock the user instead of the returned rows for active keys. that way another transaction just wont be able to proceed if one is already running.

- [x] Yes
- [] No

theAnuragMishra:
## Proposed change

Resolves #3634 

currently the way we count the number of api keys only locks the api key row. but the count itself isn't dependent on the locked rows but the whole user. the pr updates the code to lock the user.

## Checklist


theAnuragMishra:
@coderabbitai review

arkid15r (changes requested):
I don't think this is true

Image

arkid15r:
Fixed in https://github.com/OWASP/Nest/pull/3981

PR #3786, opened 2026-02-05
CI: Validate Docker Compose files in CI and locally

### LINKED ISSUE #3748: [CI/CD] Validate Docker Compose files in CI
The repository has multiple Docker Compose files (production, staging, proxy, local), but CI currently does not validate them. Invalid YAML syntax or schema errors are only caught during deployment or manual testing, which creates a risk of broken builds in production/staging.

I propose adding a CI job (or step) that runs `docker compose -f <file> config --quiet` for each Compose file located in `docker-compose/**`. This will validate the syntax and schema automatically on every PR.

- [x] Yes
- [ ] No

**Additional context**
This is a fast, low-cost check that prevents broken deploys due to simple Compose errors. I plan to implement this by adding a new workflow or extending the existing CI pipeline.

hassaansaleem28:
## Proposed change

Resolves #3748 

### Description 

This PR adds a local Make target and CI job to validate all Docker Compose files on every PR. It prevents invalid compose changes from reaching deployments. To test run make validate-compose locally.

### Changes:
- New Make target to validate compose files.
- New CI job to run the validation target.

### Notes
- The target creates temporary placeholder env files for production and staging compose files only when missing, then cleans them up.
- If the env files already exist locally, the target uses them as-is and does not overwrite them.
- Uses the same logic locally and in CI.

## Checklist

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -112,6 +112,10 @@ load-data-fuzz:
 	@echo "Loading Nest fuzz data"
 	@CMD="pg_restore -U nest_user_fuzz -d nest_db_fuzz < ./backend/data/nest.dump" $(MAKE) exec-db-command-fuzz
 
+load-data-staging:
+	@echo "Loading Nest staging data"
+	@CMD="pg_restore -U nest_user_staging -d nest_db_staging < ./backend/data/nest.dump" $(MAKE) exec-db-command-fuzz
+
 merge-migrations:
 	@CMD="python manage.py makemigrations --merge" $(MAKE) exec-backend-command
 
--- a/backend/apps/github/api/internal/nodes/issue.py
+++ b/backend/apps/github/api/internal/nodes/issue.py
@@ -2,10 +2,12 @@
 
 import strawberry
 import strawberry_django
+from django.db.models import Exists, OuterRef
 
 from apps.github.api.internal.nodes.pull_request import PullRequestNode
 from apps.github.api.internal.nodes.user import UserNode
 from apps.github.models.issue import Issue
+from apps.github.models.pull_request import PullRequest
 
 
 @strawberry_django.type(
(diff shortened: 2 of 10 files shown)


hassaansaleem28:
@arkid15r one thing i want to ask is currently validate-compose creates placeholders only for the 4 env files referenced in production/staging (.env.backend, .env.cache, .env.db, .env.frontend). We can keep this simple approach or upgrade it to parse env_file entries directly from compose files (fwill work even if new env files are added). Which approach do you prefer?

Makefile
```diff
@@ -63,6 +65,30 @@ run:
 	docker compose -f docker-compose/local/compose.yaml --project-name nest-local build && \
 	docker compose -f docker-compose/local/compose.yaml --project-name nest-local up --remove-orphans
 
+ENV_FILES := .env.backend .env.cache .env.db .env.frontend
```
arkid15r, line 68: This looks quite cumbersome. I'm not sure we need it that much.
We can consider reusing existing pre-commit hooks but it looks like our `check-yaml` covers enough.

hassaansaleem28: Noted. But see, check-yaml validates YAML syntax only, while docker compose config also validates compose schema and env_file presence. Happy to simplify.
Would you prefer we drop Compose validation entirely and rely on check-yaml, or keep a minimal Compose check (no env placeholders)?

arkid15r: I think it has never been an issue for us so it's safe to skip this check or at least reuse existing pre-commit hooks (I saw a couple).

hassaansaleem28: Understood. 

I'll switch to the **[`iamthefij/docker-pre-commit`](https://github.com/iamthefij/docker-pre-commit)** hook. 
I'll push the update shortly.

hassaansaleem28: I have tried with this docker hook that validates compose files, but it doesn't work for envs present in staging and production compose files 

Screenshot 2026-02-06 111533

arkid15r: I believe we can keep it simple and just don't introduce this check.

arkid15r:
Let's revisit this if we face a real problem related to docker compose files linting.

PR #3788, opened 2026-02-05
fix: optimize interested users field in IssueNode

### LINKED ISSUE #3659: Optimize `interestedUsers` field in `IssueNode`
When querying the `interestedUsers` field, we get N+1 queries.

Image

The current optimization hint is not enough for optimizing the query. We should use `Prefetch` instead.

- [ ] Yes
- [x] No

**Additional context**

Utkarsh-0304:
## Proposed change

Resolves #3659 

This PR optimises the interestedUsers field in RepositoryNode by eliminating the N + 1 query problem through the use of the Prefetch object.

## Checklist

### DIFF
--- a/backend/tests/apps/github/api/internal/nodes/issue_test.py
+++ b/backend/tests/apps/github/api/internal/nodes/issue_test.py
@@ -118,7 +118,7 @@ class TestIssueNode(GraphQLNodeBaseTest):
         assert not result
 
     def test_interested_users(self):
-        """Test interested_users field returns list of users."""
+        """Test interested_users field returns list of users from prefetched interests_users."""
         mock_issue = Mock()
         mock_user1 = Mock()
         mock_user1.login = "user1"
@@ -130,12 +130,7 @@ class TestIssueNode(GraphQLNodeBaseTest):
         mock_interest2 = Mock()
         mock_interest2.user = mock_user2
 
-        mock_queryset = Mock()
-        mock_queryset.select_related.return_value.order_by.return_value = [
-            mock_interest1,
-            mock_interest2,
-        ]
-        mock_issue.participant_interests = mock_queryset
+        mock_issue.interests_users = [mock_interest1, mock_interest2]
 
         field = self._get_field_by_name("interested_users", IssueNode)
         result = field.base_resolver.wrapped_func(None, mock_issue)


backend/apps/github/api/internal/nodes/issue.py
```diff
@@ -62,12 +63,15 @@ def is_merged(self, root: Issue) -> bool:
...
-        """Return all users who have expressed interest in this issue."""
-        return [
-            interest.user
-            for interest in root.participant_interests.select_related("user").order_by(
-                "user__login"
+    @strawberry_django.field(
+        prefetch_related=[
+            Prefetch(
+                "participant_interests",
+                queryset=IssueUserInterest.objects.select_related("user").order_by("user__login"),
```
ahmedxgouda: It is optimized now for this simple query:
Image
But we still have N+1 queries for a query like this:

Image

ahmedxgouda (dismissed):
LGTM.

backend/apps/github/api/internal/nodes/user.py
```diff
@@ -28,25 +41,10 @@
 class UserNode:
     """GitHub user node."""
 
-    @strawberry_django.field(prefetch_related=["user_badges"])
-    def badge_count(self, root: User) -> int:
-        """Resolve badge count."""
-        return root.user_badges.filter(is_active=True).count()
-
-    @strawberry_django.field(prefetch_related=["user_badges__badge"])
+    @strawberry_django.field(prefetch_related=[USER_BADGES_PREFETCH])
```
kasya, line 44: I removed this resolver completely - the only place where we show badge count get's that from Algolia.

backend/apps/github/api/internal/queries/user.py
```diff
@@ -51,4 +51,8 @@ def user(
             User or None: The user object if found, otherwise None.
 
         """
-        return User.objects.filter(has_public_member_page=True, login=login).first()
+        return (
+            User.objects.filter(has_public_member_page=True, login=login)
+            .prefetch_related(USER_BADGES_PREFETCH)
```
kasya, line 56: We only show badges on the single user page that is resolved here. With just adding prefetch to a @strawberry_django.field this was not working and not adding it here - this was always going into a fallback for `[]`. 

Prefetch on the field will be triggered when we fetch badges via another connection (example: `Issue.author.badges` if we'd ever need that).

Any optimization needs to be tested to make sure returned results are the same as before.

frontend/src/server/queries/userQueries.ts
```diff
@@ -1,24 +1,5 @@
 import { gql } from '@apollo/client'
 
-export const GET_LEADER_DATA = gql`
```
kasya, line 3: Dead code.

kasya (approved):
Hi @Utkarsh-0304 !

I pushed changes to this PR to address some of my concerns and issues that I found.

I believe we'll need to look more into other ways to optimize backend calls without prefetches. My main concern is that some of these cases are only "theoretical" and are only possible when you go into GraphQL explorer. For example, there's no query where we'd need to have badges on `interested_users` for example. 

But since this PR addresses the issues and also has some clean up that was needed - we'd push this as is 👌🏼 
Thank you!

backend/apps/github/api/internal/queries/user.py
Utkarsh-0304: Yes, I realised I forgot to handle the scenario involving only users. 

I will ensure to test that the returned results remain consistent in future PRs.

Thanks for your review, @kasya

Utkarsh-0304:
Hi @arkid15r, I noticed that this [issue](https://github.com/OWASP/Nest/issues/3659) related to this PR is not assigned to the team `GSOC 2026`. Does this mean that this won't count towards the leaderboard?

image

Issue #3741, opened 2026-02-02
Light mode breadcrumb text color mismatch causing inconsistent UI


mrkeshav-05:
**Describe the bug**
In light mode, the breadcrumb navigation shows a color mismatch between items.

**To Reproduce**

1. Go to `/projects/[projectkey]` in light mode.

**Are you going to work on fixing this?**

- [x] Yes
- [ ] No

**Screenshots**
<img width="461" height="57" alt="Image" src="https://github.com/user-attachments/assets/d45c1277-f940-45b4-b20d-538465fd7290" />

moizulfiqar:
Hi @arkid15r, I would like to work on this issue, kindly assign it to me.

PR #3808, opened 2026-02-06
Light mode breadcrumb text color mismatch causing inconsistent UI

mrkeshav-05:
### Proposed change
This PR solves the issue of colour mismatch between items.
Resolves: #3741 

image

## Checklist

### DIFF
--- a/frontend/__tests__/unit/components/PageLayout.test.tsx
+++ b/frontend/__tests__/unit/components/PageLayout.test.tsx
@@ -65,4 +65,37 @@ describe('PageLayout', () => {
       expect(screen.getByText('OWASP ZAP')).toBeInTheDocument()
     })
   })
+
+  describe('breadcrumbClassName', () => {
+    test('applies default breadcrumb class when breadcrumbClassName is not passed', async () => {
+      ;(usePathname as jest.Mock).mockReturnValue('/projects/zap')
+
+      renderWithProvider(
+        <PageLayout title="OWASP ZAP">
+          <div>Content</div>
+        </PageLayout>
+      )
+
+      await screen.findByRole('navigation', { name: 'breadcrumb' })
+      const breadcrumbNav = screen.getByRole('navigation', { name: 'breadcrumb' })
+      const wrapper = breadcrumbNav.parentElement?.parentElement
+      expect(wrapper).toHaveClass('bg-white')
+      expect(wrapper).toHaveClass('dark:bg-[#212529]')
+    })
+
+    test('applies custom breadcrumb class when breadcrumbClassName is passed', async () => {
+      ;(usePathname as jest.Mock).mockReturnValue('/projects/zap')
+
+      renderWithProvider(
+        <PageLayout title="OWASP ZAP" breadcrumbClassName="custom-bg">
+          <div>Content</div>
+        </PageLayout>
+      )
+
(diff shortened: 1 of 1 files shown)


frontend/src/components/BreadCrumbsWrapper.tsx
```diff
@@ -13,8 +13,16 @@ export default function BreadCrumbsWrapper() {
 
   if (pathname === '/' || items.length <= 1) return null
 
+  const isDetailPage =
+    /^\/(projects|chapters|organizations|members|committees|board|community\/snapshots)\/[^/]+$/.test(
+      pathname
+    )
+  const backgroundClass = isDetailPage 
+    ? 'bg-white dark:bg-[#212529]'
+    : 'bg-[#F5F6FC] dark:bg-[#212529]'
```
kasya: @mrkeshav-05 why do we need this check? What pages would go into that fallback? 🤔 

I can see mentorship pages are not included in that test check, which means they still show "old" background, but I'm mostly wondering if even need that check at all.

kasya:
@mrkeshav-05 I can see you updated the PR with *main*.. but I also left some questions above ⬆️ Could you get to those when you have a moment?

frontend/src/components/BreadCrumbsWrapper.tsx
mrkeshav-05: Hi @kasya ,
The check exists because some pages (like detail card pages) use a white background (`bg-white dark:bg-[#212529]`), while the breadcrumb currently defaults to `bg-[#F5F6FC] dark:bg-[#212529]`. This mismatch causes the visible color difference.

mrkeshav-05: Pages using `CardDetailsPage.tsx` (for example project, chapter, org, etc.) are meant to align with the white background, which is why I added this condition.

You’re right that mentorship (and also the project dashboard) are not covered yet — those should be included or handled consistently. 
I also agree the name `isDetailPage` may not be ideal. I can rename it to better reflect that it’s checking for pages using the CardDetails layout.

mrkeshav-05:
I think I mistakenly requested a review. 
I’m still working on this once it’s completed, I’ll make it ready for review.

frontend/src/components/BreadCrumbsWrapper.tsx
mrkeshav-05: I think instead of checking routes inside `BreadCrumbsWrapper`(since it’s not feasible to include all URL checks), it would be cleaner if the page or layout decides the breadcrumb background.

If the breadcrumb container is made transparent, it will automatically inherit the background color of the parent layout, keeping it consistent without any route-based checks.

mrkeshav-05: Hi @kasya ,
Could you please suggest how you’d like me to proceed with this?

kasya: @mrkeshav-05 I like this 👍🏼 :
> If the breadcrumb container is made transparent, it will automatically inherit the background color of the parent layout, keeping it consistent without any route-based checks.

Can we try this approach?

kasya (approved):
@mrkeshav-05 thanks for working on this! 👍🏼 
Just pushed some code to address code duplication.

PR #3844, opened 2026-02-08
Resolve case-sensitivity in QueryParser to support Chapters/Members search

### LINKED ISSUE #3791: [Blocker for #3777] Fixing QueryParser case-sensitivity to support Chapters/Members search
**Describe the bug**
The current QueryParser implementation is a bit too aggressive with .lower(). Right now, it converts the entire query string to lowercase at the very start of the parse() method. 

While we want field names (like stars: or name:) to be case-insensitive for ease of use, forcing the *values* to lowercase is a problem. If a user is searching for a member in a specific city or a company with proper casing (e.g., "New York" or "OWASP"), the parser destroys that casing before it ever hits the database. As we extend this to Chapters and Members in #3777 , this is going to cause issues where case-sensitive filtering might be needed or where we just want to maintain clean data.

**To Reproduce**
 
1. Initialize QueryParser in a test or shell.
2. Pass a string with mixed casing: location:"New York".
3. Observe the parsed output dictionary.
4. The value is returned as "new york" instead of the original "New York".
5. The same happens with unquoted values like company:OWASP becoming owasp.

**Expected behavior**
The parser should normalize the "keys" (the fields) to lowercase so the logic can map them correctly, but it should preserve the "values" exactly as the user typed them. If I type Name:Sai, the parser should return field: "name" and value: "Sai".

- [x] Yes
- [ ] No

**Additional context**
I found this while working on the extension of structured search to core entities (#3777). It wasn't a blocker for numeric fields like stars, but it's a regression for string-based fields like Chapter regions or Member names. 

I also plan to refine the regex in _split_tokens during this refactor to ensure multi word strings like company:"Google LLC" are tokenized reliably without tripping over the internal space. Fixing this now will give us a much more stable foundation for the rest of the search expansion.

If we don't fix this, users searching for company:OWASP might get zero results if the underlying database lookup is performing an exact case match, leading to a confusing user experience where the data exists but the search 'misses' it.

saichethana28:
## Proposed change

Resolves #3791 

Fixed the logic in `query_parser.py` to resolve the reported blocker regarding search query processing. The changes ensure that query tokens are handled with proper case integrity and logical structure, preventing the parser from failing on specific edge cases.

**Changes included:**
* Refactored token handling in `backend/apps/common/search/query_parser.py`.
* Added and updated unit tests in `backend/tests/apps/common/search/query_parser_test.py` to verify the fix and ensure no regressions.
* Verified that all backend tests pass and coverage is maintained.

## Checklist


saichethana28:
Hi @arkid15r, could you please review this? I've fixed the QueryParser case-sensitivity for Chapters/Members. Once this blocker is resolved, I’m ready to start extending structured search to other entities

backend/apps/common/search/query_parser.py
```diff
@@ -170,8 +170,6 @@ def parse(self, query: str) -> list[dict]:
         """
         conditions: list[dict] = []
         query = query.strip()
-        if not self.case_sensitive:
```
arkid15r, line 173: Would it help to use `case_sensitive=True` instead?

saichethana28:

saichethana28:

backend/apps/common/search/query_parser.py
saichethana28: Implemented as suggested. Added case_sensitive to the parser and enabled it for structured search

saichethana28:
@arkid15r, let me know if this looks good to you or if any changes are needed

saichethana28:
> Would this work?

Yes, this works. Moving the .lower() call to the start of the parser is a much cleaner optimization. Since structured_search.py explicitly passes case_sensitive=True, it correctly skips that block and preserves the casing for Chapter names. The refactor looks solid

arkid15r (approved):

Issue #3860, opened 2026-02-08
Incorrect Error Message for Duplicate Module Name


mrkeshav-05:
**Describe the bug**
When a user tries to create a module with a name that already exists in the same program, the system shows a technical GraphQL/database error message.

This message is confusing for users because:
- It talks about “unique constraint” and database keys
- It does not clearly say that the module name is already used
- It looks like an error related to the program, not the module

Instead of showing a technical error, the system should show a simple and clear message, like:
> “This module name already exists in this program.”

**Are you going to work on fixing this?**

- [x] Yes
- [ ] No

**Screenshots**
| Current Behaviour | Expected Behaviour |
|--------|--------|
| <img width="382" height="163" alt="Image" src="https://github.com/user-attachments/assets/dc5eb905-64c2-41e5-a755-5b7a4ceb8daf" /> | <img width="340" height="111" alt="Image" src="https://github.com/user-attachments/assets/0f9fede0-6de9-4232-8e78-1ff259229734" /> |

PR #3879, opened 2026-02-09
Incorrect Error Message for Duplicate Module Name

mrkeshav-05:
## Proposed change
This PR fixes the incorrect error message shown when a duplicate module name is used and displays a clear message:

> “This module name already exists in this program.”

Resolves: #3860 

## Checklist

### DIFF
--- a/backend/apps/mentorship/api/internal/mutations/module.py
+++ b/backend/apps/mentorship/api/internal/mutations/module.py
@@ -89,6 +89,18 @@ def _validate_module_dates(started_at, ended_at, program_started_at, program_end
     return started_at, ended_at
 
 
+def _handle_module_save_integrity_error(exc: IntegrityError) -> None:
+    """Translate module save IntegrityError to GraphQLError for known constraints."""
+    error_message = str(exc)
+    if "unique_module_key_in_program" in error_message:
+        msg = "This module name already exists in this program."
+        raise GraphQLError(
+            msg,
+            extensions={"code": "VALIDATION_ERROR", "field": "name"},
+        ) from exc
+    raise exc
+
+
 @strawberry.type
 class ModuleMutation:
     """GraphQL mutations related to the mentorship Module model."""
@@ -131,14 +143,7 @@ class ModuleMutation:
                     project=project,
                 )
         except IntegrityError as e:
-            error_message = str(e)
-            if "unique_module_key_in_program" in error_message:
-                msg = "This module name already exists in this program."
-                raise GraphQLError(
-                    msg,
-                    extensions={"code": "VALIDATION_ERROR", "field": "name"},
-                ) from e
-            raise
+            _handle_module_save_integrity_error(e)
 
         if module.experience_level not in program.experience_levels:
             program.experience_levels.append(module.experience_level)
@@ -409,14 +414,7 @@ class ModuleMutation:
             with transaction.atomic():
                 module.save()
         except IntegrityError as e:
-            error_message = str(e)
-            if "unique_module_key_in_program" in error_message:
-                msg = "This module name already exists in this program."
-                raise GraphQLError(
-                    msg,
-                    extensions={"code": "VALIDATION_ERROR", "field": "name"},
-                ) from e
-            raise
+            _handle_module_save_integrity_error(e)
 
         if module.experience_level not in module.program.experience_levels:
             module.program.experience_levels.append(module.experience_level)


mrkeshav-05:
@cubic-dev-a review it again.

mrkeshav-05:
Hi @kasya , PR is ready for review.

mrkeshav-05:
@cubic-dev-ai review this PR.

frontend/src/components/ModuleForm.tsx
```diff
@@ -104,6 +121,38 @@ const ModuleForm = ({
...
+      if (!name.trim()) {
+        return undefined
+      }
+
+      try {
+        const { data } = await client.query({
+          query: GetModulesByProgramDocument,
+          variables: { programKey },
+          fetchPolicy: 'network-only',
+        })
```
kasya: This works with making another request to the backend to see if there's already a module with this name. However, there's actually a better solution. the only problem - it needs way more setup. I did consider doing that at some point but just haven't gotten to it yet.

GraphQL can already raise an error on name not being unique. The only problem we don't have GraphQL error handlers to pass in to frontend. 

How this whole thing should work (example code):
backend raises an error...
```
except IntegrityError as e:
    error_message = str(e)
    if "unique_module_key_in_program" in error_message:
        raise ValidationError(
            message="This module name already exists in this program",
            field="name"  # This needs to be included in error extensions
        ) from e
    raise
```
...GraphQL backend formats it to consume on the frontend:
```
# In your GraphQL schema/error handling
def format_error(error):
    if isinstance(error.original_error, ValidationError):
        return {
            "message": error.message,
            "extensions": {
                "code": "VALIDATION_ERROR",
                "field": error.original_error.field,  # Pass field through
            }
        }
    return error
```
..then we parse error on the frontend with error handler..
```
function handleGraphQLError(error: ApolloError, form: any) {
  error.graphQLErrors.forEach(err => {
    // Access the field from extensions
    if (err.extensions?.code === 'VALIDATION_ERROR') {
      const field = err.extensions.field  // "name"
      const message = err.message  // "This module name already exists..."
      
      if (field) {
        // Set error on specific field
        form.setError(field, {
          type: 'server',
          message: message
        })
      } else {
        // Fallback to root error
        form.setError('root', {
          type: 'server',
          message: message
        })
      }
    }
  })
}
```
... and apply it to form on catching error.
```
} catch (error) {
      if (error instanceof ApolloError) {
        handleGraphQLError(error,  form) 
      } else {
        // Unexpected error
        setError('root', {
          type: 'server',
          message: 'An unexpected error occurred'
        })
      }
    }
  }
```

But, as I said - that would require making a lot more updates. I don't believe we have error formatters for GraphQL right now 🤔 

Would you like to update this PR to try this approach or should we continue with your current solution? Personally I would prefer "the correct" way

kasya (changes requested):
Hi @mrkeshav-05 !
This works with your current solution, but I left a comment about how normally these errors should be handled in GraphQL. Please check my message ⬇️  and let me know your thoughts.

frontend/src/components/ModuleForm.tsx
mrkeshav-05: Yes, I’ll try to go with the more optimized approach you described. It also looks like `program.tsx` has the same issue of checking twice in the DB. I’d love to address that as well. 

I’ll get back to you once I’ve updated the PR.

frontend/src/utils/helpers/handleGraphQLError.ts
```diff
@@ -0,0 +1,44 @@
...
+        unmappedErrors.push(gqlError.message)
+      }
+    }
+  } else if (error instanceof Error && error.message) {
+    unmappedErrors.push(error.message)
+  }
+
+  const hasFieldErrors = Object.keys(fieldErrors).length > 0
+  return { fieldErrors, hasFieldErrors, unmappedErrors }
+}
```
kasya, line 44: Can we make this more reusable? Like using `ValidationError` or something similar and not `fieldErrors` naming? 

The idea is that this will be used throughout the code base for any graphql errors eventually, and should work with different kinds of errors passed in from the backend. So we better come up with the most appropriate names right away.

backend/apps/mentorship/api/internal/mutations/module.py
```diff
@@ -96,18 +98,29 @@ def create_module(self, info: strawberry.Info, input_data: CreateModuleInput) ->
...
+                    program=program,
+                    project=project,
+                )
+        except IntegrityError as e:
+            error_message = str(e)
+            if "unique_module_key_in_program" in error_message:
+                msg = "This module name already exists in this program."
+                raise GraphQLError(
+                    msg,
+                    extensions={"code": "VALIDATION_ERROR", "field": "name"},
```
kasya: I don't see you using 'VALIDATION_ERROR' anywhere right now 🤔

mrkeshav-05: I’ve replaced it with `VALIDATION_ERROR` to make the handling more generic and consistent.

preeeetham:
Good improvement — showing inline validation errors instead of generic toasts is much better UX, and the GraphQL error extraction utility is a nice reusable pattern. A few things I noticed:

The IntegrityError handling in `module.py` matches the constraint name via `if "unique_module_key_in_program" in error_message:`. This works on PostgreSQL where the constraint name appears in the error string, but it's fragile — the format is DB-engine specific, and if the constraint name ever changes you'll get silent failures where the raw IntegrityError bubbles up. A more robust approach would be to do a pre-check before creating: `if Module.objects.filter(program=program, key=slugify(name)).exists(): raise GraphQLError(...)`. That way you avoid relying on DB-specific error string parsing entirely.

Also, the exact same try/except/if block is duplicated in both `create_module` and `update_module` (~20 lines each). If you add more constraints or change the error message, you'd need to update both places. Extracting it into a helper or context manager would keep things DRY.

On the frontend side, `extractGraphQLErrors` returns empty results for non-Error values like strings and numbers — `{ hasValidationErrors: false, validationErrors: {}, unmappedErrors: [] }`. This means if someone does `throw "something failed"` (which happens more often than it should), the user gets zero feedback. Treating string errors as unmapped errors would be safer.

One last thing — the EditModule test at line ~324 was weakened. The old test verified that a permission denied error showed a specific toast message. The new test just checks `expect(mockUpdateModule).toHaveBeenCalled()` without verifying any error handling. Since permission errors aren't validation errors with a `field` extension, they'd fall through to `unmappedErrors` in the new system, but that path isn't being tested anymore.

PR #3953, opened 2026-02-15
fix: Handle NoneType page_tree in get_urls to prevent crash

### LINKED ISSUE #3922: AttributeError in get_urls() when page_tree is None
**Bug Description**
In `backend/apps/owasp/scraper.py`, the `get_urls()` method (line 89) does not check whether `self.page_tree` is None before calling `.xpath()` on it. When `self.page_tree` is `None`, which happens whenever the HTTP request fails and returns a 404, or produces a parser error. `get_urls()` will raise an `AttributeError: 'NoneType' object has no attribute 'xpath'.`
```
def get_urls(self, domain=None):
        """Return scraped URLs."""
        return set(
            self.page_tree.xpath(f"//div[@class='sidebar']//a[contains(@href, '{domain}')]/@href")
            if domain is not None
            else self.page_tree.xpath("//div[@class='sidebar']//a/@href")
        )
```
The sibling method `get_audience()` (line 55) correctly handles this case by returning [] early

**Expected behavior**
`get_urls` should check whether self.page_tree is None and return `set()` early if it is.

- [X] Yes
- [ ] No

dhaval314:
## Proposed change

In scraper.py, if self.page_tree is None (which happens if requests fails or the parser errors out), the get_urls method attempts to call .xpath() on it immediately, causing a crash.

The sibling method get_audience() already handles this correctly by returning [], but get_urls() was missing the guard clause.

Resolves #3922

Added a guard clause at the start of get_urls():
```
if self.page_tree is None:
    return set()
```

## Checklist


.github/ansible/.ansible/.lock
ahmedxgouda, line 1: What is this file for?

dhaval314: My bad, the file should get deleted automatically after the checks but it likely got stuck and committed by mistake.

PR #3975, opened 2026-02-17
fix: replace empty ResizeObserver methods with jest.fn mocks

nazimmuhammed:
Replaced empty ResizeObserver mock methods with jest.fn().

This resolves SonarQube typescript:S1186 warnings and improves testability by allowing assertions on mock method calls.


arkid15r (approved):

arkid15r (commented):
jest setup changes are good while security.md file is not related to the issue. not sure why you added it here but I merged those after updates anyway.

keep your contributions narrow scoped, that's easier to review

PR #3981, opened 2026-02-17
Fix race condition in API key creation by locking user row

### LINKED ISSUE #3634: race conditions could result in more api keys created for a user than is allowed

ItsImran10140:
## Proposed change

Resolves #3634 

The `select_for_update()` was previously called on the existing API key rows, which didn't prevent concurrent transactions from reading the same count simultaneously. This could allow a user to create more API keys than the `MAX_ACTIVE_KEYS` limit.

This fix locks the **user row** instead using `select_for_update().get(pk=user.pk)` before checking the active key count. This ensures that any concurrent `create()` call for the same user will block until the first transaction completes, guaranteeing an accurate count check and preventing the race condition.

### Changes:
- **`backend/apps/api/models/api_key.py`**: Lock the user row with `get_user_model().objects.select_for_update().get(pk=user.pk)` and use the locked user object for the active key count check.
- **`backend/tests/apps/api/models/api_key_test.py`**: Updated `test_create_success` and `test_create_max_keys_exceeded` to properly mock the `get_user_model` call chain, configuring the `locked_user` mock with the expected `active_api_keys.count` return values.

## Checklist

### DIFF
--- a/backend/apps/api/models/api_key.py
+++ b/backend/apps/api/models/api_key.py
@@ -5,9 +5,10 @@ import secrets
 import uuid
 
 from django.conf import settings
-from django.contrib.auth import get_user_model
 from django.db import models, transaction
 from django.utils import timezone
+
+from apps.nest.models.user import User
 
 API_KEY_LENGTH = 32
 MAX_ACTIVE_KEYS = 3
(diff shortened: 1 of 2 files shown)


ItsImran10140:
@arkid15r Requested Review

PR #3985, opened 2026-02-18
Fix SSR Apollo cookie overwrite in server-side GraphQL requests

### LINKED ISSUE #3940: Bug: SSR Apollo client overwrites Cookie header and drops auth/session cookies
## Summary
SSR GraphQL requests can lose authenticated context because the server Apollo client overwrites the `Cookie` header with only `csrftoken`.

## Affected code
- `frontend/src/server/apolloClient.ts`

## Problem
In the server Apollo `setContext`, headers are set as:
- `X-CSRFToken: <token>`
- `Cookie: csrftoken=<token>`

This replaces the full incoming cookie header instead of preserving/merging it, so session/auth cookies are dropped.

## Why this is an issue
Any SSR GraphQL operation that depends on authenticated session cookies may behave as unauthenticated (or inconsistently), because the outgoing request no longer carries full cookie state.

## Reproduction
1. Sign in to Nest in browser (session/auth cookies present).
2. Open a route that performs server-side GraphQL via `server/apolloClient`.
3. Add temporary debug logging in `frontend/src/server/apolloClient.ts` to inspect outbound request headers.
4. Observe `Cookie` header contains only `csrftoken=...` and omits other cookies.

## Expected
Server-side GraphQL requests should preserve incoming cookies (including session/auth cookies) and include CSRF safely.

## Actual
`Cookie` is overwritten to only `csrftoken=...`.

## Proposed fix
- Preserve full incoming cookie header from request context.
- Merge or append CSRF cookie/token without discarding existing cookies.
- Keep `X-CSRFToken` behavior intact.

## Acceptance criteria
- SSR Apollo requests include original session/auth cookies.
- CSRF token handling remains functional.
- Add regression test for header propagation/merge behavior in server Apollo client.

xDipzz:
## Proposed change

Resolves #3940

This PR fixes SSR Apollo cookie propagation so authenticated context is preserved during server-side GraphQL requests.

### What changed
- Added `frontend/src/server/apolloCookieHeader.ts` with utilities to merge cookies safely and keep exactly one `csrftoken` value.
- Updated `frontend/src/server/apolloClient.ts` to build request cookies from `next/headers` (`cookies().getAll()`), then merge with server CSRF token via `mergeApolloHeadersWithCsrf(...)`.
- Added regression tests in `frontend/__tests__/unit/server/apolloCookieHeader.test.ts` for:
  - preserving existing session/auth cookies,
  - replacing conflicting `csrftoken` without duplicates,
  - case robustness for `Cookie`/`cookie`,
  - null-CSRF behavior,
  - preserving non-cookie headers.

### Verification
- Ran:
  - `pnpm --dir frontend exec jest __tests__/unit/server/apolloCookieHeader.test.ts --runInBand --coverage=false`
- Result:
  - `PASS` (6/6 tests)

## Checklist


Shofikul-Isl4m:
you have to address  sonarcloud issues 
Screenshot from 2026-02-21 13-07-43

xDipzz:
> you have to address sonarcloud issues Screenshot from 2026-02-21 13-07-43

Thanks ,SonarCloud shows Quality Gate passed with 1 new issue. I’ll review it and address it.

xDipzz:
@arkid15r  @kasya  Please review PR #3985 and let me know if any changes are required.

preeeetham:
Good fix — the root cause was correctly identified. The old code was setting `Cookie: csrftoken=...` which nuked all existing session cookies during SSR, and the new approach of reading all cookies from `next/headers` and merging them properly makes sense. The deduplication logic in `mergeCookieHeader` is clean and the case-insensitive header handling is a nice touch.

A couple things I'd flag though: `getRequestCookieHeaderOnServer` forwards *every* cookie the client sent to the backend via `cookieStore.getAll()`. That includes analytics cookies, third-party cookies, anything the browser attached. Ideally only the relevant auth cookies like `sessionid` and `csrftoken` should be forwarded — forwarding everything violates least-privilege and could hit header size limits if there are many cookies. A simple whitelist filter before joining would tighten this up.

On the cookie merging side, `mergeCookieHeader` concatenates values directly with `split(';')` and string joining without any encoding or sanitization. Since the values come from `cookies().getAll()` which returns already-parsed cookies this is probably fine in practice, but if a cookie value ever contained a semicolon it would break the parsing. Not a likely scenario but worth being aware of. Also, an empty-string CSRF token currently gets treated as a valid value and produces `csrftoken=` in the header — might want to treat that the same as `null` and skip setting CSRF headers entirely, since an empty token is almost certainly a bug upstream.

xDipzz:
> Good fix — the root cause was correctly identified. The old code was setting `Cookie: csrftoken=...` which nuked all existing session cookies during SSR, and the new approach of reading all cookies from `next/headers` and merging them properly makes sense. The deduplication logic in `mergeCookieHeader` is clean and the case-insensitive header handling is a nice touch.
> 
> A couple things I'd flag though: `getRequestCookieHeaderOnServer` forwards _every_ cookie the client sent to the backend via `cookieStore.getAll()`. That includes analytics cookies, third-party cookies, anything the browser attached. Ideally only the relevant auth cookies like `sessionid` and `csrftoken` should be forwarded — forwarding everything violates least-privilege and could hit header size limits if there are many cookies. A simple whitelist filter before joining would tighten this up.
> 
> On the cookie merging side, `mergeCookieHeader` concatenates values directly with `split(';')` and string joining without any encoding or sanitization. Since the values come from `cookies().getAll()` which returns already-parsed cookies this is probably fine in practice, but if a cookie value ever contained a semicolon it would break the parsing. Not a likely scenario but worth being aware of. Also, an empty-string CSRF token currently gets treated as a valid value and produces `csrftoken=` in the header — might want to treat that the same as `null` and skip setting CSRF headers entirely, since an empty token is almost certainly a bug upstream.

Thanks for the review. I agree on least-privilege and header-size concerns, but I didn’t
  want to guess a cookie allowlist and risk breaking SSR auth/session flows. If maintainers
  confirm the expected cookie set to forward, I can follow up with a small allowlist + tests;
  for now this PR keeps SSR behavior consistent by preserving the incoming request cookies.

arkid15r (changes requested):
Please resolve conflicts

xDipzz:
> Please resolve conflicts

@arkid15r  Conflicts have been resolved and the branch is up to date with main. Please let me know if any further changes are required.

PR #4020, opened 2026-02-20
feat: implement module ordering with drag-and-drop support

### LINKED ISSUE #3016: Add module ordering functionality for mentorship modules
Program admins need to control the order in which modules appear within each mentorship program. Currently, there is no way to set or persist module order, which causes confusion when viewing programs in both admin and public views.

- Enable Program admins to set the order of modules in a program from the frontend.
- Ordering should be possible by drag-and-drop or other intuitive UI element, visible only to Program admins.
- The specified order must be saved and persisted all the way to the backend.
- When a program is loaded in admin or public view, modules must appear in the set order.
- Restrict module ordering capability strictly to Program admins.
- Keep in mind different UI on the frontend for programs that have only 1 or multiple modules.

- [ ] Yes
- [x] No

**Additional context**
This improves usability for admins and makes programs more logically organized for public viewing.

HarshitVerma109:
## Proposed change
Resolves #3016 

This PR adds a feature that lets Program Admins reorder modules by simply dragging and dropping them

https://github.com/user-attachments/assets/6366e988-5a71-4752-828f-05c7b4e792f4

## Checklist

### DIFF
--- a/backend/apps/mentorship/api/internal/mutations/module.py
+++ b/backend/apps/mentorship/api/internal/mutations/module.py
@@ -504,7 +504,7 @@ class ModuleMutation:
             msg = "Provided module keys do not match the program's modules."
             raise ValidationError(msg)
 
-        modules = list(modules_query.select_for_update())
+        modules = modules_query.select_for_update()
 
         key_to_order = {key: idx for idx, key in enumerate(input_data.module_keys)}
 


HarshitVerma109:
@kasya , please review it and let me know if any changes are required.

backend/apps/mentorship/models/module.py
```diff
@@ -47,6 +48,11 @@ class Meta:
         blank=True,
         default="",
     )
+    order = models.PositiveIntegerField(
+        default=0,
```
kasya, line 52: This works great with reordering existing modules 👍🏼 
However, when I create new modules they seem to be returned in random order.
Example (steps I took):

Image

- had second and third modules;
- ordered third module to be first, then second;
- Created 4th module - it was placed between third and second (second position)
- Created 5th module - it was placed on first position.
- I then ordered them again (2dn, 3d, 4th, etc)
-  created 7th module. It was placed on the second position again.

Image

I think it would make sense if newly created modules were placed either at the beginning or at the end. But not at random 🤔 And with us using positions to order these on the frontend now I feel like adding newly created modules to the end would make the most sense to not interrupt current ordering (if any)

kasya (changes requested):
@HarshitVerma109 thanks for working on this so quickly!

Ordering works good 👍🏼 Noticed some issues with creating new modules when we already have some that are ordered ⬇️

backend/apps/mentorship/models/module.py
HarshitVerma109: Yeah, that totally makes sense! Sorry about that—I didn't add the logic to handle where newly created modules get placed initially. I've fixed it now, so new modules will automatically be added to the end of the list. 👍🏼

HarshitVerma109:
@kasya , please review it and let me know if any changes are required.

kasya (dismissed):
Works great now! 
Thanks @HarshitVerma109 👍🏼

backend/apps/mentorship/api/internal/mutations/module.py
```diff
@@ -391,3 +392,48 @@ def update_module(self, info: strawberry.Info, input_data: UpdateModuleInput) ->
...
+            raise ObjectDoesNotExist(msg) from e
+
+        if not program.admins.filter(nest_user=user).exists():
+            raise PermissionDenied
+
+        if len(set(input_data.module_keys)) != len(input_data.module_keys):
+            msg = "Duplicate module keys are not allowed."
+            raise ValidationError(msg)
+
+        modules = list(
```
arkid15r: I'm not sure about list() and select_for_update() parts.

backend/apps/mentorship/api/internal/mutations/module.py
```diff
@@ -391,3 +392,48 @@ def update_module(self, info: strawberry.Info, input_data: UpdateModuleInput) ->
...
+            Module.objects.filter(
+                program=program, key__in=input_data.module_keys
+            ).select_for_update()
+        )
+
+        if len(modules) != len(input_data.module_keys):
```
arkid15r: .count() of the queryset.

backend/apps/mentorship/models/module.py
```diff
@@ -47,6 +48,11 @@ class Meta:
         blank=True,
         default="",
     )
+    order = models.PositiveIntegerField(
```
arkid15r: Would [small integer](https://docs.djangoproject.com/en/6.0/ref/models/fields/#positivesmallintegerfield) be enough here?

frontend/__tests__/unit/components/ModuleCard.test.tsx
```diff
@@ -9,6 +9,54 @@ import { ExperienceLevelEnum } from 'types/__generated__/graphql'
...
+  DndContext: ({ children }: { children: React.ReactNode }) => (
+    <div data-testid="dnd-context">{children}</div>
+  ),
+  closestCenter: jest.fn(),
+  useSensor: jest.fn(() => ({})),
+  useSensors: jest.fn(() => []),
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  KeyboardSensor: jest.fn(),
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  PointerSensor: jest.fn(),
```
arkid15r, line 33: Why do we need these? DnD?

arkid15r (changes requested):

frontend/__tests__/unit/components/ModuleCard.test.tsx
HarshitVerma109: Yes, exactly! They are for @dnd-kit. Our tests run in JSDOM, which doesn't support the complex pointer and keyboard sensors that @dnd-kit uses. If we don't mock them to return empty, the library throws errors and crashes the component render during the tests.

backend/apps/mentorship/models/module.py
HarshitVerma109: Yes, you are completely right! It will optimize the database storage by saving 2 bytes per row.

backend/apps/mentorship/api/internal/mutations/module.py
HarshitVerma109: I used select_for_update() to lock the rows during the transaction, preventing race conditions if two admins try to reorder simultaneously. For list(), I will update this to use .count() on the queryset as you suggested.

HarshitVerma109:

anurag2787:
Hi @HarshitVerma109 Could you please verify this?
When I reorder the modules and navigate back, then return to the same page, the modules do not appear in the reordered sequence. However, after refreshing the page, they display in the correct order that I set.
I am attaching a screen recording for reference.

[Screencast from 2026-02-25 20-41-24.webm](https://github.com/user-attachments/assets/b9249963-3a64-43ce-bf63-c81bae7937d1)

anurag2787 (changes requested):
Hi @HarshitVerma109 the ordering feature should be available only to Program Admins. Mentors shouldn’t be able to see or access the ordering option at all.

[Screencast from 2026-02-26 13-11-59.webm](https://github.com/user-attachments/assets/ad434ad3-2747-4554-a491-e9d994c64711)

preeeetham:
Solid full-stack implementation here — the backend concurrency handling with `select_for_update` inside `@transaction.atomic` is well done, and the frontend optimistic update with rollback on failure is great UX. A few things worth addressing though:

The main concern is a race condition in `Module.save()`. The auto-order logic does a `Max("order")` aggregate query followed by `self.order = (max_order or 0) + 1`, but this read-then-write isn't atomic. Two modules created at the same time could get the same order value. Unlike the reorder mutation, `save()` doesn't use `select_for_update`. Wrapping it in `transaction.atomic` or adding a DB-level unique constraint on `(program, order)` would prevent that.

The reorder mutation validation also has a gap — it checks that all *provided* keys exist in the program, but doesn't verify that *all* program modules are included in the input. Someone could submit a partial list of keys, reordering only some modules while leaving others with potentially conflicting order values. Comparing the count against `Module.objects.filter(program=program).count()` would close that.

One smaller thing on the frontend side — the `SortableContext` items use `m.key || m.id` as identifiers, but the mutation payload sends `newOrder.map(m => m.key)` without the fallback. If a module ever has an empty key, the sortable ID would be the `id` but the mutation would send an empty string. Using the same `key || id` logic in both places would be safer.

HarshitVerma109:
@arkid15r, I’ve resolved the merge conflicts. Please review the changes and let me know if any further adjustments are needed.

backend/apps/mentorship/api/internal/mutations/module.py
```diff
@@ -475,3 +476,46 @@ def delete_module(
...
+            msg = "Duplicate module keys are not allowed."
+            raise ValidationError(msg)
+
+        modules_query = Module.objects.filter(program=program, key__in=input_data.module_keys)
+
+        if modules_query.count() != len(input_data.module_keys):
+            msg = "Provided module keys do not match the program's modules."
+            raise ValidationError(msg)
+
+        modules = list(modules_query.select_for_update())
```
arkid15r: I don't see list() removed. Do you need it as a list here?

HarshitVerma109:
@arkid15r, I’ve resolved the merge conflicts. Please review the changes and let me know if any further adjustments are needed.

backend/apps/mentorship/migrations/0008_module_order.py
kasya, line 1: Just reminding you for next time - when you have a migration in your PR, you need to update `compose.yaml` file to use a new db volume and push it with the PR. 
Otherwise when we test locally we might have db conflicts.

PR #4104, opened 2026-02-27
This PR fixes an issue where fetchCsrfTokenServer.ts could return cached CSRF tokens

### LINKED ISSUE #4082: Server CSRF token fetch is cacheable in Next (stale token risk)
## Describe the bug

fetchCsrfTokenServer.ts uses fetch() without explicitly disabling caching. In Next.js App Router, server-side fetch() calls are cached by default unless configured otherwise.
This may cause the CSRF token response to be cached and reused across multiple requests, leading to intermittent CSRF validation failures due to stale or reused tokens.

## To Reproduce

Trigger a server-side request that calls fetchCsrfTokenServer.ts
Perform multiple authenticated form submissions in sequence
Observe intermittent CSRF validation failures
Inspect network/server logs to notice reused CSRF token responses

## Expected behavior
Each request to fetch a CSRF token should return a fresh, non-cached token. The server should not reuse previously fetched CSRF tokens across requests.

Are you going to work on fixing this?
 Yes

Image

nios-x:
## Proposed change

Resolves #4082

## Description
This PR fixes an issue where fetchCsrfTokenServer.ts could return cached CSRF tokens due to default caching behavior in Next.js App Router.
In Next.js, server-side fetch() calls are cached by default unless explicitly configured otherwise. Since caching was not disabled, the CSRF token response could be reused across multiple requests, leading to intermittent CSRF validation failures caused by stale tokens.

## Checklist

### DIFF
--- a/frontend/src/server/fetchCsrfTokenServer.ts
+++ b/frontend/src/server/fetchCsrfTokenServer.ts
@@ -3,6 +3,7 @@ export const fetchCsrfTokenServer = async (): Promise<string> => {
     throw new Error('NEXT_SERVER_CSRF_URL is not configured')
   }
   const response = await fetch(process.env.NEXT_SERVER_CSRF_URL, {
+    cache: 'no-store',
     credentials: 'include',
     method: 'GET',
   })


nios-x:
@arkid15r

PR #4111, opened 2026-02-28
fix: improve OpenAI error handling with specific exception types

### LINKED ISSUE #4024: Bug: OpenAI API Exception Handling Too Broad - Masks Auth, Rate Limit, and Model Errors
Hi @arkid15r @kasya!

I'd like to work on this issue. I've analyzed the codebase and found the exact exception handling problem at lines 103-106 of `backend/apps/common/open_ai.py`.

I've prepared a complete fix that follows the same error handling pattern from Issue #3987 (Algolia 503 configuration guard) for consistency across the codebase.

The implementation includes:
- Specific exception handlers for: AuthenticationError, RateLimitError, BadRequestError, APIError
- Descriptive log messages for each error type
- 5 comprehensive test cases
- Maintains backward compatibility

Expected timeline: **2-3 days** (1 day implementation + 1 day review cycles)

Can I be assigned to this issue?

Shashwat-Darshan:
## Proposed change

Resolves #4024

- Replace broad `except Exception` with specific handlers for:
  - `AuthenticationError` (invalid API key)
  - `RateLimitError` (HTTP 429)
  - `BadRequestError` (invalid parameters)
  - `APIConnectionError` (network issues)
  - Generic exceptions with error type in logs
- Add 5 new test cases covering each exception type
- Maintains backward compatibility (returns `None` as before)
- Improves debuggability and observability in production

## Checklist

---


preeeetham:
Clean improvement — breaking down the generic `except Exception` into specific OpenAI error types makes debugging way easier. Couple of things worth mentioning:

The exception handler ordering matters here. `AuthenticationError`, `RateLimitError`, and `BadRequestError` all inherit from `openai.APIStatusError`, which itself inherits from `openai.APIError`. The current ordering is fine since they're all separate leaf exceptions, but if someone later adds a catch for `APIStatusError` or `APIError` above the specific ones, it would shadow them. Adding a brief comment like `# Most specific exceptions first` would make the intent clear for future contributors.

The `RateLimitError` handler uses `logger.warning()` while all others use `logger.exception()`. The docstring-level `Raises` section was removed, which is correct since they're now caught internally, but `warning` won't include the traceback. For rate limits that's probably fine since it's expected behavior, but it might be worth logging the `retry-after` header value if it's available in the exception — that would make it easier to tune retry logic later.

The catch-all `except Exception as e` now logs `type(e).__name__`, which is useful, but the format string uses `%s` style (lazy logging) — that's consistent with the project's style so that's good. One minor thing though: since `logger.exception` already includes the traceback, the type name in the message is somewhat redundant with the exception info that gets logged. Not a blocker, just worth noting.

PR #4137, opened 2026-03-01
fix:zap baseline security headers

### LINKED ISSUE #4090: Fix staging post-migration ZAP findings

---

After zappa/terraform migration ZAP baseline scan fails with https://github.com/OWASP/Nest/actions/runs/22426114656/job/64936382717#step:3:136

Fix it, don't add to ignore list -- migrate what's needed into new infrastructure configuration

- [ ] Yes
- [x] No

HarshitVerma109:
## Proposed change
Resolves #4090 
Migrate security headers from Nginx proxy into Next.js config and ALB listener to fix ZAP baseline scan failures on staging.

## Checklist


HarshitVerma109:
@arkid15r, please review it and let me know if any changes are required.

preeeetham:
Strong security improvement — adding CSP, HSTS, X-Frame-Options, and the other hardening headers at both the Next.js and ALB layers is good defense-in-depth. A few things I'd flag:

The CSP policy is defined in two places — `frontend/next.config.ts` as a structured array and `infrastructure/modules/alb/main.tf` as a single string. These need to stay in sync manually, and they're in completely different formats (JS array vs Terraform string). If someone updates the CSP in the Next.js config but forgets the ALB, the headers will conflict depending on which one takes precedence. Consider extracting the CSP to a shared config file or at minimum adding a comment in each location pointing to the other one.

The CSP includes `'unsafe-inline'` for both `script-src` and `style-src` in production. For styles, `'unsafe-inline'` is sometimes unavoidable with CSS-in-JS frameworks, but for scripts it significantly weakens the CSP since it allows inline script injection — which is exactly what CSP is supposed to prevent. The `isLocal` check only gates `'unsafe-eval'`, not `'unsafe-inline'`. If the app doesn't rely on inline scripts in production, this should be removed. If it does (e.g., Next.js injects inline scripts for hydration), then adding a `nonce`-based approach would be more secure.

The `Cross-Origin-Embedder-Policy: credentialless` is a relatively new header. It's less restrictive than `require-corp` but still might break third-party resources that don't support it — worth verifying that the OpenStreetMap tiles, PostHog scripts, and Google Analytics all work correctly with this header enabled.

On the Terraform side, the ALB listener now sets headers directly on the listener resource. This is clean, but note that these headers will apply to all responses including error pages and redirects. The `X-Frame-Options: DENY` is good but it's redundant with `frame-ancestors 'none'` in the CSP — browsers that support CSP will use `frame-ancestors` and ignore `X-Frame-Options`. Not a problem, just redundant for modern browsers (but good to keep for older ones).

The test coverage for the ALB headers is solid — assertions for each header value are a nice touch.

Issue #4247, opened 2026-03-12
Fix issue timestamp on /contribute page


arkid15r:
Community & Support: [LinkedIn Group](https://www.linkedin.com/groups/14656108/) · [Slack #project-nest](https://owasp.slack.com/archives/project-nest)

Active project leaders: Arkadii Yakovets -- [GitHub](https://github.com/arkid15r/) · [LinkedIn](https://www.linkedin.com/in/arkid15r/) · [Slack](https://owasp.slack.com/team/U060W3NKLTF); Kate Golovanova -- [GitHub](https://github.com/kasya/) · [LinkedIn](https://www.linkedin.com/in/kate-golovanova/) · [Slack](https://owasp.slack.com/team/U07PWB1JZ6Z)

[Contributing](https://github.com/owasp/nest/blob/main/CONTRIBUTING.md) · [Code of Conduct](https://github.com/owasp/nest/blob/main/CODE_OF_CONDUCT.md) · [GSoC Mentors](https://github.com/owasp/nest/blob/main/MENTORS.md)

---

**Describe the bug**
Timestamp is not set correctly.

**To Reproduce**

See https://nest.owasp.org/contribute

**Expected behavior**
The correct timestamp is shown.

**Are you going to work on fixing this?**

- [ ] Yes
- [x] No

**Screenshots**

<img width="1250" height="761" alt="Image" src="https://github.com/user-attachments/assets/a45aa683-79bc-4e51-862d-c5607a2a7be3" />

mrkeshav-05:
Hi @arkid15r , Can I work on this?

saichethana28:
Hi @arkid15r can I work on this?

aryanghai12:
@arkid15r can i work on this one?

HarshitVerma109:
Hi @arkid15r,
Could you please assign this issue to me ?

Utkarsh-0304:
Hi @arkid15r, could you please assign this issue to me?

anurag2787:
Hi @arkid15r Could you please assign this to me also ?

anirudhprmar:
would like to work on it

moizulfiqar:
Hey! Id like to work on this

ItsImran10140:
Hi @arkid15r Can I also work on this one please.

PR #4260, opened 2026-03-13
Fix: Timestamp display on /contribute page

mrkeshav-05:
### Proposed change
This PR fixes incorrect timestamp display on the `/contribute` page where issues appeared as "56 years ago". The backend was returning Unix timestamps in seconds, while the frontend interpreted them as milliseconds, causing incorrect dates. Updated the backend to return ISO 8601 timestamps and adjusted the frontend utility to correctly handle both numeric and string timestamps.

Resolves: #4247

## Checklist

## Release Notes

* **Bug Fixes**
  * Standardized timestamp formatting to ensure consistent, human-readable date display across the application.
  * Enhanced date parsing logic to reliably support various date format inputs.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,7 +55,7 @@ repos:
           - --args=--config=__GIT_WORKING_DIR__/infrastructure/.tflint.hcl
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.5
+    rev: v0.15.6
     hooks:
       - id: ruff-check
         args:
--- a/backend/poetry.lock
+++ b/backend/poetry.lock
@@ -275,18 +275,18 @@ files = [
 
 [[package]]
 name = "boto3"
-version = "1.42.66"
+version = "1.42.67"
 description = "The AWS SDK for Python"
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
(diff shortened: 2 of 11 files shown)


frontend/src/utils/utility.ts
```diff
@@ -24,7 +24,9 @@ export const getFilteredIcons = (project: ProjectType, params: string[]): Icon =
   const filteredIcons = params.reduce((acc: Icon, key) => {
     if (ICONS[key as IconKeys] && project[key as keyof typeof project] !== undefined) {
       if (key === 'createdAt') {
-        acc[key] = dayjs(project[key as keyof ProjectType] as unknown as string).fromNow()
+        const timestamp = project[key as keyof ProjectType]
+        const dayjsDate = typeof timestamp === 'number' ? dayjs.unix(timestamp) : dayjs(timestamp)
```
arkid15r, line 28: What are the cases when createdAt is still a number?

mrkeshav-05: `createdAt` was previously stored in the Algolia index as a Unix timestamp (number). I updated the backend to serialize it as an ISO date string, but existing indexed records may still contain the old numeric format until a full re-index happens.
This check ensures the frontend safely handles both formats and prevents incorrect time rendering for older records.

arkid15r: We re-index data upon each deploy. Please clean up the unused code when you have a chance.

mrkeshav-05: Thanks for the clarification! 
 I’ll remove the legacy number handling. Should I create a separate issue for the cleanup?

arkid15r: Yes, otherwise your PR will be closed automatically.

PR #4418, opened 2026-03-28
Fix Aria Select label on SortBy trigger

kasya:
HeroUI `SortBy` rendered the chapters (and other) sort Select without a visible label or aria-label, which triggered React Aria’s useLabel dev warnings on each render - especially when sorting. Added aria-label when `showLabel` is false and set `data-scroll-behavior="smooth"` on <html> so it matches `globals.css` and satisfies Next.js. 
Updated SortBy tests to target the select trigger reliably.

This would add a bunch of these warnings on any interaction with the `SortBy` component, completely flooding the console.

Screenshot 2026-03-28 at 1 54 07 PM

## Checklist


arkid15r (approved):
LGTM

Issue #3960, opened 2026-02-15
CI/CD: Release version calculated but not used for Docker image tags


hassaansaleem28:
**Describe the bug**

The CI/CD pipeline correctly calculates a unique RELEASE_VERSION, but it doesn't use this version when tagging the Docker images. Currently, it looks like owasp/nest:backend-production and owasp/nest:frontend-production 
This means that while the application code knows its version, the Docker Registry does not have a corresponding versioned tag.

**To Reproduce**

- Open [.github/workflows/run-ci-cd.yaml](https://github.com/OWASP/Nest/blob/main/.github/workflows/run-ci-cd.yaml#L381)
- Observe the set-release-version job, which outputs a release_version
- Scroll down to the build-production-images job.
- No version tags with imgs.

**Expected behavior**

 The Docker images should be tagged with both:
- The mutable pointer: owasp/nest:backend-production
- The immutable version: owasp/nest:backend-VERSION

This ensures that we can rollback to a specific previous version if a deployment fails.

**Are you going to work on fixing this?**

- [x] Yes
- [ ] No

hassaansaleem28:
Hey @arkid15r,
Could you please approve this and assign me?

preeeetham:
Hi @arkid15r @kasya,

**Approach:** Update `build-production-images` job in `.github/workflows/run-ci-cd.yaml` to use `release_version` from `set-release-version` when tagging images (e.g. `owasp/nest:backend-${RELEASE_VERSION}`) alongside the mutable `backend-production` tag.

Please assign this issue to me.

david-wagih:
@arkid15r 
hi Ark can you assign this to me ?

arkid15r:
I suggest waiting for https://github.com/OWASP/Nest/pull/4292 merge

preeeetham:
> I suggest waiting for [#4292](https://github.com/OWASP/Nest/pull/4292) merge

Sure!!

hassaansaleem28:
Hi @arkid15r, #4292 is merged
Could you plz assign this issue to me?

arkid15r:
It seems it's no longer relevant -- see https://github.com/OWASP/Nest/pull/4463#pullrequestreview-4042947088

PR #4463, opened 2026-04-01
Fix Deployment Race Condition

### LINKED ISSUE #4462: Fix Deployment Race Condition

---

Fix deployment timeout due to indefinite deploy.
Suggested fix is to increase the `unhealthy_threshold` to prevent a race condition.

- [x] Yes
- [ ] No

**Additional context**

rudransh-shrivastava:
## Proposed change
Resolves #4462
Fix deployment race condition by increasing unhealthy threshold. Previously, only a 3*30 second window was available for both tasks to be ready, this caused a race condition. If the tasks failed to stabilize under that time, the health checker would stop those tasks. Increasing the threshold to 10 allows a 300 second window for a task to be ready.

Successful CI/CD: https://github.com/rudransh-shrivastava/Nest/actions/runs/23840249166/job/69494432112

Some steps in CI/CD were removed (reasoning in comments)
## Checklist

### DIFF
--- a/.github/workflows/run-ci-cd.yaml
+++ b/.github/workflows/run-ci-cd.yaml
@@ -490,9 +490,19 @@ jobs:
   set-release-version:
     name: Set release version
     if: |
-      github.event_name == 'schedule' ||
-      github.event_name == 'workflow_dispatch' ||
-      (github.event_name == 'release' && github.event.action == 'published')
+      github.repository == 'OWASP/Nest' &&
+      (
+        github.event_name == 'schedule' ||
+        github.event_name == 'workflow_dispatch' ||
+        (github.event_name == 'release' && github.event.action == 'published')
+      )
+    needs:
+      - run-backend-tests
+      - run-frontend-a11y-tests
+      - run-frontend-e2e-tests
+      - run-frontend-unit-tests
+      - run-infrastructure-tests
+      - run-rest-fuzz-tests
     outputs:
       release_version: ${{ steps.set.outputs.release_version }}
     permissions: {}
@@ -854,7 +864,6 @@ jobs:
           frontend_image_tag="$RELEASE_VERSION"
           project_name="$PROJECT_NAME"
           tasks_use_fargate_spot=$TASKS_USE_FARGATE_SPOT
-          db_backup_retention_period=0
           EOF
 
       - name: Terraform Init
@@ -1074,6 +1083,7 @@ jobs:
       RELEASE_VERSION: ${{ needs.set-release-version.outputs.release_version }}
     environment: production
     if: |
+      github.repository == 'OWASP/Nest' &&
       github.event_name == 'release' &&
       github.event.action == 'published'
     needs:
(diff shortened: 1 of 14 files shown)


.github/workflows/run-ci-cd.yaml
```diff
@@ -1561,17 +1531,7 @@ jobs:
...
-          aws ecs update-service \
-            --cluster "$CLUSTER_NAME" \
-            --service "$SERVICE_NAME" \
-            --force-new-deployment
-
-      - name: Wait for frontend service stability
+      - name: Wait for backend service stability
         env:
           CLUSTER_NAME: ${{ steps.tf-outputs.outputs.frontend_cluster_name }}
           SERVICE_NAME: ${{ steps.tf-outputs.outputs.frontend_service_name }}
```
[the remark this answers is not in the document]
rudransh-shrivastava: fixed. can you check again @cubic-dev-ai

infrastructure/modules/alb/main.tf
```diff
@@ -165,7 +165,7 @@ resource "aws_lb_target_group" "backend" {
     path                = var.backend_health_check_path
     protocol            = "HTTP"
     timeout             = 5
-    unhealthy_threshold = 3
```
rudransh-shrivastava, line 168: This fix ensures successful deployments but tasks health will remain `unknown` as task health check is not enabled by us.

It is fine for now (deployments will succeed) but we should add an explicit health check in the task definition later.

.github/workflows/run-ci-cd.yaml
```diff
@@ -978,26 +978,6 @@ jobs:
 
           echo "Index-data task started successfully (runs async, ~30 min)"
 
-      - name: Force new backend deployment
```
rudransh-shrivastava, line 981: Not required, adding a new task revision will automatically start a new deployment. This was required when we had `MUTABLE` image tags because the task definition ARN didn't change (we used `latest` tag on all images), so ECS had no reason to redeploy.

Now that we have `IMMUTABLE` tags, every new deployment creates a new task definition (because `image_tag` changes). ECS picks this up and triggers a new deployment.

infrastructure/modules/alb/main.tf
arkid15r: > It is fine for now (deployments will succeed) but we should add an explicit health check in the task definition later.


Sounds like enough for a new issue.

.github/workflows/run-ci-cd.yaml
arkid15r: Could you check whether https://github.com/OWASP/Nest/issues/3960 is still relevant?

arkid15r (approved):
Thanks for the fix 👍

.github/workflows/run-ci-cd.yaml
rudransh-shrivastava: Since we use ECR + ECS for deployments, I don't think images in Docker Hub do anything. The only use-case I found was cache to/from:

https://github.com/OWASP/Nest/blob/4a9bbd0c247692f5c5c627dc0be24c02c93816df/.github/workflows/run-ci-cd.yaml#L571-L575

Do we want to keep uploading to Docker Hub? Please correct me if i'm missing any use-cases for the docker hub repositoes.

Edit: Another point I would like to add, our Docker Hub repositories are public, but our ECR repositories are private.

arkid15r: I don't think we need to maintain the dockerhub images. The CI/CD is going to be migrated to docker-less approach later so we won't be relying on caching that much. The only downside is that AWS solution is a bit vendor lock but it's not the only part that's locked on AWS -- so I think we're fine to migrate to ECR completely.

rudransh-shrivastava: > The CI/CD is going to be migrated to docker-less approach later so we won't be relying on caching that much.

Hmm, interesting...

Alright, I think we may close that issue then.

PR #4829, opened 2026-06-04
Prevent NestBot auto reply from re-processing AI responses

### LINKED ISSUE #4828: NestBot auto reply re-processes AI responses before posting
## Description

While investigating the NestBot auto reply flow, I noticed that the generated AI response is being passed to `get_blocks()` before posting to Slack.

`get_blocks()` is intended to process a user query and internally calls `process_ai_query()`. In the auto reply flow, the AI response has already been generated by:

process_ai_query(query=message.text)

Passing the generated response into `get_blocks()` causes the response to be processed again.

## Current Behavior

1. User message is received.
2. `process_ai_query()` generates an AI response.
3. The AI response is passed to `get_blocks()`.
4. `get_blocks()` invokes `process_ai_query()` again.

## Expected Behavior

The generated AI response should be converted directly into Slack blocks without re-running AI processing.

## Proposed Fix

Use the already generated AI response to construct Slack blocks directly before posting the message, avoiding a second invocation of AI processing.

Tanishq-mellu:
## Summary

Fixes an issue in the NestBot auto reply flow where the generated AI response was being passed into `get_blocks()`, causing the response to be processed a second time.

Instead, the already generated AI response is now converted directly into a Slack block before posting.

Fixes #4828

## Changes

- Remove `get_blocks()` usage from the auto reply flow.
- Generate Slack blocks directly from the existing AI response.
- Update unit tests to reflect the new behavior.

## Testing

```bash
poetry run pytest tests/unit/apps/slack/services/message_auto_reply_test.py --no-cov
```

Result:

```text
8 passed
```


Tanishq-mellu:
While investigating https://github.com/OWASP/Nest/issues/4828, I noticed that OWASP question classification appears to happen in both:
* MessagePosted.handle_event()
* process_ai_query()
This may result in duplicate retrieval/OpenAI classification work before a response is generated. Is this intentional because process_ai_query() is a shared entry point or would maintainers consider consolidating the classification step ?

PR #5029, opened 2026-06-23
feat(o11y): Instrument backend with Otel metrics

### LINKED ISSUE #4975: Instrument backend with Otel metrics

---

Instrument backend with Otel and export metrics over OTLP/HTTP
to VM.

- Add Otel SDK + instrumentation pkgs
- Configure a meter provider
- process runtime metrics
- Testing

- [x] Yes
- [ ] No

hassaansaleem28:
## Proposed change

Resolves #4975 

- Adds Otel instrumentation to the backend, exporting request metrics (http.server.duration,
http.server.active_requests) over OTLP/HTTP to o11y-metrics.
- Uses manual instrumentation from AppConfig.ready() --  the otel-instrument auto-wrapper is incompatible with django-configurations (it loads settings too early and crashes on DEBUG/ALLOWED_HOSTS).
- Gated behind DJANGO_OTEL_METRICS_ENABLED (off by default).
- HTTP-only OTLP exporter -- metric setup is wrapped in error handling and an idempotency guard so it never blocks app startup.
- Verified locally: metrics appear in o11y-metrics.

## Checklist


docker-compose/local/compose.yaml
hassaansaleem28, line 1: a real issue: o11y-metrics is gated behind --profile o11y, but the backend service is always-on with DJANGO_OTEL_METRICS_ENABLED=true. So plain make run (no profile) instruments the backend and tries to export to o11y-metrics, which isn't running so OTel logs export warnings every ~15s. The metrics DB is opt-in, but the backend instrumentation isn't -- they need to switch on/off together.

Compose can't conditionally set a service's env from a profile in a single file, so i see two clean ways to fix it:

A) a separate compose.o11y.yaml 
B) contributors set DJANGO_OTEL_METRICS_ENABLED=true in backend/.env + run make run-o11y. 

cc @arkid15r

arkid15r: Yes, we normally use .env for configuration settings -- you don't need a separate `environment:` option in the compose file.

The separate compose file looks even better from separation of concerns perspective. How would that work?

hassaansaleem28: In that case 
make run-o11y will run compose.yml also compose.o11y.yaml with that env set in it
while make run will only compose.yml

arkid15r: I though about keeping it in a single file, update .env example w/ the default `o11y` values off (disabled) so folks who need this locally would just enable it manually. But we can start with separate files approach. We just need to make sure the new compose file is tracked by Dependabot.

docker-compose/local/compose.o11y.yaml
arkid15r, line 1: Will dependabot track this file?

backend/apps/common/apps.py
```diff
@@ -1,9 +1,24 @@
...
+
+logger = logging.getLogger(__name__)
 
 
 class CommonConfig(AppConfig):
     """Common app config."""
 
     name = "apps.common"
+
+    def ready(self):
```
arkid15r: Let's create a new `o11y` app?

docker-compose/local/compose.o11y.yaml
hassaansaleem28: Yes it seems so. [Dependabot regex](https://github.com/dependabot/dependabot-core/blob/main/docker/lib/dependabot/docker_compose/file_fetcher.rb#L9) allows to have a suffix before .yaml so it should track compose.o11y.yaml as it already tracks compose.yaml.

backend/tests/unit/apps/observability/apps_test.py
```diff
@@ -0,0 +1,29 @@
+from apps.observability.apps import ObservabilityConfig
+
+
+def test_ready_configures_otel_when_enabled(mocker, settings):
```
arkid15r, line 4: Please group test under classes.

arkid15r (approved):
LGTM

PR #5054, opened 2026-06-26
Remove unused langsmith dependency

Adarshkumar0509:
## DESCRIPTION
- langchain itself is already gone from main (must Have been cleaned up in an earlier PR). What's actually still was around unused is langsmith not imported anywhere in backend/apps. so i Removed it. 

- langgraph and langchain core stay exactly where they are (1.2.6 / 1.4.8) since langgraph is actively used in backend/apps/ai/agent/agent.py, and langchain-core is a required dep of langgraph.Ran osv-scanner against the updated lockfile so no issue is there .

## Checklist

- [X] **Required:** I followed the [contributing workflow](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md#contributing-workflow)
- [X] **Required:** I verified that my code works as intended and resolves the issue as described
- [X] **Required:** I ran all required checks and tests locally; all warnings addressed and failures resolved


arkid15r (approved):
Let's try 👍

Issue #5080, opened 2026-06-29
Improve backend type annotations and re-enable stricter type checking


arkid15r:
Community & Support: [LinkedIn Group](https://www.linkedin.com/groups/14656108/) · [Slack #project-nest](https://owasp.slack.com/archives/project-nest)

Active project leaders: Arkadii Yakovets -- [GitHub](https://github.com/arkid15r/) · [LinkedIn](https://www.linkedin.com/in/arkid15r/) · [Slack](https://owasp.slack.com/team/U060W3NKLTF); Kate Golovanova -- [GitHub](https://github.com/kasya/) · [LinkedIn](https://www.linkedin.com/in/kate-golovanova/) · [Slack](https://owasp.slack.com/team/U07PWB1JZ6Z)

[Contributing](https://github.com/owasp/nest/blob/main/CONTRIBUTING.md) · [Code of Conduct](https://github.com/owasp/nest/blob/main/CODE_OF_CONDUCT.md) · [GSoC Mentors](https://github.com/owasp/nest/blob/main/MENTORS.md)

---

**Describe the solution you'd like**

Continue improving type annotations across the OWASP Nest backend to increase code quality, improve IDE support, and catch potential issues earlier during development.

This follows up on the previous effort tracked in https://github.com/OWASP/Nest/issues/1258, where we started expanding backend type hints and improving static analysis coverage. Some related warnings have been suppressed in [pyproject.toml](https://github.com/OWASP/Nest/blob/eae67ceea20f6f3e027822c2162679b28ba1931f/backend/pyproject.toml#L81) to reduce noise while the migration is still in progress. The goal of this issue is to gradually eliminate those suppressions by annotating the remaining code and addressing the underlying type-checking issues.

Tasks
-   Add missing type annotations across the backend.
-   Improve generic type usage where appropriate.
-   Reduce or eliminate unnecessary Any usage.
-   Fix type-checking issues currently hidden by configuration.
-   Remove related warning suppressions from pyproject.toml once they are no longer needed.
-   Ensure CI passes with the stricter configuration.

Acceptance Criteria
- Backend modules have improved type coverage.
- Related suppressions are removed (or significantly reduced) from pyproject.toml.
- Static type checking completes successfully without introducing new warnings.
- Existing functionality and tests continue to pass.

**Are you going to work on implementing this?**
- [ ] Yes
- [x] No

**Additional context**
 - https://github.com/OWASP/Nest/pull/1290
[...]

PR #5100, opened 2026-06-30
Draft: Improve typing for github_user attributes on GraphQL nodes

Tanishq-mellu:
Refs #5080

## Summary

This draft PR improves type information for several GraphQL node classes by making the `github_user` attribute visible to mypy.

Several resolver methods access `github_user`, but the attribute is not declared on the corresponding Strawberry types, which results in `attr-defined` errors during static type checking.

## Changes

- Add `TYPE_CHECKING` imports for `GitHubUser`
- Add `github_user: GitHubUser | None` annotations to:
  - `AdminNode`
  - `MentorNode`
  - `AuthUserNode`

## Validation

- Ran `pre-commit run mypy --all-files`
- No mypy errors were reported

## Feedback requested

I'd appreciate feedback on whether this is the preferred pattern for exposing internal model attributes on Strawberry GraphQL node types while improving static typing.

## Question for reviewers

Is explicitly annotating model-backed attributes on Strawberry node types the preferred approach for resolving these `attr-defined` mypy errors, or would you recommend a different typing pattern?

### DIFF
--- a/backend/src/apps/mentorship/api/internal/nodes/admin.py
+++ b/backend/src/apps/mentorship/api/internal/nodes/admin.py
@@ -1,13 +1,18 @@
 """GraphQL node for Admin model."""
 
+from typing import TYPE_CHECKING
+
 import strawberry
 
+if TYPE_CHECKING:
+    from apps.github.models.user import User as GitHubUser
 
 @strawberry.type
 class AdminNode:
     """A GraphQL node representing a mentorship admin."""
 
     id: strawberry.ID
+    github_user: GitHubUser | None
 
     @strawberry.field(name="avatarUrl")
     def avatar_url(self) -> str:
--- a/backend/src/apps/mentorship/api/internal/nodes/mentor.py
+++ b/backend/src/apps/mentorship/api/internal/nodes/mentor.py
@@ -1,14 +1,18 @@
 """GraphQL node for Mentor model."""
 
+from typing import TYPE_CHECKING
+
 import strawberry
 
+if TYPE_CHECKING:
+    from apps.github.models.user import User as GitHubUser
 
 @strawberry.type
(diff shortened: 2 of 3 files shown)


Tanishq-mellu:
Hi maintainers! Thanks for taking a look.

This is my first contribution to the typing improvements tracked in #5080, so I wanted to open it as a draft to validate the approach before making similar changes elsewhere.

The goal of this PR is only to resolve the attr-defined mypy errors by explicitly annotating the github_user attribute on the affected Strawberry node types. I intentionally kept the scope small so it's easier to review.

If this isn't the preferred pattern, I'd really appreciate guidance on the recommended approach. I'm happy to revise the implementation or explore an alternative solution based on your feedback.

Also, I noticed the automated review comments regarding runtime annotations and Strawberry private fields. I haven't addressed those yet because I'd first like to confirm the intended direction with the maintainers rather than making assumptions.
Thanks for your time and feedback

arkid15r (changes requested):
https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md#contributing-workflow

Also make it holistic, if you fix a class of issues in the code I expect some `pyproject.toml` changes as well.

Tanishq-mellu:
Thanks for the clarification. I investigated the scope by temporarily removing the attr-defined suppression locally. That exposes 18 attr-defined errors across ProgramNode, ModuleNode and MilestoneQuery, many of which are related to Strawberry-Django model-backed nodes. I'll work on addressing the issue holistically and update the corresponding pyproject.toml configuration as part of the fix. If there's a preferred typing pattern for these Strawberry-Django nodes that you'd like me to follow, please let me know; otherwise I'll proceed with the approach that removes the need for the suppression.

PR #5058, opened 2026-06-26
Feature/snapshot subscription graphql

HarshitVerma109:
## Proposed change 

Resolves #5056 

Adds GraphQL mutations and query for managing snapshot digest subscriptions. Users can create, update, and cancel their subscription, and unsubscribe directly from email links using a token-based endpoint that requires no login.

## Checklist

- [x] **Required:** I followed the [contributing workflow](https://github.com/OWASP/Nest/blob/main/CONTRIBUTING.md#contributing-workflow)
- [x] **Required:** I verified that my code works as intended and resolves the issue as described
- [x] **Required:** I ran all required checks and tests locally; all warnings addressed and failures resolved
- [x] I used AI for code, documentation, tests, or communication related to this PR

PR #5136, opened 2026-07-04
Snapshot subscription frontend

### LINKED ISSUE #5135: Add snapshot subscription settings page

## Description

Users currently have no way to view or change their subscription preferences from the frontend. This sub-issue adds a settings page where users can manage their snapshot digest subscription, choose what content they want to receive, and select specific chapters and projects.

## Changes

- Settings page with frequency selection and content preference toggles
- Chapter and project search for selecting specific subscriptions
- Per-project toggles for issues, pull requests, and releases
- Settings link in the user dropdown menu
- Unit tests and accessibility tests for the settings page

<br>

- [x] Yes
- [ ] No

HarshitVerma109:
## Proposed change

Resolves #5135 

This PR adds the frontend `/settings` page to allow users to subscribe to snapshot digests and manage their preferences 

## Checklist

- [x] **Required:** I ran all required checks and tests locally; all warnings addressed and failures resolved


HarshitVerma109:
Hi @kasya, please review it and let me know if any changes are required.

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,821 @@
...
+
+          <Button
+            onPress={handleSubscribe}
+            isDisabled={creating}
+            radius="md"
+            className="border border-[#1D7BD7] bg-[#1D7BD7] font-medium text-white transition-all hover:bg-[#1565b0]"
+          >
+            <FaBell className="mr-2" />
+            {creating ? 'Subscribing...' : 'Subscribe'}
+          </Button>
```
kasya: Can we use existing `<ActionButton />` component here for visual consistency? This one doesn't match other buttons we have.

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,821 @@
...
+  handleToggleProjectContent: (projectId: string, key: ProjectContentKey) => void
+}>) {
+  const selectedProjectItems = projectPreferences.map((p) => p.project)
+
+  return (
+    <SecondaryCard>
+      <h2 className="mb-2 text-xl font-semibold">Project Subscriptions</h2>
+      <p className="mb-4 text-sm text-gray-500 dark:text-gray-400">
+        Add projects and choose which content types (Issues, Pull Requests, Releases) you want from
+        each one. Leave empty to skip project-specific content.
```
kasya:
```suggestion
        Select projects and choose which updates you would like to receive (Issues, Pull Requests, Releases). Leave this empty if you don't want project-specific updates.
```

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,821 @@
...
+            <Button
+              variant="bordered"
+              onPress={handleCancel}
+              isDisabled={cancelling}
+              radius="md"
+              className="text-red-600 hover:bg-red-50 dark:text-red-400 dark:hover:bg-red-900/20"
+            >
+              <FaBellSlash className="mr-2" />
+              {cancelling ? 'Cancelling...' : 'Unsubscribe'}
+            </Button>
```
kasya: Can we add a confirmation modal for this? With the setup we have there might be quite a lot of specific details about the subscription - specific projects setup, general OWASP subscription.
With just a single click (including accidental clicks) this can all be lost. I think it would be a good idea to add a confirmation modal.

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,821 @@
...
+          <div className="flex gap-3">
+            <Button
+              onPress={handleUpdate}
+              isDisabled={updating}
+              radius="md"
+              className="border border-[#1D7BD7] bg-[#1D7BD7] font-medium text-white transition-all hover:bg-[#1565b0]"
+            >
+              <FaFloppyDisk className="mr-2" />
+              {updating ? 'Saving...' : 'Save Changes'}
+            </Button>
```
kasya: Just realized we have single buttons for each of this. Can we switch *all* of these to `<ActionButton />` component instead?

Also, Do we really need separate buttons for subscribe and save/update logic? I can imagine this should be doable and reduce code duplication 🤔

kasya (changes requested):
@HarshitVerma109 hi! Great work overall! 👍🏼 
I left some requests bellow ⬇️

HarshitVerma109:
Hi @kasya, please review it and let me know if any changes are required.

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,833 @@
...
+            </p>
+          </ModalBody>
+          <ModalFooter className="flex justify-end gap-3 border-t border-gray-200 px-5 py-4 dark:border-gray-700">
+            <Button
+              variant="ghost"
+              onPress={() => setShowCancelModal(false)}
+              className="rounded-md bg-gray-600 px-4 py-1 text-sm font-medium text-white hover:bg-gray-700 dark:bg-gray-700 dark:hover:bg-gray-600"
+            >
+              Keep Subscription
+            </Button>
```
kasya: Same for this one and button next to it.

I meant that we need to switch *all* buttons to use `<ActionButton />` unless we have some very specific use case and we can't use existing component. If we can't use `<ActionButton />` we can at least use the same styling - same size, same paddings, same border radius etc.

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,833 @@
...
+          <Button
+            variant="bordered"
+            onPress={handleCancel}
+            isDisabled={cancelling}
+            radius="md"
+            className="text-red-600 hover:bg-red-50 dark:text-red-400 dark:hover:bg-red-900/20"
+          >
+            <FaBellSlash className="mr-2" />
+            {cancelling ? 'Cancelling...' : 'Unsubscribe'}
+          </Button>
```
kasya: Is there a reason why you did not switch this button to `<ActionButton />` too? 🤔 

Image

kasya: Can we align these buttons to the right? with `Subscribe` button being on the far right and unsubscribe button to the left of it? I feel like this is more common layout for any buttons placement.

Image

kasya (changes requested):

HarshitVerma109:
Hi @kasya, please review it and let me know if any changes are required.

kasya (dismissed):
The front end looks good now 👍🏼 

@arkid15r would need your help with the backend review.

arkid15r:
> The front end looks good now 👍🏼
> 
> @arkid15r would need your help with the backend review.

If that's from PR https://github.com/OWASP/Nest/pull/5058 then I've just done it.

arkid15r:
@HarshitVerma109 could you make it ready for review?

HarshitVerma109:
@coderabbitai review

HarshitVerma109:
Hi @kasya, please review it and let me know if any changes are required.

frontend/src/app/settings/page.tsx
kasya, line 1: Why am I only seeing these options? 🤔 I believe there were more items.
Also, `Projects` should be here too for general OWASP Snapshot scubscription.

Image

kasya, line 1: Let's update UI so that it's clear that this Frequency is a setting for general OWASP scubscriptions bellow.
Combine these 2 blocks.

Image

Also, let's switch to something smaller for frequency setting - it's taking too much space.
What if we use something like `ToggleButtonGroup` from [HeroUI](https://heroui.com/en/docs/react/components/toggle-button-group)
When one is enabled, another should be disabled (working like a switch). If you know of a better component with the same functionality - feel free to use that! We could even do a dropdown, but I wanted it to be more visual since we only have 2 options.

kasya, line 1: Let's also revert some of the changes.. I think we can remove the per project settings. I know I was the one initially requesting this, but I think this makes the view cluttered and not super user-friendly.

Image

Once you remove per project settings these blocks should only include frequency - still set per project.
This should be the same view and behavior for Chapters subscriptions now.

Also, since we don't need so much space now (we only need to fit it frequency setting) maybe we could switch to a grid view for subscribed projects/chapters. But I'll leave that up to you to decide.

kasya: Please set title for this page. Right now it only says `OWASP Nest`

kasya (changes requested):

kasya:
@HarshitVerma109 Also, I believe we decided that users will be subscribing to project/chapter/committee snapshots from within their pages, but I don't see anything like that added to this PR. 

Do you plan to address that separately?

backend/src/apps/owasp/models/entity_subscription.py
kasya, line 1: We went over this PR with Ark today and have quite a few requests. 
First, we want each subscription support m2m relation to projects, chapters and committees.
With this being said - I don't think we need to have a separate EntitySubscription model. The `SnapshotSubscription` should be the one handling every subscription and this is where we need to set a limit of 5 subscriptions.

Main idea behind it is that users would receive at most 5 emails even if they use all of the available slots, but they still be able to get info about as many entities as they want to.

backend/src/apps/owasp/models/snapshot_subscription.py
```diff
@@ -58,18 +58,6 @@ class Status(models.TextChoices):
...
-        "owasp.Project",
-        blank=True,
-        related_name="snapshot_subscriptions",
-    )
-    subscribed_chapters = models.ManyToManyField(
-        "owasp.Chapter",
-        blank=True,
-        related_name="snapshot_subscriptions",
-    )
-
```
kasya: Here, we should revert this change and also add a Committee M2M relationship to this list.

The intended behavior is that a single subscription acts as a container for everything a user wants to follow.

For example, a user could create one monthly subscription that includes global OWASP Issues and Events, along with a few Projects and a Chapter. All of those entities belong to the same subscription, which counts as **one** of the five subscriptions the user is allowed to create.

The user can then create additional subscriptions for different groups of content. For example, they might have one subscription for AI-related OWASP content, another for AppSec topics, or organize them using any other grouping that makes sense to them.

frontend/src/app/settings/page.tsx
kasya, line 1: On this page, we should remove the distinction between **Snapshot** and **Entity** subscriptions. Everything here represents a **Snapshot Subscription**.
Image
The existing content in this block should remain, but underneath the content toggles we should bring back the multi-selects for **Projects**, **Chapters**, and **Committees**. This allows users to build a subscription like the example I described above - a single subscription containing any combination of global OWASP content and specific OWASP entities.

Ideally, each subscription should also have a user-defined name (though we could fall back to a generic name if needed).

Users should be able to create additional subscriptions directly from this page. My thought is to have a simple `+` button that adds another subscription card below the existing one. Each card would have its own frequency, global OWASP content toggles, and selected Projects, Chapters, and Committees.

We'll display up to **5** subscription cards, matching the current subscription limit.

We will show up to 5 cards like these, due to our current limit.

frontend/src/components/SubscribeButton.tsx
```diff
@@ -0,0 +1,211 @@
...
+        </button>
+      </Tooltip>
+    )
+  }
+
+  return (
+    <>
+      {renderButton()}
+
+      <Modal isOpen={showModal} onClose={() => setShowModal(false)} size="md">
```
kasya, line 216: Image

For this modal, we should display the user's existing subscriptions (if any) so they can choose which subscription to add this new entity to.

Each subscription should show its **frequency**, and ideally its **name** as well. That's one of the reasons I suggested allowing users to name their subscriptions - it makes them much easier to identify when choosing between multiple subscriptions.

Once the user selects a subscription, we simply add the new entity to that subscription's list of tracked entities.

If the user doesn't have any subscriptions yet, the modal should instead show the "Create Subscription" view with the frequency selection, similar to what we have today.

Even when the user already has subscriptions, the "Create Subscription" view should remain available (as long as they haven't reached the limit of **5** subscriptions). This gives them the choice to either add the entity to an existing subscription or create a brand new subscription instead.

kasya (changes requested):
Hi @HarshitVerma109 ! I left quite a few comments on the recent changes. Please read through them carefully, as they involve some fairly significant updates to the implementation.

If anything is unclear, I strongly encourage you to reach out on Slack before implementing the changes. It'll be much quicker to clarify things up front than to go through another round of large revisions, and we're a bit tight on time. I'm happy to answer any questions or talk through the specifics.

HarshitVerma109:
Hi @kasya, please review it and let me know if any changes are required.

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,861 @@
...
+  const client = useApolloClient()
+  const [inputValue, setInputValue] = useState('')
+  const [items, setItems] = useState<{ id: string; name: string }[]>([])
+  const [isLoading, setIsLoading] = useState(false)
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const fetchSuggestions = useCallback(
+    debounce(async (query: string) => {
+      const trimmedQuery = query.trim()
+      if (trimmedQuery.length < 2) {
```
kasya, line 150: Let's make it `<=` - I couldn't find any AI projects, which I expect could be something being searched a lot. 

Image

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,861 @@
...
+            <div className="mb-4">
+              <h4 className="mb-2 text-sm font-semibold text-gray-600 dark:text-gray-300">
+                Content
+              </h4>
+              <ContentToggleGrid
+                fields={SNAPSHOT_CONTENT_FIELDS}
+                preferences={preferences}
+                onToggle={togglePreference}
+              />
+            </div>
```
kasya, line 395: When I created a new subscription from the Project page and only selected frequency and added 1 Project this also subscribed me to ALL global events. 

Image

I think this should be the opposite by default. Also, when we create a new subscription from the settings page - the default *is* none of the toggles selected, so these should match.

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,861 @@
...
+                  {isMutating ? 'Saving...' : 'Save Changes'}
+                </ActionButton>
+              ) : (
+                <Tooltip content="Select at least one content type">
+                  <div className="inline-block cursor-not-allowed">
+                    <ActionButton onClick={handleSave} isDisabled={true}>
+                      <FaFloppyDisk />
+                      Save Changes
+                    </ActionButton>
+                  </div>
```
kasya, line 453: I don't think this is a correct behavior. User should be able to subscribe only to projects, or only to chapters, or to a mix without having to select any global content too, they are in no way connected to each other:

Image

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,861 @@
...
+      </div>
+
+      <Modal isOpen={showDeleteModal} onClose={() => setShowDeleteModal(false)} size="md">
+        <ModalContent className="rounded-lg bg-white shadow-xl dark:border dark:border-gray-800 dark:bg-[#212529]">
+          <ModalHeader className="border-b border-gray-200 px-5 py-4 dark:border-gray-700">
+            <h2 className="text-lg font-bold text-gray-900 dark:text-white">Confirm Delete</h2>
+          </ModalHeader>
+          <ModalBody className="px-5 py-4">
+            <p className="text-gray-600 dark:text-gray-300">
+              Are you sure you want to permanently delete &quot;{name}&quot;? This cannot be undone.
```
kasya:
```suggestion
              Are you sure you want to permanently delete subscription &quot;{name}&quot;? This cannot be undone.
```

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,861 @@
...
+                onDelete?.(subscription.id)
+              }}
+              className={destructiveButtonStyles}
+            >
+              <FaTrash />
+              Yes, Delete
+            </Button>
+          </ModalFooter>
+        </ModalContent>
+      </Modal>
```
kasya: Can we reuse the modal that we already have added? No need to create a separate component at this point (I will add this as a good first issue after we have your project merged in), but at least make it match with the modal we show on "Delete Module" in EntityActions.tsx:290 

Image

frontend/src/app/settings/page.tsx
```diff
@@ -0,0 +1,861 @@
...
+              />
+
+              <EntityMultiSelect
+                label="Committees"
+                searchQuery={SEARCH_COMMITTEES}
+                searchResultKey="searchCommittees"
+                selectedEntities={committees}
+                onAdd={(entity) => setCommittees((prev) => [...prev, entity])}
+                onRemove={(id) => setCommittees((prev) => prev.filter((c) => c.id !== id))}
+              />
```
kasya, line 423: These should be Hero UI [Autoselect](https://heroui.com/en/docs/react/components/autocomplete) component instead. Don't create custom components if we have them in the library we use. And, morevoer, if we already user them elsewhere.

if you need a reference of usage, check `<ModuleForm />` component 👍🏼

kasya (changes requested):
@HarshitVerma109 this is very close to be done! 👍🏼 Looks much better now.
Left some comments ⬇️

HarshitVerma109:
Hi @kasya, please review it and let me know if any changes are required.

kasya (commented):
@HarshitVerma109 This looks good 👍🏼 

One thing I noticed - name seems to be optional, and I think we discussed this before.
The thing with how it is setup right now, if user does not add name we show something like this on their settings page:
Image
And like this when trying to add to an existing sub:

Image

These views differ. And a bigger problem - because names must be unique per user and the name defaults to empty, a user can only ever have one unnamed subscription. Creating a second one without typing a name fails with "A subscription with this name already exists" - which is a confusing message if you didn't realize blank counts as a name.

Three ways out, roughly in order of how little they change:

1. Generate a default in the mutation when name is blank — e.g. "Subscription 3" or seed it from the frequency/first picked entity. Keeps the constraint meaningful - I think **this** was what I suggested originally.
2. Make name required in both create forms. Simplest, but forces naming on the one-click Subscribe flow, which is the flow most likely to want a blank.
3. Make the constraint ignore blanks - store NULL instead of "" (Postgres treats NULLs as distinct in unique constraints). Needs a migration and a model change in this case.

kasya:
@HarshitVerma109 Left  question/comment for you ⬆️

HarshitVerma109:
Hi @kasya, please review it and let me know if changes are required.

kasya (dismissed):
FE looks good to me 👍🏼

arkid15r (approved):

PR #5553, opened 2026-09-08
Feat/unsubscribe filtered view

### LINKED ISSUE #5552: Add One-Click Unsubscribe and Filtered Subscriptions View

## Description

Users need the ability to easily unsubscribe from emails and view a personalized snapshot directly from the links in their digest emails. This sub-issue adds the one-click unsubscribe backend logic and the token-based link logic so users can view their filtered snapshot data securely without needing to log in.

## Changes
- Added a one-click Unsubscribe backend endpoint to handle the unsubscribe link from emails.
- Added a backend GraphQL query to fetch a user's subscription securely using their email token.
- Updated the Snapshot frontend page to read the token from the URL.
- Added a "Your Subscribed Entities" section on the frontend to display their personalized Pull Requests, Issues, and Releases.

<br>

- [x] Yes
- [ ] No

HarshitVerma109:
## Proposed change

Resolves #5552

Adds one-click unsubscribe and personalized snapshot filtering based on subscriber preferences.

## Checklist

- [x] **Required:** I ran all required checks and tests locally; all warnings addressed and failures resolved


HarshitVerma109:
Hi @kasya, you can test the entire email system from this PR. It adds the one-click unsubscribe functionality and a 'view more' button which redirects to the filtered snapshot page.

HarshitVerma109:
Hi @kasya, please review it and let me know if changes are required.

kasya:
> Hi @kasya, you can test the entire email system from this PR. It adds the one-click unsubscribe functionality and a 'view more' button which redirects to the filtered snapshot page.

Hi @HarshitVerma109 ! Did you stack this PR on top of the newsletter digest PR? I'm not able to send emails from this one with the command you provided before 👀 

> Then, you can send the digest emails by running:
>
> make owasp-send-snapshot-emails NEWSLETTER_SNAPSHOT_KEY=2025
>
> And if you ever need to clear the email logs to resend the email for the same snapshot, you can just run this command:
>
> docker exec nest-backend python manage.py shell -c "from apps.owasp.models.email_log import EmailLog; EmailLog.objects.filter(snapshot__key='2025').delete(); print('Cleared')"

Screenshot 2026-09-17 at 10 31 32 PM

HarshitVerma109:
Hi @kasya, please review it and let me know if changes are required.

kasya (changes requested):
@HarshitVerma109 I was able to review this 👌🏼 By the way, just running `make clean` does not fix the issue with `premailer` package missing. Since this is a feature branch you also need to remove the volumes in the docker manually  to really "clean" those.

Anyway, I looked at your changes and the unsubscribe flow works. However, I feel like there should be another screen/confirmation before we unsubscribe and it should be on page we redirect to. Something like `Are you sure you want to be unsubscribed from subscription [name]?` and then on confirmation we can show what you show right now or just a toast, but that should also have a name of the subscription we unsubscribed from. Right now it's not clear on that screen.

Image

HarshitVerma109:
Hi @kasya, for the unsubscribe confirmation screen, would you prefer it to be a modal popup, or should I just render it inline as a normal page?

HarshitVerma109:
I have another question regarding the codebase. Right now, we still have mutations and logic for cancelSnapshotSubscription (which sets isActive to false) and reactivateSnapshotSubscription. However, we previously decided that when a user unsubscribes, we should just permanently delete the subscription from the database instead of deactivating it.

Since we're using deletion now, would it be a good idea to completely remove the cancel/reactivate subscription logic from the backend and frontend to clean up the code? Let me know what you think!

kasya:
@HarshitVerma109 Can we reuse `Delete subscription` that we have right now on that unsubscribe? I mean this thing:
Screenshot 2026-09-20 at 8 59 56 PM
So we could redirect to the settings page (or login first if needed) and then show this pop up?

> Since we're using deletion now, would it be a good idea to completely remove the cancel/reactivate subscription logic from the backend and frontend to clean up the code? Let me know what you think!

Yes, this sounds like a good idea, especially if we'll reuse the logic as I mentioned above 👍🏼

HarshitVerma109:
So @kasya, just to clarify would you like me to remove the unsubscribe page entirely and redirect the user straight to the settings page when they click the unsubscribe link from the email? Or would you prefer that I reuse the same delete confirmation modal on the unsubscribe page itself, show the "Confirm Delete" dialog there, and then display a success message with the subscription name? And yes, I'll remove the cancel and reactivate logic from the codebase (mutations and related files).

kasya:
@HarshitVerma109 I think we can remove the unsubscribe page entirely and redirect the user straight to the settings page when they click the unsubscribe link from the email.

HarshitVerma109:
@kasya, but if we redirect to the settings page, the user would need to log in first anyway. In that case, what would be the purpose of the UUID token in the unsubscribe link? The whole idea behind the token was to allow users to unsubscribe with a single click without needing to log in. If we require login, it basically becomes a regular settings flow and the token-based system would be unnecessary. What do you think?

HarshitVerma109:
@kasya, Instead of removing the unsubscribe page, can we do this?

image

kasya:
> @kasya, Instead of removing the unsubscribe page, can we do this?
> 
> image

@HarshitVerma109 hmm.. good point. I just wanted us to reuse as much functionality that we already have without multiplying pages for no reason. 
I like the option you suggested above, but can we show Subscription name in the Breadcrumbs instead of UUID? 🤔 I know that normally comes from the slug in the URL, but I don't like us showing that on the page - this is not something end user should see. Can we programmatically change it? If not, we can keep the UUID I guess..

